Issue #458, opened 2014-01-09
Can't compile e5265.  Was: Regression introduced by: e5265 Reimplement timers to be light weight


gittywithexcitement:
Hello,
Commit e52652a3231d490741f8280d21c2c9e5d2aa34f0 caused a regression in my program, which now reliably hangs about 7 seconds after startup. It's single threaded, with 4 spawned tasks (using vibe concurrency). 3 of those tasks talk to 1-2 HTTP requests and then sleep for 100 ms. The 4th task runs an external process and polls that for completion then makes an HTTP request and then sleeps for 60 s. The process main calls runEventLoop().

I ran git bisect to narrow it down to this commit:

```
~/.dub/packages/vibe-d-master# git bisect log
git bisect start
# good: [f284b479c90b16bcb4374490e563c35d542468e7] Revert "don't subtract 2 seconds from keepAliveLimit"
git bisect good f284b479c90b16bcb4374490e563c35d542468e7
# bad: [ace498266718b655f5964064a3f6f3580770be4d] Narrow down use of std.range to investigate a strange linker error on Linux/64
git bisect bad ace498266718b655f5964064a3f6f3580770be4d
# good: [7da96c047b1fdbe84e5096ddf20e868131214bbf] Move socket error handling code to an own module and also use pretty error messages in the Win32 driver.
git bisect good 7da96c047b1fdbe84e5096ddf20e868131214bbf
# good: [4186407f968bc9168bbb1ee7fb1d4da98c81bda9] Fix invalid memory access for debug assertion. See #446.
git bisect good 4186407f968bc9168bbb1ee7fb1d4da98c81bda9
# bad: [6b6645adab7130b0bc973706eff0ade752e12077] Don't yield for a "grow_when_full" TaskPipe - as this is specifically used to avoid yielding.
git bisect bad 6b6645adab7130b0bc973706eff0ade752e12077
# bad: [e52652a3231d490741f8280d21c2c9e5d2aa34f0] Reimplement timers to be light weight - Timer is now a struct instead of an interface.
git bisect bad e52652a3231d490741f8280d21c2c9e5d2aa34f0
# good: [edabcc884c3ac71f9913f73818afcf2bac45b89c] Optimize AutoFreeListAllocator.realloc to reuse buckets if possible and to use m_baseAllocator.realloc if possible. See #446.
git bisect good edabcc884c3ac71f9913f73818afcf2bac45b89c
[...]

PR #460, opened 2014-01-13
Use getaddrinfo and parse /etc/hosts on linux

gittywithexcitement:
The motivation for this change is that my application needs /etc/hosts to be parsed and used.
So I changed usage of evdns_base_resolve_ipv4/ipv6 to getaddrinfo, because evdns_base_resolve_XXX doesn't use the hosts file.

You may be wondering why I then applied a commit "Change EventDriver.resolveHost parameter..."  The answer is that I can't compile the current master HEAD (see https://github.com/rejectedsoftware/vibe.d/issues/458). So I branched from the most recent commit that compiles for me, and made my changes. Then I cherry picked d49f6152067633ee26f31bfaf63c12f752a8bbe6 and resolved conflicts.

So far I've only tested with an /etc/hosts containing localhost. Tomorrow I will test with a more challenging /etc/hosts file that maps nameFoo to 127.0.0.1 and nameBar to a remote IP. I have every reason to believe it will work.

### DIFF
--- a/source/vibe/core/driver.d
+++ b/source/vibe/core/driver.d
@@ -79,7 +79,7 @@ interface EventDriver {
 
 	/** Resolves the given host name or IP address string.
 	*/
-	NetworkAddress resolveHost(string host, ushort family, bool no_dns);
+	NetworkAddress resolveHost(string host, ushort family, bool use_dns);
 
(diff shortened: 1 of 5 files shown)


s-ludwig:
This PR was already integrated into master manually and can be closed.

Issue #724, opened 2014-07-16
Add support for reverse DNS lookups.


s-ludwig:
This is important for SSL client certificate validation. See #723.

PR #728, opened 2014-07-16
Added reverse dns lookup for https server client certificate validation

EricMCornelius:
Using the std.socket method for reverse dns lookups for now - I don't know enough about the various driver internals to know if there's a better asynchronous option available in some cases, but it at least meets the immediate needs.

reference: https://github.com/rejectedsoftware/vibe.d/issues/724

Dunno if we need to worry about the potential for a SocketFeatureException on the toHostNameString() call or not.

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -43,7 +43,7 @@ import std.functional;
 import std.string;
 import std.typecons;
 import std.uri;
-
+import std.socket;
 
 /**************************************************************************************************/
 /* Public functions                                                                               */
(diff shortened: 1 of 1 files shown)


WebFreak001:
@s-ludwig would a rewrite of this PR still be appropriate or should that be moved into the new vibe-core package?

s-ludwig:
I've opened a ticket against vibe-core. I'd say any kind of new feature should be focused on vibe-core, as that is going to be the new default soon (probably for 0.8.3), after which vibe-d:core will be deprecated anyway. Of course if someone needs this specifically in conjunction with vibe-d:core and wants to implement it, there is no reason to not include it anyway.

PR #737, opened 2014-07-22
Support path normalization before matching routes

### LINKED ISSUE #667: Support path normalization before matching routes in URLRouter
See #410. `URLRouter` could get an optional mode, where the request path is first normalized (e.g. "/a//b/../c" getting "/a/c"). This could also be achieved using an intermediate request handler that alters the path. Another approach would be to add an intermediate handler that redirects to the normalized path instead (at least for GET requests).

sobels:
Adds support for path normalization (fixes #667) by adding a new class, `NormalizedURLRouter`. Rationale for using a new class:
- Using a new class allows `URLRouter` to encapsulate only the logic for matching routes, and lets another class deal with what paths are considered canonical.
- It seems to be more convenient than adding an intermediate request handler: it is more brief, and the odds are probably pretty low that someone would only want to normalize paths for some subset of routes.

Several sites claim that redirects are better for SEO, but [several](https://support.google.com/webmasters/answer/139066?rd=1) [documents](http://googlewebmastercentral.blogspot.com/2010/04/to-slash-or-not-to-slash.html) show that there are multiple solutions to this (e.g.  `rel="canonical"` links, sitemaps). Perhaps redirects should be an optional mode for `NormalizedURLRouter`, but for now I think a simpler and slightly faster default would be to avoid the redirect.

For now, the only normalization performed is to resolve `..`, `.`, and `/`.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -200,6 +200,56 @@ final class URLRouter : HTTPRouter {
 	}
 }
 
+/**
+	Does everything a URLRouter does, but also normalizes URLs before matching them against routes.
+*/
+final class NormalizedURLRouter : HTTPRouter {
+	private URLRouter m_router;
+	
+	this(string prefix = null)
+	{
+		m_router = new URLRouter(prefix);
+	}
+	
+	@property string prefix() const { return m_router.prefix; }
+
+	/// Adds a new route for requests matching the specified HTTP method and pattern.
+	URLRouter match(HTTPMethod method, string path, HTTPServerRequestDelegate cb)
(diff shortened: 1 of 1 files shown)


Geod24:
> But I'd keep this separate from the issue of resolving `.`, `..` and multiple consecutive slashes. I'm currently leaning towards integrating this into `URLRouter`, so that it's possible to later implement a fast and allocation-less match algorithm that normalizes while matching.

Is it still something you'd be keen to accept ? This path issue is becoming more and more present on my end and I'm also quite interested in #1062 .

PR #763, opened 2014-08-04
Adds a persist flag in FileDescriptorEvent. Fixes #695

### LINKED ISSUE #695: High CPU usage when using FileDescriptorEvent
See the [related forum thread](http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/17924/) for a full description.

oglu:
This fixes the high cpu usage when using FileDescriptorEvent. 

It adds a persist flag (true by default, so as to preserve the original behavior) that controls whether the event is persisted with libevent or reattached after each callback.

I found it useful with zmq and createFileDescriptor


s-ludwig:
Got implemented with a more general API as part of #1596.

PR #816, opened 2014-09-10
Fixed worker thread silent assertions

etcimon:
All worker threads are currently silent when assertions are thrown outside of tasks. This fix closes the event loops from other worker threads when one of them errors out, and logs the error appropriately.


s-ludwig:
I've added this now with the twist that `Error` causes an `exit(-1)` instead of `exitEventLoop(true)`.

PR #842, opened 2014-09-24
Turned vibe.web.rest into a package

Geod24:
There are various addition / enhancement to `vibe.web.rest` that can be done.
Some of them (handling of different origin of parameter, automatic generation of OPTIONS handler, custom serialization...) while still being discussed with @Dicebot , will require reasonable amount of code.
Thus, this commit split the existing `vibe.web.rest` into a package, including modules `client`, `server`, and `common`.
Besides the new `package.d`, and some import tweaking, no code has been changed.


Geod24:
@s-ludwig : Is this still a no-go ? From https://github.com/D-Programming-Language/dub/issues/424, it seems like unittests runs as expected. In addition, the REST module doesn't have name clash, so there's no need for disambiguation.
We could introduce `package.d`, then deprecate it on the next release (once the doc is rewritten). This way, we won't break any code.

Geod24:
Ping @s-ludwig

s-ludwig:
Hm, splitting up in general sounds like a good idea, but what about moving the private parts to `vibe.web.internal.*` or `vibe.internal.rest.*` instead? The proposed `vibe.web.rest.client` and `vibe.web.rest.server` modules would each just contain a single function/class, which is a little ugly API-wise (plus the potential breakage/deprecation). By moving only the private implementation, we can solve the module size issue without any publicly visible restructuring.

s-ludwig:
`vibe.web.internal.*` would be better than `vibe.internal.*` w.r.t the upcoming split into sub packages.

s-ludwig:
A module containing just a single public symbol seems a little empty, just like a package with only a single module. It also requires more text to import for no perceivable benefit. On the other hand, splitting a module like `std.algorithm` makes sense from both, an implementor's view and the user's view.

s-ludwig:
> or best compile times

I somehow suspect that there will be a turning point, once you have thousands of source files ;)

No, but seriously, first, since these are both templates, moving the private code should really have about the same effect on compile time if local imports are used. And more importantly, while compile times are of course important, a logical public interface is more important to me.

I don't agree that reducing module size is not a worthy goal. I like to keep a mental model of the source code I'm working with and at a certain size that is simply not possible anymore. Modules like the classing `std.datetime` basically degrade into a black box manageable only with search or "got to definition" features. The REST module is not yet as far, but it's already quite difficult at times to find a particular piece of code.

>  two symbols are completely unrelated

Err... no. They are *obviously* extremely related, even if you might usually only use one of them when you import the module. I bet that the percentage of modules where all symbols are typically always used is very close to zero.

Geod24:
> I like to keep a mental model of the source code I'm working with and at a certain size that is simply not possible anymore.

So do I. But a mental model is not only about the size, it's also about intent of a module / package.
Things like `std.datetime` are such a mess because it's not always obvious how to split that up into manageable units. OTOH, client/server is a very sweet spot, because it provide an obvious separation, and even better, symmetry.

Also, most of the stuff from `vibe.internal.*` are generic code, not module specific.

s-ludwig:
I'd rather put it in `vibe.web.internal` than in `vibe.internal` for that reason. You can easily have `vibe.web.internal.client` and `vibe.web.internal.server`. `vibe.web.rest` would then just contain to very thin skeleton templates. Nothing that should disrupt the mental model.

Geod24:
I'm sorry but I don't understand how providing a client implementation in a module and a client implementation in another is not okay ('A module containing just a single public symbol seems a little empty'), while providing two thin wrapper in the same module is. This is just reinventing `package.d`, but without the possibility of accessing symbols individually.

Geod24:
> This is just reinventing `package.d`, but without the possibility of accessing symbols individually.

You can actually access them individually, because those symbols would need to be public anyway. They can't be `package`, since we don't have access to `package(identifier)` yet.

s-ludwig:
This is about the *API*, the interface that is facing the library user. IMO, the most important thing there is that it is "harmonic" and well organized when looking at it from the outside. The fact that it may be desirable to split up *private* code into multiple modules should simply not affect the public API. It would be another matter if there were a multitude of public symbols involved.

Geod24:
> I'd rather put it in `vibe.web.internal` than in `vibe.internal` for that reason.

I did a sketch overnight, since the code base becomes more and more tricky to work with.

Would something like the current diff works ? The plan would be:
- `vibe.web.internal.rest_{client,server}` (client is too generic, and so is server);
- `vibe.web.internal.{routing,serialization}: Only routing for the moment. I'd like to unify routing ability between both module. One of the idea you add and I'd like to implement is switch the `MethodStyle` at compile time.
- `vibe.web.internal.rest` for things that need to be shared.

At some point, the validation part (which is currently in `internal.rest` for arbitrary reason) could be made available to both.

Geod24:
Guess this can be killed now.

PR #860, opened 2014-10-09
Updated json to correctly handle double and float nan

dmonagle:
Previously a double nan was serializing to the string "nan" which is invalid JSON.

Updated to use undefined instead of nan for both serializing and deserializing.

Rationale:

ECMA-262 states "The undefined value is [...] used when a variable has not been assigned a value", while "The null value [...] represents the null, empty or non-existent reference."

While I was at it, cleaned up incorrect grammar and spelling mistakes.

### DIFF
--- a/source/vibe/data/json.d
+++ b/source/vibe/data/json.d
@@ -360,7 +360,7 @@ struct Json {
 		else static if (is(T == string)) return m_string;
 		else static if (is(T == Json[])) return m_array;
 		else static if (is(T == Json[string])) return m_object;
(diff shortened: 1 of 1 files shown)


yazd:
I think `nan` should be serialized as `null` instead of `undefined`. `undefined` is invalid JSON.

s-ludwig:
You are right (see also #958). I was hoping that I could avoid most work on `vibe.data.json` in favor of `stdx.data.json`, but this should be fixed. This could also be made configurable like for `stdx.data.json`.

PR #877, opened 2014-10-21
Accepting scoped default initialization

etcimon:
I'm currently trying to carry over scoped default declaration/initializations of class (inheritance) objects from C++.

This minor modifications seems to fit the bill pretty well, although I could probably use this in `scoped!` objects.

Here's how it looks:

	import vibe.utils.memory;
	import std.stdio;
	class A { void print() { writeln("A"); } }
	class B : A { override void print() { writeln("B"); } }
	void print(A a) { a.print(); }
	void main()
	{
		FreeListRef!B b;
		print(b);
	}

This prints `B`, whereas without the change in this pull request, the above code triggered a memory error.

### DIFF
--- a/source/vibe/utils/memory.d
+++ b/source/vibe/utils/memory.d
@@ -660,7 +660,15 @@ struct FreeListRef(T, bool INIT = true)
 	}
 
 	@property const(TR) get() const { checkInvariants(); return m_object; }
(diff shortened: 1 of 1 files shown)


etcimon:
Pushed into memutils

PR #881, opened 2014-10-23
[WIP] An REPL module for vibe.d ?

Geod24:
I found myself needing this while working on a non-web project.
I thought this could be a good addition to ab asynchronous I/O framework, so here it is.
Few questions:

1) Will you want such functionality in the framework ?
2) Will you want it in it's current form ? (Yeah the code is ugly).
3) Any idea of worthy addition ? I think a debug console for HTTP would be kinda cool. Most basic example would be to set the debug level using the console. But I'm a bit short on idea right now, as I wrote it for other purpose.

Note: Needs testing on Windows. There is currently a small bug, which forces you to hit enter after you typed `exit` or `quit`. My money's on the thread (as Ctrl+D works just fine).

### DIFF
--- a/source/vibe/core/repl.d
+++ b/source/vibe/core/repl.d
@@ -1,0 +1,168 @@
+/**
+	Provides facility to easily set up a $(LINK2 http://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop, REPL) with an event loop.
+
+	The following code demonstrate the possibility of this module:
+
+	----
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm this has been sitting here far too long. So I generally think a REPL would be nice to have, but it should take the form of a separate project/package, especially with the upcoming separation into multiple sub-libraries. I'll close this PR, but maybe this could be revived one day when there is some spare time.

PR #886, opened 2014-10-29
Implementation of I18n model field

jdavidls:
I18n model field implements i18n fields compatible with mongodb text index.

usage:

``` D
struct MyModel {
    struct Text{
        string description;
        string more_i18n_info;
    }
    I18n!string name;
    I18n!Text texts;
}

// under requestContext
auto obj = new MyModel();
obj.name = "model-name";
obj.texts.description = "...";
obj.texts.more_i18n_info = "...";

/*
serializeJson(obj):
{
    name: [{
        language: req.language,
        translation: "model-name"
    }],
    texts: [{
        language: req.language,
        translation: {
            description: "...",
            translation: "...."
        }
    }]
}

 */
```

### DIFF
--- a/source/vibe/web/web.d
+++ b/source/vibe/web/web.d
@@ -687,3 +687,78 @@ private RequestContext createRequestContext(alias handler)(HTTPServerRequest req
 
 	return ret;
 }
+
+
+/* JDAVIDLS
+	I18n model field
+
+	i must to place here intread of i18n.d to can access to private s_requestContext
+*/
+
+
+
+
+private string generateI18nProperties(T)(){
+	string ret;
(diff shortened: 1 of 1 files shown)


WebFreak001:
@s-ludwig why is this still open and unlooked at?

s-ludwig:
I remember having looked at this, but it wasn't clear to me how to proceed with this on a conceptual level, so I couldn't give useful feedback. I'm still not sure if this is general enough to be included in the code framework. I could imagine lots of alternative JSON representations that could be desirable, so that this seems to be very application/protocol specific.

Geod24:
Closing as it seems to have been abandoned, and the code presented is simple, but doesn't come with tests or much documentation.

PR #896, opened 2014-11-12
Feature/i18n enhancement

ncoe:
Adding some unittests to the translation framework in the vibe.web.web module.
Adding support for the msgctxt tag that can be specified in po files.

### DIFF
--- a/source/vibe/web/i18n.d
+++ b/source/vibe/web/i18n.d
@@ -93,9 +93,9 @@ html
 
 mixin template translationModule(string NAME)
 {
-	mixin template decls_mixin(string LANG, size_t i) {
-		mixin("enum "~LANG~"_"~keyToIdentifier(decl_strings[i].key)~" = "~decl_strings[i].value~";");
-	}
+//	mixin template decls_mixin(string LANG, size_t i) {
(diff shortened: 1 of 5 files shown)


source/vibe/web/i18n.d
```diff
@@ -191,27 +193,67 @@ package template GetTranslationContext(alias METHOD)
...
+ * msgctxt - disabmbiguating context, like variable scope (optional, defaults to null)
+ * msgid - key to translate from (required)
+ * msgid_plural - plural form of the msg id (optional)
+ * msgstr - value to translate to (required)
+ * msgstr[0] - indexed translation for handling the various plural forms
+ * msgstr[1] - ditto
+ * msgstr[2] - ditto and etc...
+ */
+
+// TODO: Handle msgctxt
```
s-ludwig: I there something still left for msgctxt?

s-ludwig:
I'd pull this and then remove the commented out code if there are no objections.

source/vibe/web/i18n.d
ncoe: The msgcntxt should be complete, and that TODO can be removed. The only thing remaining should be additional overloads that support plurals and context for plurals.

s-ludwig:
Okay, thanks! I've removed the commented out code; if necessary, we can always dig it up from the GIT history.

PR #902, opened 2014-11-24
supporting serving an index.html for directories

MartinNowak:


source/vibe/http/fileserver.d
```diff
@@ -162,7 +160,32 @@ class HTTPFileServerSettings {
...
 		serverPathPrefix = path_prefix;
 	}
-} 
+
+	deprecated("Use .options and HTTPFileServerOption.failIfNotFound instead.")
+	@property bool failIfNotFound() const { return options & HTTPFileServerOption.failIfNotFound; }
+
+	deprecated("Use .options and HTTPFileServerOption.failIfNotFound instead.")
+	@property void failIfNotFound(bool val) {
+		if (val)
```
Temtaime: options = (options & ~HTTPFileServerOption.failIfNotFound) | (-cast(int)val & HTTPFileServerOption.failIfNotFound);
Without ifs it's shorter and may work.

s-ludwig: Seems a little too cryptic for too little benefit, IMO. In cases where not both is possible, I personally usually favor the clearer and more readable variant over the more concise one.

PR #911, opened 2014-12-03
Allow findAndModify to take an options parameter: BREAKING CHANGE

intrica:
The third parameter to findAndModify was previously set as the returnFieldSelector. The problem with this is that the findAndModify function in mongo takes quite a large number of optional parameters, such as the very commonly used "new" parameter.

This fix replaces the third parameter with an options argument which should take anything that can be iterated as an associative array (such as a Json or Bson object)

Example:

```D
auto result = collection.findAndModify(
    ["_id": Bson(_id), "count": Bson(count)],
    ["$inc": Bson(["count": Bson(1)],
    ["new": true]
);
```

To achieve this, I have strayed away from using the CMD struct, but the struct doesn't allow for optional parameters.

### DIFF
--- a/source/vibe/db/mongo/collection.d
+++ b/source/vibe/db/mongo/collection.d
@@ -198,19 +198,14 @@ struct MongoCollection {
 	  Throws Exception if a DB communication error occured.
 	  See_Also: $(LINK http://docs.mongodb.org/manual/reference/command/findAndModify)
 	 */
-	Bson findAndModify(T, U, V)(T query, U update, V returnFieldSelector)
+	Bson findAndModify(T, U, V)(T query, U update, V options)
 	{
-		static struct CMD {
-			string findAndModify;
-			T query;
-			U update;
-			V fields;
-		}
-		CMD cmd;
+		auto cmd = Bson.emptyObject;
 		cmd.findAndModify = m_name;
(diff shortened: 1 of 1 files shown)


s-ludwig:
With `find`, there is always the possibility to define the query like this:
```
{
  "$query": {"_id": ..., "count": ...},
  "$inc": {"count": 1"},
  "$new": true
}
```
I don't know if this works for `findAndModify`, though, possibly not. The problem is that we can't make a silent breaking change like the one here. Alternatively, adding an overload with an *additional* options parameter would be a workable solution.

It would also be nice to assemble the command without doing dynamic memory allocations (which `Bson` does when appending fields), but I don't have a handy solution for this (might require some extensions to `vibe.data.serialization`).

vuaru:
I was wanting to use findAndModify with ```"remove": true``` because findAndRemove didn't exist, but it seems neither option is possible at the moment.
+1

sigod:
@vuaru, I've studied code a little. It seems to me we can invoke `findAndModify` by calling `find` with
```
{
    "findAndModify": "<name of the collection",
    ...
}
```
Haven't tested it, yet.

s-ludwig:
I've simply created a new method with a different name to make this a non-breaking change.

PR #914, opened 2014-12-04
writeFormBody accepts ranges of pairs

Panke:
I've needed this, because the formdata is stored in a SQL table and I did not want to construct a AA just to send it directly.


s-ludwig:
Looks good, merging in. Sorry for the long review delay.

s-ludwig:
Cherry picked instead of merged due to conflicts.

PR #926, opened 2014-12-14
REST enhancement #1.1: Refactoring for client side code generator

Geod24:
While I was working on implementing `@queryParam` / `@bodyParam` parameters (following #908), I was blocked by the client-side code generator which hardcoded [some stuff](https://github.com/rejectedsoftware/vibe.d/tree/master/source/vibe/web/rest.d#L366).

In addition, the usage of `Json` was a bit annoying, and blocking for the evolution of this module (while not on my priority list, I'd like to make serialization customizable).

Downsides:
- Potentially breaks code, but none that #908 did not already broke. If someone extended the client implementation, parameter change, so CT error. However, I think that this use case is rare and that some might be doing to mimic newly added functionalities (e.g., modifying the headers).
- Moar template: Obviously, `genQuery` / `genBody` introduce some more template. It is however not so important, as the code is small. On the REST example, file size went from 20178584 to 20193288 (Linux x86_64). This is an increase of 0.7%.

Upsides:
- Moar doc;
- Less symbols introduced inside the scope of the function;
- request now respects the `Content-Type` set by the user, if any;
- Reduce dependency to Json;
- More optimized code;
- Makes any addition that want the parameters at a specific place easy to implement !

Note:
There is a style update (first commit) and the refactoring itself (second). It's way better seen splitted :)

Usual ping to @Dicebot .

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -351,36 +351,29 @@ class RestInterfaceClient(I) : I
 		import vibe.data.json : Json;
 		import vibe.textfilter.urlencode;
 
-		Json request(HTTPMethod verb, string name, Json params, bool[string] param_is_json, in ref InetHeaderMap hdrs) const
+		/**
+		 * Perform a request to the interface using the given parameters.
+		 *
+		 * Params:
+		 * verb = Kind of request (See $(D HTTPMethod) enum).
+		 * name = Location to request. For a request on https://github.com/rejectedsoftware/vibe.d/issues?q=author%3ASantaClaus,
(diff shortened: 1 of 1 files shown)


Geod24:
Ping @s-ludwig .

s-ludwig:
Okay merging in. Sorry for the delay!

Geod24:
Yaaayyyy ! Thanks :)

s-ludwig:
I only have a vague feeling how long this PR has been sitting around because I was so busy, but it must have felt like ages ;) (or like waiting for a new DMD release...)

PR #937, opened 2014-12-20
Allow policy-based serialization

lultimouomo:
An explicit Policy parameter allows to customize serialization of types
without having to modify them, which is useful if the types are supplied
by a library or if different serializations are needed.

The PR is not ready yet, needing documentation and tests, but it's working, so I would like some feedback.

An usage example (that pushed me to write the code) is serializing objects synthesized by [dproto](https://github.com/msoucy/dproto), which can be done with such a policy:

```
template BufferPolicy(Buf) if (__traits(compiles, Buf.init == Buf.init.opGet))
{
    auto toRepresentation(Buf b) {
        return b.opGet;
    }
    Buf fromRepresentation(typeof(Buf.init.opGet) v){
        Buf b;
        b = v;
        return b;
    }
}
```
which transforms the serialized form from:
```
{"fieldname":{"opGet":"fieldvalue"}}
```
to
```
{"fieldname": "fieldvalue"}
```
(Actually, serializing dproto objects without a custom policy fails, due to https://issues.dlang.org/show_bug.cgi?id=12363 ; but it would look like the former if it worked)

I also would have liked to add a template parameter (defaulted to DefaultPolicy) to ```serialize``` and ```deserialize```, but I don't think it can be done without breaking some existing code.

It should be possible to compose policies with something like
```
template CompositePolicy(T)
{
	static if (isPolicySerializable!(FirstPolicy, T))
	{
		alias toRepresentation = FirstPolicy!T.toRepresentation;
		alias fromRepresentation = FirstPolicy!T.fromRepresentation;
	} else {
		alias toRepresentation = FallbackPolicy!T.toRepresentation;
		alias fromRepresentation = FallbackPolicy!T.fromRepresentation;
	}
}
```
but I still have to work on it, and I wonder if it would be possible to create a general policy composer.

### DIFF
--- a/source/vibe/data/serialization.d
+++ b/source/vibe/data/serialization.d
@@ -23,6 +23,10 @@
 				and will always be serialized as a string.)
 			$(LI Any $(D Nullable!T) will be serialized as either $(D null), or
 				as the contained value (subject to these rules again).)
+			$(LI Types satisfying the $(D isPolicySerializable) trait for the
+				supplied $(D Policy) will be serialized as the value returned
+				by the policy $(D toRepresentation) function (again subject to
+				these rules).)
 			$(LI Types satisfying the $(D isCustomSerializable) trait will be
 				serialized as the value returned by their $(D toRepresentation)
 				method (again subject to these rules).)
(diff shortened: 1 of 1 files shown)


s-ludwig:
I like this a lot and it would also be able to handle some of the more specific serialization needs that came up in the past. I'll have to allocate some more time to look at the details of the interface in the next days.

lultimouomo:
Hi Ludwig,

did you have the chance to look at this?

s-ludwig:
Took the time now finally and all looks good! This is a nice concept for later extension, too (e.g. adding an UDA for customizing fields or maybe supporting more ways to customize the output within the Policy).

PR #942, opened 2014-12-24
Better JSON.opt comment

GeorgeSapkin:
Added a matching JSON.opt comment from BSON.opt


s-ludwig:
I've cherry picked the commit instead, because something seems to have gone wrong with the rebase.

PR #943, opened 2014-12-24
Fix libasync apache bench and quick optimization

etcimon:
Here are a couple quick fixes for the libasync driver:

- The apache bench tool didn't work because the connection wasn't being closed correctly.

- The circular buffer has been moved to heap to avoid making the GC scan it, which speeds up the server by ~70%.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -39,7 +39,7 @@
 		{
 			"name": "libasync",
 			"dependencies": {
-			  "libasync": { "version": "0.6.5" }
+			  "libasync": { "version": "0.6.6" }
 			},
 			"versions": ["VibeUseNativeDriverType", "VibeLibasyncDriver"]
 		},
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -12,6 +12,7 @@ import vibe.core.core;
 import vibe.core.driver;
 import vibe.core.drivers.threadedfile;
 import vibe.core.log;
+import vibe.inet.path;
 
 import libasync.all;
 import libasync.types : Status;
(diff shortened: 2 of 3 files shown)


dariusc93:
@etcimon is there anything within libevent that you could port/rewrite into libasync to improve the performance or do you know what the overhead would be?

etcimon:
> @etcimon is there anything within libevent that you could port/rewrite into libasync to improve the performance or do you know what the overhead would be?

The code itself isn't much different and mostly optimal. Most of the optimization at this point would come from using LDC, and a better allocator strategy. I have this one partly figured out with lockless, Task-scoped allocation, but it's only a proposal for now.

Of course, the speed improvements of a task-scoped allocator are going to trickle down into vibe.d and improve overall performance many times over.

s-ludwig:
Looks like something went wrong with the GIT history. There are some commits duplicated from the main master branch. Can you rebase again and pick only your five original commits?

etcimon:
All good!

By the way, Botan is already passing many tests on git master (all block cipher and hash tests). I'll have a pull for a new stream coupled with a native TLS/Crypto library in less than 2 weeks :)

s-ludwig:
Very nice! More and more building blocks to emancipate D from C/C++! BTW, do you hand translate everything or do you use some kind of (semi-)automation?

s-ludwig:
Oh, by the way, 70% for not scanning a 64k buffer sounds an awful lot! I need to remember going through the source to find all similar places (win32 driver and zlib stream come to mind).

etcimon:
> BTW, do you hand translate everything or do you use some kind of (semi-)automation?

I used 95% search & replace with regex in Mono-D (there must've been 200 custom regex expressions overall), then I moved all the implementations into the headers and renamed all the files to .d through a powershell comment. Then I wrote a little mod in DFix to change the naming to camelcase. However, DFix symbols don't seem to catch most function calls so it wasn't very useful. I also added `m_` throughout the library manually. I must've went through each file 10 times =)

I also had to modify the containers to be compatible with custom allocators, and made them all reference counted. The hardest part was getting const-correctedness to work in D and through the `FreeListRef`. You should look at the monster that your `FreeListRef` has become for operator overload / interface casting to work ;)

The second hardest part was writing the simd and asm library in dmd (botan/utils/simd/).

I'll probably move all those containers and custom allocators into an overhead library called `memutils` =)

PR #945, opened 2014-12-29
Add HTTPClientResponse.switchProtocol()

lultimouomo:
Replaces rejectedsoftware/vibe.d#929 which I accidentally opened on my master branch.

Since the response does not have access to the request headers,
the expected protocol must be supplied as a parameter.

This is needed, for instance, to create client-side websockets.

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -20,6 +20,7 @@ import vibe.inet.url;
 import vibe.stream.counting;
 import vibe.stream.ssl;
 import vibe.stream.operations;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Maybe `switchProtocol` would be a better naming choice, because it would be consistent with `HTTPServerResponse.switchProtocol`. Other than that, I just noticed that for both, `HTTPServerRespose` and `HTTPClientResponse`, an additional overload taking a scoped delegate to handle the connection would be good to have, to avoid the GC `new`. But that is independent of this pull request.

PR #946, opened 2014-12-29
Add doHandleWebsocket function

lultimouomo:
This allows establishing a websocket connection directly in an
HTTPServerRequestDelegate, which is useful if one needs to make some
checks on the request (and possibly send an http response with an error
code) before opening the websocket.

Signed-off-by: Luca Niccoli <l.niccoli@awtech.it>

### DIFF
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -66,7 +66,7 @@ class WebSocketException: Exception
 /**
     Establishes a web socket conection and passes it to the $(D on_handshake) delegate.
 */
-void doHandleWebsocket(WebSocketHandshakeDelegate on_handshake, HTTPServerRequest req, HTTPServerResponse res)
+void handleWebsocket(scope WebSocketHandshakeDelegate on_handshake, HTTPServerRequest req, HTTPServerResponse res)
 {
 	auto pUpgrade = "Upgrade" in req.headers;
 	auto pConnection = "Connection" in req.headers;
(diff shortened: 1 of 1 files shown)


source/vibe/http/websockets.d
```diff
@@ -64,55 +64,60 @@ class WebSocketException: Exception
 }
 
 /**
-	Returns a HTTP request handler that establishes web socket conections.
+    Establishes a web socket conection and passes it to the $(D on_handshake) delegate.
 */
-HTTPServerRequestDelegate handleWebSockets(WebSocketHandshakeDelegate on_handshake)
+void doHandleWebsocket(WebSocketHandshakeDelegate on_handshake, HTTPServerRequest req, HTTPServerResponse res)
```
s-ludwig: The `on_handshake` parameter could be `scope WebSocketHandshakeDelegate` here to avoid creating heap delegates at the call site.

s-ludwig:
Thanks, this change makes a lot of sense. I'm just still a little unsure about the best name. Usually I'd simply use an additional overload of `handleWebSockets` (analogous to `handleBasicAuth`), but of course the plural form doesn't really fit here. The `doXYZ` style generally shouldn't appear in the public API, ideally everything should be a simple `verb` or `verbNoun` combination. If no better alternative comes up, I'd suggest simply `handleWebSocket`.

lultimouomo:
```handleWebSocket``` seems fine to me.

PR #947, opened 2014-12-31
Add automatic ping on websockets

lultimouomo:
The ping interval is fixed at 60 seconds.
The best thing would probably be to set it in the server settings, thus also allowing to disable it entirely.
For this, the ```WebSocket``` must have access to ```HTTPServerSettings```; the best way would probably be to expose a const reference to ```HTTPServerRequest.m_settings```, which I'm going to do in the next commit of this pull request.

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -522,11 +522,19 @@ final class HTTPServerSettings {
 	*/
 	bool useCompressionIfPossible = false;
 
+
+	/** Interval between WebSocket ping frames.
+
+		The default value is 60 seconds; set to Duration.zero to disable pings.
+	*/
+	Duration webSocketPingInterval;// = dur!"seconds"(60);
+
 	this()
 	{
 		// need to use the contructor because the Ubuntu 13.10 GDC cannot CTFE dur()
 		maxRequestTime = 0.seconds;
 		keepAliveTimeout = 10.seconds;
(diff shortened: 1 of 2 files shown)


source/vibe/http/server.d
```diff
@@ -690,6 +698,13 @@ final class HTTPServerRequest : HTTPRequest {
 			Remarks: Requires the HTTPServerOption.parseCookies option.
 		*/
 		Session session;
+
+		/** The settings of the server serving this request.
+		 */
+		@property const(HTTPServerSettings) serverSettings() const
+		{
+			return m_settings;
+		}
```
s-ludwig: I'd leave this as `package` for now to avoid any potential necessity for API breakage later. If it turns out that this is useful for external libraries/applications, we can still make it public.

s-ludwig:
Looks good to merge, except for the public `serverSettings` property.

lultimouomo:
I made the settings accessor ```package```, and I fixed a bug that came out testing ping timeouts:
the ```m_pingTimer``` task was trying to directly close the connection, which was most probably acquired by ```m_reader```. The code now makes ```m_reader``` close the connection itself, like the other code paths that lead to the connection being closed.

s-ludwig:
Great, thanks! I'll merge as soon as the Travis build is ready.

PR #965, opened 2015-01-25
Add access to SSL peer certificate

rix0rrr:
Provide access to a client's certificate information, which makes client
cert based auth/cert SSO a lot more useful.

### DIFF
--- a/source/vibe/stream/openssl.d
+++ b/source/vibe/stream/openssl.d
@@ -158,10 +158,10 @@ final class OpenSSLStream : SSLStream {
             ASN1_OBJECT *obj = X509_NAME_ENTRY_get_object(e);
             ASN1_STRING *val = X509_NAME_ENTRY_get_data(e);
 
-            auto longName = fromStringz(OBJ_nid2ln(OBJ_obj2nid(obj)));
+            auto longName = OBJ_nid2ln(OBJ_obj2nid(obj)).to!string;
             auto valStr = cast(string)val.data[0 .. val.length];
 
-            m_peerCertificate.subjectName[longName] = valStr;
+            m_peerCertificate.subjectName.addField(longName, valStr);
         }
     }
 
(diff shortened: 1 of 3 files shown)


source/vibe/stream/openssl.d
```diff
@@ -143,6 +146,25 @@ final class OpenSSLStream : SSLStream {
 		checkExceptions();
 	}
 
+    /** Read certificate info into the clientInformation field */
+    private void readPeerCertInfo(X509 *cert)
+    {
+        X509_NAME* name = X509_get_subject_name(cert);
+
+        int c = X509_NAME_entry_count(name);
+        for (int i = 0; i < c; i++) {
```
s-ludwig: Just a little bit shorter and more D-ish would be `foreach (i; 0 .. c)`

source/vibe/stream/openssl.d
```diff
@@ -266,6 +288,11 @@ final class OpenSSLStream : SSLStream {
 			throw m_exceptions[0];
 		}
 	}
+
+    @property SSLCertificateInformation peerCertificate()
+    {
+        return m_peerCertificate;
+    }
```
s-ludwig: Can this be `const`? Or does `DictionaryList` complain about the copy then?

s-ludwig:
Looks good to merge, except that lines are indented using spaces instead of tabs. Depending on what editor you use, there might be an [EditorConfig](http://editorconfig.org/) plugin, which is very useful for such formatting issues.

rix0rrr:
Changed spaces to tabs

s-ludwig:
Thanks! Merging.

SerialVelocity:
@s-ludwig @rix0rrr Is there a D client library which allows setting the client certificate? (WebSocket 2-way SSL is actually what I'm looking for)

PR #966, opened 2015-01-27
Feature/i18n enhancement

ncoe:
Extending i18n support to include plural translations.


s-ludwig:
Added this rebased and with a simplified diff to a local branch.

s-ludwig:
See #1290.

PR #969, opened 2015-01-28
REST enhancement #1.2: Implementation of @queryParam & @bodyParam

Geod24:
Okay, this had been stalling in my computer for a while now, time to push it.
Those are the complements of `@headerParam`. Since most of the work had been done in #926 , this is pretty straightforward.


Geod24:
Ping @s-ludwig .

We'll need to issue a new release once DMD 2.067 is out (beta should be available by the 15th, released by the 1st of March IIRC), and I'd really like this to complete `@headerParam`.

PR #970, opened 2015-01-28
Get rid of missing SSL context errors

rix0rrr:
This occurs a lot when using client cetificates, and then connections
are dropped a couple of times before they are succesfully estbalished:

    [0B573E01:0B613C01 2015.01.28 17:11:07.064 WRN] Handling of
    connection failed: Failed to accept SSL tunnel: error:140D9115:SSL
    routines:SSL_GET_PREV_SESSION:session id context uninitialized
    (336433429)

Assigning the local hostname as a context ID fixes this error.

As per OpenSSL documentation:

https://www.openssl.org/docs/ssl/SSL_CTX_set_session_id_context.html

### DIFF
--- a/source/vibe/stream/openssl.d
+++ b/source/vibe/stream/openssl.d
@@ -322,7 +322,7 @@ final class OpenSSLContext : SSLContext {
 		if (kind == SSLContextKind.server) {
 			setDHParams();
 			static if (haveECDH) setECDHCurve();
-			setContextID();
+			guessSessionIDContext();
 		}
 
 		setCipherList();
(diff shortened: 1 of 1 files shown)


schuetzm:
Can't comment on the content, but your indentation is off (see the diff view https://github.com/rejectedsoftware/vibe.d/pull/970/files).

rix0rrr:
Updated indentation.

source/vibe/stream/openssl.d
```diff
@@ -469,6 +471,17 @@ final class OpenSSLContext : SSLContext {
 			SSL_CTX_set_cipher_list(m_ctx, toStringz(list));
 	}
 
+	/** Make up a context ID to assign to the SSL context.
+
+		This is very required when doing client cert authentication,
+		otherwise many connections will go aborted.
+	*/
+	void setContextID()
```
s-ludwig: Since it is called in the constructor already, does this method have to be public? If yes, I'd change the name to `setSessionIDContext` to match the semantics more closely. What I didn't understand yet when reading the documentation of `SSL_CTX_set_session_id_context` is if using the host name as the context name is the right choice. From the description it sounds like it should rather be an application specific string, which would have to be passed as a parameter.

rix0rrr: I wanted to pick a sane default so it will Just Work if you don't care about the particular value. It has to be unique within a group of servers behind a load balancer, so the protocol can tell it doesn't have a cached session on that particular server and do the handshake again.

BUT, it also has to be configurable for client code in case someone wants to go to the trouble to configure a session pool. Apparently, "tickets" will be better in any case, but not necessarily supported by every client and adding a lot more complexity. 

You're right about the name change, will fix.

rix0rrr: And you know what, I totally forgot what I actually implemented.

You're right, this function doesn't need to be public.

s-ludwig:
Okay, thanks. Merging.

PR #972, opened 2015-01-30
Get Vibe.d ready for 2.067

Geod24:
This add a bit of nothrowness in utilities (probably more to come), re-enable an unittest, and disable the new std.concurrency support when building with master (because it is simply not implemented).
There is still a LOT of nothrow to put here, and I think it'll be better to start with a small diff.

Btw, the 2 overloads of opApply would never be called. Signature selection for `opApply` is not at clever sadly :(

### DIFF
--- a/source/vibe/core/file.d
+++ b/source/vibe/core/file.d
@@ -13,8 +13,11 @@ public import vibe.inet.url;
 import vibe.core.drivers.threadedfile; // temporarily needed tp get mkstemps to work
 import vibe.core.driver;
 
-import std.conv;
-import std.c.stdio;
+static if (__VERSION__ > 2066)
+	import core.stdc.stdio;
+else
+	import std.c.stdio;
+
 import std.datetime;
 import std.exception;
 import std.file;
@@ -136,6 +139,7 @@ void writeFileUTF8(Path path, string contents)
 FileStream createTempFile(string suffix = null)
 {
 	version(Windows){
+		import std.conv : to;
 		char[L_tmpnam] tmp;
 		tmpnam(tmp.ptr);
 		auto tmpname = to!string(tmp.ptr);
--- a/source/vibe/core/file.d
+++ b/source/vibe/core/file.d
@@ -214,19 +214,24 @@ void removeFile(Path path)
 	removeFile(path.toNativeString());
 }
 /// ditto
-void removeFile(string path) {
+void removeFile(string path)
+{
 	std.file.remove(path);
 }
 
 /**
 	Checks if a file exists
 */
-bool existsFile(Path path) {
+bool existsFile(Path path) nothrow
+{
 	return existsFile(path.toNativeString());
 }
 /// ditto
-bool existsFile(string path)
+bool existsFile(string path) nothrow
 {
+	// This was *annotated* nothrow in 2.067.
+	static if (__VERSION__ < 2067)
+		scope(failure) assert(0, "Error: existsFile should never throw");
 	return std.file.exists(path);
 }
 
--- a/source/vibe/http/common.d
+++ b/source/vibe/http/common.d
@@ -601,7 +601,7 @@ struct CookieValueMap {
 
 	inout(string)* opBinaryRight(string op)(string name) inout if(op == "in")
 	{
-		foreach(c; m_entries)
+		foreach(ref c; m_entries)
 			if( c.name == name )
 				return &c.value;
 		return null;
--- a/source/vibe/http/common.d
+++ b/source/vibe/http/common.d
@@ -601,9 +601,15 @@ struct CookieValueMap {
 
 	inout(string)* opBinaryRight(string op)(string name) inout if(op == "in")
 	{
-		foreach(ref c; m_entries)
-			if( c.name == name )
-				return &c.value;
+		static if (__VERSION__ < 2066) {
+			foreach(c; m_entries)
+				if( c.name == name )
+					return &c.value;
+		} else {
+			foreach(ref c; m_entries)
+				if( c.name == name )
+					return &c.value;
+		}
 		return null;
 	}
 }
--- a/source/vibe/inet/path.d
+++ b/source/vibe/inet/path.d
@@ -94,17 +94,17 @@ struct Path {
 	}
 	
 	/// Converts the Path object to a native path string (backslash as path separator on Windows).
-	string toNativeString()
+	string toNativeString() nothrow
 	const {
 		Appender!string ret;
 		
 		// for absolute unix paths start with /
-		version(Posix) { if(absolute) ret.put('/'); }
+		version(Posix) { if (m_absolute) ret.put('/'); }
 		
 		foreach( i, f; m_nodes ){
 			version(Windows) { if( i > 0 ) ret.put('\\'); }
 			version(Posix) { if( i > 0 ) ret.put('/'); }
-			else { enforce("Unsupported OS"); }
+			else { static assert(0, "Unsupported OS"); }
 			ret.put(f.toString());
 		}
 		
@@ -366,7 +366,7 @@ struct PathEntry {
 		m_name = str;
 	}
 	
-	string toString() const { return m_name; }
+	string toString() const nothrow { return m_name; }
 
 	Path opBinary(string OP)(PathEntry rhs) const if( OP == "~" ) { return Path(cast(immutable)[this, rhs], false); }
 	
--- a/source/vibe/core/concurrency.d
+++ b/source/vibe/core/concurrency.d
@@ -320,7 +320,7 @@ template Isolated(T)
 
 
 // unit tests fails with DMD 2.064 due to some cyclic import regression
-version (none) unittest
+unittest
 {
 	static class CE {}
 	static struct SE {}
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -29,6 +29,7 @@ import core.sync.mutex;
 import core.stdc.stdlib;
 import core.thread;
 
+alias TaskEventCb = void function(TaskEvent, Task) nothrow;
 
 version(Posix)
 {
@@ -677,7 +678,7 @@ void lowerPrivileges()
 	analyze the life time of tasks, including task switches. Note that
 	the callback will only be called for debug builds.
 */
-void setTaskEventCallback(void function(TaskEvent, Task) func)
+void setTaskEventCallback(TaskEventCb func)
 {
 	debug s_taskEventCallback = func;
 }
@@ -1210,7 +1211,7 @@ private {
 	__gshared ThreadContext[] st_threads;
 	__gshared TaskFuncInfo[] st_workerTasks;
 	__gshared Condition st_threadShutdownCondition;
-	__gshared debug void function(TaskEvent, Task) s_taskEventCallback;
+	__gshared debug TaskEventCb s_taskEventCallback;
 	shared bool st_term = false;
 
 	bool s_exitEventLoop = false;
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1074,33 +1074,40 @@ private class VibeDriverCore : DriverCore {
 		}
 	}
 
-	void resumeTask(Task task, Exception event_exception = null)
+	void resumeTask(Task task, Exception event_exception = null) nothrow
 	{
 		resumeTask(task, event_exception, false);
 	}
 
-	void resumeTask(Task task, Exception event_exception, bool initial_resume)
+	void resumeTask(Task task, Exception event_exception, bool initial_resume) nothrow
 	{
 		assert(initial_resume || task.running, "Resuming terminated task.");
 		resumeCoreTask(cast(CoreTask)task.fiber, event_exception);
 	}
 
-	void resumeCoreTask(CoreTask ctask, Exception event_exception = null)
+	void resumeCoreTask(CoreTask ctask, Exception event_exception = null) nothrow
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		assert(ctask.thread is Thread.getThis(), "Resuming task in foreign thread.");
-		assert(ctask.state == Fiber.State.HOLD, "Resuming fiber that is " ~ to!string(ctask.state));
+		assert(ctask.state == Fiber.State.HOLD, "Resuming fiber that is not on HOLD");
 
 		if( event_exception ){
 			extrap();
 			ctask.m_exception = event_exception;
 		}
 
-		static if (__VERSION__ > 2067)
+		static if (__VERSION__ > 2066)
 			auto uncaught_exception = ctask.call!(Fiber.Rethrow.no)();
 		else
 			auto uncaught_exception = ctask.call(false);
 		if (auto th = cast(Throwable)uncaught_exception) {
 			extrap();
+
 			assert(ctask.state == Fiber.State.TERM);
 			logError("Task terminated with unhandled exception: %s", th.msg);
 			logDebug("Full error: %s", th.toString().sanitize);
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1095,7 +1095,10 @@ private class VibeDriverCore : DriverCore {
 			ctask.m_exception = event_exception;
 		}
 
-		auto uncaught_exception = ctask.call(false);
+		static if (__VERSION__ > 2067)
+			auto uncaught_exception = ctask.call!(Fiber.Rethrow.no)();
+		else
+			auto uncaught_exception = ctask.call(false);
 		if (auto th = cast(Throwable)uncaught_exception) {
 			extrap();
 			assert(ctask.state == Fiber.State.TERM);
--- a/source/vibe/core/driver.d
+++ b/source/vibe/core/driver.d
@@ -165,7 +165,7 @@ interface EventDriver {
 interface DriverCore {
 	@property void eventException(Exception e);
 	void yieldForEvent();
-	void resumeTask(Task f, Exception event_exception = null);
+	void resumeTask(Task f, Exception event_exception = null) nothrow;
 	void notifyIdle();
 }
 
--- a/source/vibe/core/driver.d
+++ b/source/vibe/core/driver.d
@@ -28,14 +28,14 @@ version (VibeUseNativeDriverType) {
 /**
 	Returns the active event driver
 */
-StoredEventDriver getEventDriver(bool ignore_unloaded = false)
+StoredEventDriver getEventDriver(bool ignore_unloaded = false) nothrow
 {
 	assert(ignore_unloaded || s_driver !is null, "No event driver loaded. Did the vibe.core.core module constructor run?");
 	return s_driver;
 }
 
 /// private
-package void setupEventDriver(DriverCore core_)
+package void setupEventDriver(DriverCore core_) nothrow
 {
 	version (VibeUseNativeDriverType) {}
 	else import vibe.core.drivers.native;
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -79,15 +79,15 @@ final class Libevent2Driver : EventDriver {
 
 	this(DriverCore core) nothrow
 	{
-		debug m_ownerThread = Thread.getThis();
-		m_core = core;
-		s_driverCore = core;
-
 		// In 2067, synchronized statements where annotated nothrow.
 		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
 		// However, they were "logically" nothrow before.
 		static if (__VERSION__ <= 2066)
 			scope (failure) assert(0, "Internal error: function should be nothrow");
+
+		debug m_ownerThread = Thread.getThis();
+		m_core = core;
+		s_driverCore = core;
 
 		synchronized if (!s_threadObjectsMutex) {
 			s_threadObjectsMutex = new Mutex;
@@ -676,14 +676,14 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 
 	void acquire() nothrow
 	{
-		auto task = Task.getThis();
-		auto thread = task == Task() ? Thread.getThis() : task.thread;
-
 		// In 2067, synchronized statements where annotated nothrow.
 		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
 		// However, they were "logically" nothrow before.
 		static if (__VERSION__ <= 2066)
 			scope (failure) assert(0, "Internal error: function should be nothrow");
+
+		auto task = Task.getThis();
+		auto thread = task == Task() ? Thread.getThis() : task.thread;
 
 		synchronized (m_mutex) {
 			if (thread !in m_waiters) {
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -577,7 +577,7 @@ private class Libevent2Object {
 		m_driver.unregisterObject(this);
 	}
 
-	protected void onThreadShutdown() {}
+	protected void onThreadShutdown() nothrow {}
 }
 
 /// private
@@ -668,10 +668,16 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 		return ec;
 	}
 
-	void acquire()
+	void acquire() nothrow
 	{
 		auto task = Task.getThis();
 		auto thread = task == Task() ? Thread.getThis() : task.thread;
+
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
 
 		synchronized (m_mutex) {
 			if (thread !in m_waiters) {
@@ -701,8 +707,14 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 		}
 	}
 
-	bool amOwner()
+	bool amOwner() nothrow
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		auto self = Task.getThis();
 		if (self == Task()) return false;
 		synchronized (m_mutex) {
@@ -715,6 +727,12 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 
 	protected override void onThreadShutdown()
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		auto thr = Thread.getThis();
 		synchronized (m_mutex) {
 			if (thr in m_waiters) {
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -611,6 +611,12 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 
 	void emit()
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		atomicOp!"+="(m_emitCount, 1);
 		synchronized (m_mutex) {
 			foreach (ref sl; m_waiters)
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -611,7 +611,7 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 
 	~this()
 	{
-		foreach (ts; m_waiters)
+		foreach (ref m_waiters.Value ts; m_waiters)
 			event_free(ts.event);
 	}
 
@@ -625,7 +625,7 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 
 		atomicOp!"+="(m_emitCount, 1);
 		synchronized (m_mutex) {
-			foreach (ref sl; m_waiters)
+			foreach (ref m_waiters.Value sl; m_waiters)
 				event_active(sl.event, 0, 0);
 		}
 	}
@@ -701,8 +701,14 @@ final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 		}
 	}
 
-	void release()
+	void release() nothrow
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		auto self = Task.getThis();
 		if (self == Task()) return;
 
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -171,8 +171,8 @@ final class Libevent2Driver : EventDriver {
 		s_alreadyDeinitialized = true;
 	}
 
-	@property event_base* eventLoop() { return m_eventLoop; }
-	@property evdns_base* dnsEngine() { return m_dnsBase; }
+	@property event_base* eventLoop() nothrow { return m_eventLoop; }
+	@property evdns_base* dnsEngine() nothrow { return m_dnsBase; }
 
 	int runEventLoop()
 	{
--- a/source/vibe/core/drivers/libevent2.d
+++ b/source/vibe/core/drivers/libevent2.d
@@ -77,11 +77,17 @@ final class Libevent2Driver : EventDriver {
 		size_t m_addressInfoCacheLength = 0;
 	}
 
-	this(DriverCore core)
+	this(DriverCore core) nothrow
 	{
 		debug m_ownerThread = Thread.getThis();
 		m_core = core;
 		s_driverCore = core;
+
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
 
 		synchronized if (!s_threadObjectsMutex) {
 			s_threadObjectsMutex = new Mutex;
@@ -1195,17 +1201,17 @@ private {
 	bool s_alreadyDeinitialized = false;
 }
 
-package event_base* getThreadLibeventEventLoop()
+package event_base* getThreadLibeventEventLoop() nothrow
 {
 	return s_eventLoop;
 }
 
-package DriverCore getThreadLibeventDriverCore()
+package DriverCore getThreadLibeventDriverCore() nothrow
 {
 	return s_driverCore;
 }
 
-private int getLastSocketError()
+private int getLastSocketError() nothrow
 {
 	version(Windows) return WSAGetLastError();
 	else {
--- a/source/vibe/core/net.d
+++ b/source/vibe/core/net.d
@@ -166,6 +166,7 @@ struct NetworkAddress {
 		import std.array : appender;
 		import std.string : format;
 		import std.format : formattedWrite;
+		ubyte[2] _dummy = void; // Workaround for DMD regression in master
 
 		switch (this.family) {
 			default: assert(false, "toAddressString() called for invalid address family.");
@@ -178,7 +179,8 @@ struct NetworkAddress {
 				ret.reserve(40);
 				foreach (i; 0 .. 8) {
 					if (i > 0) ret.put(':');
-					ret.formattedWrite("%x", bigEndianToNative!ushort(cast(ubyte[2])ip[i*2 .. i*2+2]));
+					_dummy[] = ip[i*2 .. i*2+2];
+					ret.formattedWrite("%x", bigEndianToNative!ushort(_dummy));
 				}
 				return ret.data;
 		}
--- a/source/vibe/core/task.d
+++ b/source/vibe/core/task.d
@@ -34,18 +34,24 @@ struct Task {
 		}
 	}
 
-	private this(TaskFiber fiber, size_t task_counter)
+	private this(TaskFiber fiber, size_t task_counter) nothrow
 	{
 		m_fiber = cast(shared)fiber;
 		m_taskCounter = task_counter;
 	}
 
-	this(in Task other) { m_fiber = cast(shared(TaskFiber))other.m_fiber; m_taskCounter = other.m_taskCounter; }
+	this(in Task other) nothrow { m_fiber = cast(shared(TaskFiber))other.m_fiber; m_taskCounter = other.m_taskCounter; }
 
 	/** Returns the Task instance belonging to the calling task.
 	*/
-	static Task getThis()
+	static Task getThis() nothrow
 	{
+		// In 2067, synchronized statements where annotated nothrow.
+		// DMD#4115, Druntime#1013, Druntime#1021, Phobos#2704
+		// However, they were "logically" nothrow before.
+		static if (__VERSION__ <= 2066)
+			scope (failure) assert(0, "Internal error: function should be nothrow");
+
 		auto fiber = Fiber.getThis();
 		if (!fiber) return Task.init;
 		auto tfiber = cast(TaskFiber)fiber;
--- a/source/vibe/crypto/passwordhash.d
+++ b/source/vibe/crypto/passwordhash.d
@@ -62,6 +62,7 @@ string generateSimplePasswordHash(string password, string additional_salt = null
 deprecated("This function is considered insecure and will be removed. The DUB packages dauth or scrypt may be suitable alternatives.")
 bool testSimplePasswordHash(string hashstring, string password, string additional_salt = null)
 {
+	import std.string : format;
 	ubyte[] upass = Base64.decode(hashstring);
 	enforce(upass.length == 20, format("Invalid binary password hash length: %s", upass.length));
 	auto salt = upass[0 .. 4];
--- a/source/vibe/internal/meta/traits.d
+++ b/source/vibe/internal/meta/traits.d
@@ -349,3 +349,24 @@ bool areConvertibleTo(alias TYPES, alias TARGET_TYPES)()
 			return false;
 	return true;
 }
+
+/// Test if the type $(D DG) is a correct delegate for an opApply where the
+/// key/index is of type $(D TKEY) and the value of type $(D TVALUE).
+template isOpApplyDg(DG, TKEY, TVALUE) {
+	import std.traits;
+	static if (is(DG == delegate) && is(ReturnType!DG : int)) {
+		private alias PTT = ParameterTypeTuple!(DG);
+		private alias PSCT = ParameterStorageClassTuple!(DG);
+		private alias STC = ParameterStorageClass;
+		// Just a value
+		static if (PTT.length == 1) {
+			enum isOpApplyDg = (is(PTT[0] == TVALUE) && PSCT[0] == STC.ref_);
+		} else static if (PTT.length == 2) {
+			enum isOpApplyDg = (is(PTT[0] == TKEY) && PSCT[0] == STC.ref_)
+				&& (is(PTT[1] == TKEY) && PSCT[1] == STC.ref_);
+		} else
+			enum isOpApplyDg = false;
+	} else {
+		enum isOpApplyDg = false;
+	}
+}
--- a/source/vibe/internal/newconcurrency.d
+++ b/source/vibe/internal/newconcurrency.d
@@ -1,7 +1,7 @@
 /// Small helper module to determine if the new std.concurrency interop features are present
 module vibe.internal.newconcurrency;
 
-static if (__VERSION__ >= 2066) {
+static if (__VERSION__ >= 2066 && false) {
 	import std.concurrency;
 	static if (is(std.concurrency.Scheduler)) enum bool newStdConcurrency = true;
 	else enum bool newStdConcurrency = false;
--- a/source/vibe/utils/hashmap.d
+++ b/source/vibe/utils/hashmap.d
@@ -81,6 +81,14 @@ struct HashMap(Key, Value, Traits = DefaultHashMapTraits!Key)
 		return m_table[idx].value;
 	}
 
+	/// Workaround #12647
+	package Value getNothrow(Key key, Value default_value = Value.init)
+	{
+		auto idx = findIndex(key);
+		if (idx == size_t.max) return default_value;
+		return m_table[idx].value;
+	}
+
 	static if (!is(typeof({ Value v; const(Value) vc; v = vc; }))) {
 		const(Value) get(Key key, lazy const(Value) default_value = Value.init)
 		{
@@ -175,7 +183,7 @@ struct HashMap(Key, Value, Traits = DefaultHashMapTraits!Key)
 		resize(newcap);
 	}
 
-	private void resize(size_t new_size)
+	private void resize(size_t new_size) nothrow
 	{
 		assert(!m_resizing);
 		m_resizing = true;
--- a/source/vibe/utils/hashmap.d
+++ b/source/vibe/utils/hashmap.d
@@ -34,7 +34,7 @@ struct HashMap(Key, Value, Traits = DefaultHashMapTraits!Key)
 		TableEntry[] m_table; // NOTE: capacity is always POT
 		size_t m_length;
 		Allocator m_allocator;
-		hash_t delegate(Key) m_hasher;
+		hash_t delegate(Key) nothrow m_hasher;
 		bool m_resizing;
 	}
 
@@ -123,7 +123,7 @@ struct HashMap(Key, Value, Traits = DefaultHashMapTraits!Key)
 		return &m_table[idx].value;
 	}
 
-	int opApply(int delegate(ref Value) nothrow del) nothrow
+	int opApply(int delegate(ref Value) del)
 	{
 		foreach (i; 0 .. m_table.length)
 			if (!Traits.equals(m_table[i].key, Traits.clearValue))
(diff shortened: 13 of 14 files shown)


Geod24:
@s-ludwig 
FYI: http://forum.dlang.org/thread/magv9a$2641$1@digitalmars.com

Geod24:
Ping @s-ludwig .

MartinNowak:
Thanks for doing this, LGTM.

Geod24:
I'll try to put all the nothrowness ASAP. You're still planning for a beta on the 15th and a release on the 1st ?

Geod24:
Moar fixes.
My goal now is to make ManualEvent interface nothrow. I'm halfway through this.
@MartinNowak : Did you miss _d_criticalEnter / _d_criticalExit when fixing https://issues.dlang.org/show_bug.cgi?id=11216 ? Because as of today, they aren't nothrow:

````
source/vibe/core/drivers/libevent2.d(92): Error: '_d_criticalenter' is not nothrow
source/vibe/core/drivers/libevent2.d(92): Error: '_d_criticalexit' is not nothrow
````

File: source/vibe/http/common.d [hunk 1]
```diff
@@ -601,9 +601,15 @@ struct CookieValueMap {
...
-				return &c.value;
+		static if (__VERSION__ < 2066) {
+			foreach(c; m_entries)
+				if( c.name == name )
+					return &c.value;
+		} else {
+			foreach(ref c; m_entries)
+				if( c.name == name )
+					return &c.value;
+		}
```
MartinNowak: What's the problem here? If `Cookie` is a class it should work without ref.

Geod24: https://github.com/Geod24/vibe.d/blob/fixes-for-master/source/vibe/http/common.d#L532

Cookie is a struct.

MartinNowak: Ah OK, was looking at the wrong Cookie.
http://vibed.org/api/vibe.http.common/Cookie

Geod24:
@MartinNowak : I'm not an expert with `std.conv.emplace`, but should the following work ?

````
module test;
import std.conv;

struct Test {}

void main() nothrow {
        byte[4096] buff;
        auto ptr = emplace!Test(buff);
}
````

**Comment by Geod24** [comment]
Status report: I just hit https://issues.dlang.org/show_bug.cgi?id=12647 while marking `vibe.utils.memory` nothrow. Will work on a workaround, for the time being. It's a bit of a PITA that everything was made nothrow in one single release.

**Comment by Geod24** [comment]
Updated with a nothrow `vibe.utils.memory`.
SA: https://issues.dlang.org/show_bug.cgi?id=14194

**Comment by MartinNowak** [comment]
> It's a bit of a PITA that everything was made nothrow in one single release.

Yeah, sorry about that, but there was no clear path to deprecate IMonitor and it only affects very few libraries that inherit from core.sync.*.

File: source/vibe/core/file.d [hunk 2]
```diff
@@ -13,8 +13,11 @@ public import vibe.inet.url;
 import vibe.core.drivers.threadedfile; // temporarily needed tp get mkstemps to work
 import vibe.core.driver;
 
-import std.conv;
-import std.c.stdio;
+static if (__VERSION__ > 2066)
```
sinkuu: `static if` seems unnecessary.  `core.stdc` has existed since D 2.021 (according to changelog).

source/vibe/http/common.d
s-ludwig: But I'd say the first `static if` branch should go completely. No idea if it actually every worked - somehow...

Geod24: So do I. But it fails on 2.065.

s-ludwig: Hum.. ok. But then let's at least change it to this, so that the potential memory corruption bug is gone:
```
		foreach (ref c; m_entries)
			if (c.name == name) {
				static if (__VERSION__ < 2066)
					return cast(inout(string)*)&c.value;
				else
					return &c.value;
			}
```

I can also do that after the merge, though!

source/vibe/inet/path.d
```diff
@@ -94,17 +94,17 @@ struct Path {
...
 		
 		// for absolute unix paths start with /
-		version(Posix) { if(absolute) ret.put('/'); }
+		version(Posix) { if (m_absolute) ret.put('/'); }
 		
 		foreach( i, f; m_nodes ){
 			version(Windows) { if( i > 0 ) ret.put('\\'); }
 			version(Posix) { if( i > 0 ) ret.put('/'); }
-			else { enforce("Unsupported OS"); }
+			else { static assert(0, "Unsupported OS"); }
```
s-ludwig: Looks like this will fail compilation on Windows. Either `else version(Posix)` or replacing the enforce with an `assert` should be ok. Not sure why I went with a runtime error at the time...

s-ludwig: Oh damn, I just noticed now how messed up that `enforce` call was...
```
version(Windows) { if( i > 0 ) ret.put('\\'); }
else version(Posix) { if( i > 0 ) ret.put('/'); }
else static assert(false, "Unsupported OS");
```

should be fine, though.

BTW, looks like a good warning opportunity for DMD.

s-ludwig:
BTW, Big thanks for the fixes. I'd probably not have made it on time, as I'm still living with a crazy workload (on top of that, my work laptop'd display just failed... with the 1-year Lenovo warranty ending in January :angry:)

I'll pull and take over the remaining details now.

Geod24:
Thanks for merging ! Although this is not complete - There's still some nothrowness missing, and possibly one regression I need to investigate -, that's a good first step.
I'll keep working on fixing it, but I can't give any time estimate since the changes are quite involved, and I'll be moving to Berlin next saturday.

s-ludwig:
Thanks, and be sure to send in any partial fixes before the DMD release (or your move), I'll try to fix any outstanding issues until then, so that we get a clean parallel release. I've also enabled Travis testing 2.067-b2 now.

s-ludwig:
> since the changes are quite involved

BTW, I tried to make `HashMap` `@nogc`/`nothrow` as far as possible a few days ago, but had big issues, because all of the hashing functionality in object.d is not `@nogc`/`nothrow`. This seems to be the most serious blocker for making `ManualEvent` `@nogc`. I'll see what I can do there (probably using some function pointer casting).

s-ludwig:
Unfortunately the `HashMap.opApply` partial fix in 351b247cf28b45bbbd791f6991379d5583e201a8 broke foreach type inference. I'll search for a different solution.

Geod24:
Because of https://issues.dlang.org/show_bug.cgi?id=12647 it won't be possible to have a no `lazy` argument that's `nothrow`.
Because of https://issues.dlang.org/show_bug.cgi?id=14196 we lost type inference in foreach for HashMap.

Actually  there are cases where you do not want get / opApply to be `nothrow` / `@nogc`, like when the lazy parameter would allocate / throw, same goes for the delegate you apply.

If everything went as expected, you should be able (almost?) everything else as `nothrow`. I'm currently trying to make `ManualEvent.wait` `nothrow`.

Geod24:
2015-02-21 18:53 GMT+01:00 Sönke Ludwig <notifications@github.com>:

> Unfortunately the HashMap.opApply partial fix in 351b247
> <https://github.com/rejectedsoftware/vibe.d/commit/351b247cf28b45bbbd791f6991379d5583e201a8>
> broke foreach type inference. I'll search for a different solution.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rejectedsoftware/vibe.d/pull/972#issuecomment-75383318>
> .
>
I couldn't find one suitable. The compiler won't like having overrides that
differs only by their nothrowness:
test.d(21): Error: arr.opApply matches more than one declaration:
test.d(3):     @system int(int delegate(string v) del)
and:
test.d(9):     nothrow @system int(int delegate(string v) nothrow del)

s-ludwig:
Right, I remember that I've hit the same issues. In fact it was stupid to not review this pull request before starting that work, but it was started for an unrelated reason initially. I'll just commit my changes to `HashMap` (keeping `opApply` as it is for now) and refrain from further changes to not interfere with you too much.

s-ludwig:
> I couldn't find one suitable. The compiler won't like having overrides that
differs only by their nothrowness:

Same for `@nogc`, unfortunately.

MartinNowak:
> BTW, I tried to make HashMap @nogc/nothrow as far as possible a few days ago, but had big issues, because all of the hashing functionality in object.d is not @nogc/nothrow.

We have a new templated hashOf method in druntime in core.internal.hash. It's meant to replace the TypeInfo hash.

s-ludwig:
Yeah, I had used that, but it's in `core.internal` and doesn't work for GDC and LDC. But on top of that it is also neither `nothrow`, nor `@nogc`, unfortunately.

etcimon:
I feel like pushing `@nogc` in druntime so early is creating a lot of problems. Even Phobos is still missing a lot of `@nogc` declarations, because the `std.containers.array` uses `malloc` and is not `@nogc` itself.

Also, those areas of the code that use the GC but manually destroys the objects should be immune to it. It would be nice to have a `@nogc(bool force = false)` to get it out of the way

MartinNowak:
It's templates and should be inferred, might recheck that.

s-ludwig:
Looking at it, I think the reason for non-`@nogc` is `bytesHash` and `seedHash`. Nothrow'ness seems to have been fixed since I last checked, though.

Geod24:
Yeah `HashMap` should correctly be infered `nothrow` when needed, except for `get` because of the lazy param. We needed `utils.memory` to be nothrow (which in turns needed `HashMap` to be `nothrow` ;) )
I didn't bother about `@nogc` through.

s-ludwig:
I think we'll just have a dejà vu once `@nogc` gets added to `Mutex.lock`. That is, assuming that `Mutex` will stay non-`final`...

MartinNowak:
How about using a templates DG?
```d
void opApply(DG)(scope DG dg)
```

Geod24:
@MartinNowak: See my response earlier mentionning https://issues.dlang.org/show_bug.cgi?id=14196

s-ludwig:
`scope` is a good point, though.

s-ludwig:
2fb019c012eba300b614b30ea7932cd35a910585

Geod24:
@s-ludwig : Do you have any document describing the internals of Vibe.d around ? ATM `ManualEvent.wait` could not be made `nothrow` because `DriverCore` isn't `nothrow`. Now I'm not 100% sure what should be done at this point, so any doc on the core design would save me some time :)

s-ludwig:
Nothing of interest for this particular area, unfortunately. But if you mean `DriverCore.yieldForEvent`, that definitely *can* throw. We'd have to use either `scope(failure) assert(0);` (because we know that only `ManualEvent` itself should ever resume the task and does so without passing an exception\*) or use a manual try-catch-block and handle the exception somehow  differently.

\* `Task.interrupt()` wouldn't play well with this, as it also resumes the task with an exception. So actually mutexes in vibe.d are never nothrow... I didn't realize that earlier. This is bad news. Either we have to document that it is never legal to use `Task.interrupt` together with a `TaskMutex` (highly fragile), or deprecate `Task.interrupt()` altogether (which would be a shame).

The third option would be to base `synchronized` and the object monitor on an abstraction that can work for both, `nothrow` and non-`nothrow`, mutexes. Not sure how that would work acceptably for the object monitor.

MartinNowak:
> The third option would be to base synchronized and the object monitor on an abstraction that can work for both, nothrow and non-nothrow, mutexes. Not sure how that would work acceptably for the object monitor.

Indeed, it would be much nicer if synchronized were a simple lowering and used duck typing instead of relying on the interface. The runtime could easily handle Object and their monitors as well.

MartinNowak:
I'll try to add the lowering before we release 2.067.

MartinNowak:
The lowering wouldn't solve the problem because there is set monitor and setSameMutex. So eventually we need to keep the IMonitor interface and can only turn all sync stuff into nothrow.
It's a bit unfortunate that Exceptions are used to communicate with tasks, instead of message passing, thereby turning all evented methods into throwing functions. Even more so because the Exception isn't semantically related to `TaskMutex.lock`.

MartinNowak:
For a solution right now, we could introduce the notion of pending exceptions. Some evented functions are synchronization points that do might raise such exceptions.
Then there could be a yieldForEventNothrow that appends a possible exception to any pending.

etcimon:
How about attaching an error handling delegate that returns true if the code can continue?

`alias ErrorHandler = bool delegate(Exception e) nothrow;`
Just call the ManualEvent with e.g. a `void emit(ErrorHandler eh = null) nothrow`

Then you can use that in the catch blocks like this:

`} catch (Exception e) if (!eh(e)) assert(false);`

That way, every exception is handled correctly and the function stays nothrow.

s-ludwig:
> It's a bit unfortunate that Exceptions are used to communicate with tasks, instead of message passing, thereby turning all evented methods into throwing functions.

The message passing system is used for regular inter-task communication. `interrupt()` is there to have a way to be able to cancel tasks in a safe way - including proper stack cleanup. What else could be used for that instead of exceptions? Messages don't help there because they naturally can't be received while the task is blocked for anything else than receiving messages. I mean I agree, it's a bit unfortunate that it makes things non-`nothrow` that could otherwise be. However, all I/O functions can throw anyway, so it's not really an issue in the context of vibe.d.

> Even more so because the Exception isn't semantically related to `TaskMutex.lock`.

`InterruptException` in particular can be thrown by any function that performs a blocking operation. We could of course add some kind of workaround for just `TaskMutex.lock`, but that would leave any locked tasks uninterruptible.

s-ludwig:
>Then you can use that in the catch blocks like this:
>
>    } catch (Exception e) if (!eh(e)) assert(false);
>
>That way, every exception is handled correctly and the function stays nothrow.

But what would you do with that exception? The only exception type that is supposed to arrive there is `InterruptException`. And since you can't throw from the `ErrorHandler`, you can only ignore it anyway.

etcimon:
> But what would you do with that exception? The only exception type that is supposed to arrive there is InterruptException. And since you can't throw from the ErrorHandler, you can only ignore it anyway.

You could save the exception and then catch `Throwable` later but handle the saved exception instead

s-ludwig:
Any deferred handling would still mean that `TaskMutex` becomes uninterruptible. Throwing the exception late is of course better than never, but it makes `Task.interrupt()` a much less useful tool.

etcimon:
> Any deferred handling would still mean that TaskMutex becomes uninterruptible. Throwing the exception late is of course better than never, but it makes Task.interrupt() a much less useful tool.

Would it be possible to use `assert(false);` in `Task.interrupt()`?

s-ludwig:
That would leave the stack dirty. The exception handlers are needed for proper RAII/scope cleanup.

etcimon:
So, this is all because `synchronized` needs to work inside `nothrow` methods? It looks like we now need a `InterruptibleTaskMutex` that doesn't use `synchronized`.

Unless you wrap synchronized with a `catch(Throwable)` all the time, but that's not as nice of an API.

etcimon:
In my opinion, better just drop `synchronized` statements altogether for tasks and use a `ScopedLock` like in C++. The mutexes from D can be accessed by composition instead

s-ludwig:
The main issue is not `synchronized` (that would be easy to get working with and without `nothrow`-style mutexes), but `setSameMutex`, which sets the object monitor. This is currently employed in `vibe.core.concurrency` to provide the `lock()` primitive. This could be changed to use a separately defined mutex, but we'd at least need some time for a proper deprecation phase, because that also requires changes in user code.

etcimon:
> we'd at least need some time for a proper deprecation phase, because that also requires changes in user code.

I changed my pull to use a new mutex type, not sure how the deprecation would work. So far, the only change required is changing `synchronized(mutex)` to `with(mutex.locked)`, also the monitor won't work through inheritance. It's simply breaking.

If Phobos calling itself *Standard* doesn't care about breaking code every version increment, why would vibe.d care about it?

s-ludwig:
To defend Phobos/Druntime, both actually care about not breaking code. It's just sometimes that breaking changes get into a release unnoticed. I still hope that we can find a solution that defers the breaking change in Druntime until vibe.d had the chance to do a proper transition to whatever the new solution is.

MartinNowak:
I'm a little late apparently. https://github.com/rejectedsoftware/vibe.d/pull/992/files#r25751883
Turns out that TaskMutex.lock is really the only D code that appears to have a reason throwing.
I'd still argue, that the best solution to this is to defer the throwing.
There aren't too many good reasons to interrupt acquiring a mutex, other than desperately trying to resolve a dead-lock.

MartinNowak:
> with(mutex.locked)

Are you relying on the destructor here to release the mutex? The with statement was only fixed recently, so this would't work with 2.066.

s-ludwig:
> I'm a little late apparently. https://github.com/rejectedsoftware/vibe.d/pull/992/files#r25751883

I consider the current code as temporary. My main goal was to have the libasync code fixed and I didn't have the time to discuss the other details before doing the merge (and also didn't want to force Etienne to change the code over and over again). I'd rather just fix the code once as needed when we've reached a conclusion.

I think that the `synchronized` issue should be fixed by allowing both incarnations of `lock`/`unlock`. The object monitor issue on the other hand is something that could be worked around in vibe.d, but it requires a deprecation phase as far as I can see (either a new super class or a duck type solution involving a mixin would have to be introduced, but either way it would be something that the library users would have to use explicitly).

> Turns out that TaskMutex.lock is really the only D code that appears to have a reason throwing.
I'd still argue, that the best solution to this is to defer the throwing.

It's just that there are only two (bad) possibilities how to do that:
1. Immediately return from `lock()` and throw on the next implicit `yield()`. This would introduce a race condition.
2. Still wait for the lock to be taken and then throw on the next `yield()`. This would in many cases not be much different to simply ignoring the interrupt request, since `Task.interrupt()` is usually a tool that is meant for manual user intervention rather than a general control flow mechanism.

> There aren't too many good reasons to interrupt acquiring a mutex, other than desperately trying to resolve a dead-lock.

`TaskMutex` is supposed to be used (or *usable*) for long lock periods, so depending on the situation it could be quite likely that there is a mutex that needs to get interrupted although there is no dead lock. Of course using a worker queue with a `TaskCondition` is usually the much better solution, but there are enough situations where that doesn't work or where it would complicate the code considerably.

Artificial example: two tasks trying to connect to a server, the connection process is protected by a mutex. The user now wants to cancel one of the connection attempts (the one still blocked by the mutex) and has to wait until the other connection attempt, which is against a non-existing IP address, times out.

While this is a constructed example, when mutex protected I/O is involved it's easy to run into similar situations.

etcimon:
The correct new locking syntax would be *scopedLock(mtx)* although using *with* is an idea to replace *synchronized* in the future. I think lock/unlock from the Monitor isn't going to influence vibe anymore with these changes

s-ludwig:
Just that it still affects/breaks code that **uses** vibe.d, so it doesn't help much without being combined with a proper deprecation path.

etcimon:
> Just that it still affects/breaks code that uses vibe.d,

Only if you compile with dmd 2.067. The synchronized statement and object monitor is still available with vibe.d master and dmd <= 2.066.

That code would be broken anyways unless the object monitor becomes less restricted ie. throwable

s-ludwig:
People will obviously want/need upgrade to 2.067 at some point, so it doesn't matter much if their code only gets broken there or for all versions.

MartinNowak:
> The user now wants to cancel one of the connection attempts (the one still blocked by the mutex) and has to wait until the other connection attempt, which is against a non-existing IP address, times out.

What would be the intend of interrupting the connection attempt? Trying to reuse the task more quickly for something else in order to save resources? Something like this?

```d
string get(string url)
{
    synchronized (conn) // <- TaskMutex
        return conn.get(url);
}

void cacher()
{
    while (true)
    {
        receive((string url) {
            try
                db.write(url, get("dlang.org"));
            catch (InterruptException ie)
            {}            
        });
}

auto cacher = runTask(&cacher);

cacher.send("dlang.org");
if (auto v = db.get("dlang.org"))
    cacher.interrupt(); // was already cached
// wants to reuse cacher as quickly as possible
cacher.send("example.com");
```

I'm really having a hard time to come up with a sensible use-case.
The weirdest thing is that `Task.interrupt` synchronously resumes the task. So I can only interrupt a task that is suspended, but without further synchronization on the suspend state, isn't that a race with any other resumption source (e.g. the mutex is now free).

> I think lock/unlock from the Monitor isn't going to influence vibe anymore with these changes

But it makes vibe.d a little more incompatible with the rest of D.

etcimon:
> People will obviously want/need upgrade to 2.067 at some point, so it doesn't matter much if their code only gets broken there or for all versions.

The only other option I see is to make a change at the core of dmd that lets some `FiberException` go through the nothrow interface.

MartinNowak:
> The only other option I see is to make a change at the core of dmd that lets some FiberException go through the nothrow interface.

The compiler has certain assumptions about nothrow code and would no longer generate cleanup code.

MartinNowak:
The only real alternative would be to leave `synchronized (obj)` as throwing.

MartinNowak:
Good news is, we won't release 2.067.0 before we figured out how to solve this in vibe.d.

etcimon:
> What would be the intend of interrupting the connection attempt?

The `ManualEvent` needs to be throwable because that's how libasync (and libevent or every driver) pipes every error code from the OS to the fiber, when it's waiting for I/O with this `ManualEvent`. Things as simple as "Remote peer interrupted the connection". 

See here: https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/core/drivers/libasync.d#L1325

To make fibers completely nothrow in theory, would be stepping back to c-style error return codes and checking them at every call. I'm not sure if that would fix all the issues though

s-ludwig:
> What would be the intend of interrupting the connection attempt? Trying to reuse the task more quickly for something else in order to save resources? Something like this?

I was thinking about something like the cancel button in a browser that cancels loading the page. But it's not so much about the use case - it should always be possible to somehow implement this differently - but more about hidden cases where the `Task.interrupt` mechanism would fail.

> The weirdest thing is that Task.interrupt synchronously resumes the task. So I can only interrupt a task that is suspended, but without further synchronization on the suspend state, isn't that a race with any other resumption source (e.g. the mutex is now free).

`Task.interrupt` currently only works for tasks on the same thread (it fires an assertion otherwise), so there is currently no additional synchronization needed. But if cross-thread interruption ever gets implemented, this needs to be extended of course.

BTW, are there currently dependencies within Druntime/Phobos that depend on the object monitor change, or is it just to optimize user code? If there are no further dependencies, can't we at least defer the change for 2 releases and add a deprecation message to vibe.d in the meantime? I think object monitors and `synchronized(object)` are probably used rarely enough, so that this doesn't make too much of a difference anyway (could be wrong of course).

s-ludwig:
> The ManualEvent needs to be throwable because that's how libasync (and libevent or every driver) pipes every error code from the OS to the fiber, when it's waiting for I/O with this ManualEvent. Things as simple as "Remote peer interrupted the connection".

Those operations are not using a `ManualEvent`, though, at least the non-libasync drivers. That's just the basic "resume-with-exception" functionality.

etcimon:
> Those operations are not using a ManualEvent, though, at least the non-libasync drivers. That's just the basic "resume-with-exception" functionality.

So, would the `interrupt` be possible by bringing it further down into the driver? e.g. interrupting a timer by calling [killTimer](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644903%28v=vs.85%29.aspx) on windows ?

s-ludwig:
No, the problem remains that an exception is still needed for proper stack cleanup.

etcimon:
> No, the problem remains that an exception is still needed for proper stack cleanup.

Yeah, so the "resume-with-exception" won't work with tasks that use a nothrow mutex. Gotcha. So basically, if another `Task` dies with an assertion failure, is stuck in an infiniloop or dies off silently in some way, every `TaskCondition` will be instantly hung, and it will hang everything waiting on it recursively.

s-ludwig:
> Yeah, so the "resume-with-exception" won't work with tasks that use a nothrow mutex. Gotcha.

Hmm... it doesn't work as long as a task is currently waiting in the `lock()` call.

> So basically, if another Task dies with an assertion failure,

Assertions are a different issue. Any `Error` should cause the process to get terminated anyway. But yeah, if the assertion is instead caught and DMD didn't insert the proper stack cleanup code, then bad things will happen.

etcimon:
> But yeah, if the assertion is instead caught and DMD didn't insert the proper stack cleanup code, then bad things will happen.

There's also this issue about operations taking too long (in another thread) and needing to be aborted... Usually, you just make another `Task` interrupt it after a timeout. Everything that doesn't use Vibe.d's I/O suddenly becomes less manageable.

s-ludwig:
> There's also this issue about operations taking too long (in another thread) and needing to be aborted... Usually, you just make another Task interrupt it after a timeout. Everything that doesn't use Vibe.d's I/O suddenly becomes less manageable.

Sorry, I didn't really get that. You mean interrupting non-vibe.d I/O operations? Those are generally not  interruptible anyway. Or do you mean low-level libasync operations?

MartinNowak:
> Any Error should cause the process to get terminated anyway.

[Issue 7018 – thrown Error from different thread should lead to program abort](https://issues.dlang.org/show_bug.cgi?id=7018)

> Or do you mean low-level libasync operations?

I think he means something like fetching a website with timeout.

> BTW, are there currently dependencies within Druntime/Phobos that depend on the object monitor change, or is it just to optimize user code?

It's was long-standing problem that synchronized wasn't nothrow and it's one of those missing nothrow annotations that make it very hard to use nothrow at all.
And the case here is very clear, exceptions don't belong into the IMonitor API, because there is no sensible way to recover and retry.
[Issue 11216 – Make synchronized statement `nothrow`](https://issues.dlang.org/show_bug.cgi?id=11216).
Quite some luck that most of the other vibe.d async I/O functions already throw for different reasons, so this won't start to depart too much from phobos.

> If there are no further dependencies, can't we at least defer the change for 2 releases and add a deprecation message to vibe.d in the meantime?

Doable, but I'll try to keep core.sync.* and the newly added std.concurrency.Scheduler as nothrow and only revert IMonitor and `sychronized (obj)`.

MartinNowak:
Yikes, just tried to undo `synchronized (obj)`.
```sh
src/core/thread.d(640): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(640): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(591): Error: function 'core.thread.Thread.start' is nothrow yet may throw
src/core/thread.d(1641): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1641): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1624): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(1731): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1731): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1695): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(4364): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(4364): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(4355): Error: function 'core.thread.Fiber.freeStack' is nothrow yet may throw
src/core/thread.d(640): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(640): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(591): Error: function 'core.thread.Thread.start' is nothrow yet may throw
src/core/thread.d(1641): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1641): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1624): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(1731): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1731): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1695): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(4364): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(4364): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(4355): Error: function 'core.thread.Fiber.freeStack' is nothrow yet may throw
src/core/sync/barrier.d(91): Error: '_d_monitorenter' is not nothrow
src/core/sync/barrier.d(91): Error: '_d_monitorexit' is not nothrow
src/core/sync/barrier.d(89): Error: function 'core.sync.barrier.Barrier.wait' is nothrow yet may throw
src/core/sync/rwmutex.d(187): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(187): Error: '_d_monitorexit' is not nothrow
src/core/sync/rwmutex.d(185): Error: function 'core.sync.rwmutex.ReadWriteMutex.Reader.lock' is nothrow yet may throw
src/core/sync/rwmutex.d(204): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(204): Error: '_d_monitorexit' is not nothrow
src/core/sync/barrier.d(91): Error: '_d_monitorenter' is not nothrow
src/core/sync/barrier.d(91): Error: '_d_monitorexit' is not nothrow
src/core/sync/barrier.d(89): Error: function 'core.sync.barrier.Barrier.wait' is nothrow yet may throw
posix.mak:186: recipe for target 'lib/libdruntime-linux64.a' failed
make[1]: *** [lib/libdruntime-linux64.a] Error 1
make[1]: *** Waiting for unfinished jobs....
src/core/sync/rwmutex.d(187): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(187): Error: '_d_monitorexit' is not nothrow
src/core/sync/rwmutex.d(185): Error: function 'core.sync.rwmutex.ReadWriteMutex.Reader.lock' is nothrow yet may throw
src/core/sync/rwmutex.d(204): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(204): Error: '_d_monitorexit' is not nothrow
posix.mak:177: recipe for target 'lib/libdruntime-linux64.so' failed
make[1]: *** [lib/libdruntime-linux64.so] Error 1
src/core/thread.d(640): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(640): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(591): Error: function 'core.thread.Thread.start' is nothrow yet may throw
src/core/thread.d(1641): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1641): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1624): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(1731): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(1731): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(1695): Error: function 'core.thread.Thread.add' is nothrow yet may throw
src/core/thread.d(4364): Error: '_d_monitorenter' is not nothrow
src/core/thread.d(4364): Error: '_d_monitorexit' is not nothrow
src/core/thread.d(4355): Error: function 'core.thread.Fiber.freeStack' is nothrow yet may throw
src/core/sync/barrier.d(91): Error: '_d_monitorenter' is not nothrow
src/core/sync/barrier.d(91): Error: '_d_monitorexit' is not nothrow
src/core/sync/barrier.d(89): Error: function 'core.sync.barrier.Barrier.wait' is nothrow yet may throw
src/core/sync/rwmutex.d(187): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(187): Error: '_d_monitorexit' is not nothrow
src/core/sync/rwmutex.d(185): Error: function 'core.sync.rwmutex.ReadWriteMutex.Reader.lock' is nothrow yet may throw
src/core/sync/rwmutex.d(204): Error: '_d_monitorenter' is not nothrow
src/core/sync/rwmutex.d(204): Error: '_d_monitorexit' is not nothrow
```
And it won't get better in phobos.
Maybe we dismissed the lowering idea too early, because the important thing is that `core.sync.*` is nothrow. Setting monitors on classes is rarely used (and in fact we want to get rid of it).

MartinNowak:
> It's was long-standing problem that synchronized wasn't nothrow

https://github.com/D-Programming-Language/druntime/pull/760/files#r11499570

etcimon:
Ah I see. I think synchronized should be nothrow as well. If you look at the new pull request I made [here](https://github.com/rejectedsoftware/vibe.d/pull/998) the `interrupt()` now crashes the program through an assertion check in favor of the old behavior, and `TaskMutexInt` is offered for special cases where interruptions are necessary.

I've noticed here and there the issue where the assertion failure in a foreign thread doesn't close the program, it caused me a lot of headaches. I think I fixed it in my debug build by catching `Throwable` and exiting all event loops manually here with `getEventDriver().exitEventLoop();`: https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/core/core.d#L1420

It's actually very surprising that this isn't already built-in

s-ludwig:
>> Or do you mean low-level libasync operations?
> I think he means something like fetching a website with timeout.

I meant the part "Everything that doesn't use Vibe.d's I/O suddenly becomes less manageable."

> Yikes, just tried to undo synchronized (obj).

I see, bummer. I had hoped that most uses of `synchronized` would be directly using the mutex instead of the object monitor, but the `synchronized (obj)` idiom actually appears to be quite common in the standard library.

>  I think synchronized should be nothrow as well.

It would at least have the possibility to be throw/nothrow based on the nothrowness of the underlying `lock`/`unlock` methods. If there just wasn't the issue with the object monitor. There would have to be two monitor references or some LSB hack to distinguish objects with nothrow monitors from old style ones. But such solutions sound like too much.

Regarding `assert` on `interrupt()`- this behavior is out of the question. It's a receipt for rare and difficult to fix bugs. If it turns out we have to break backwards compatibility with 2.067, then it would have to be the solution that defers the throw. Since it looks like this is going to be the case, I'll have a look at how to implement that in a clean way.

s-ludwig:
One thing, it didn't really occur to me before, but `scopedLock` might be a bit error prone, when people try to use it as (forgetting to store the return value):
```
{
    scopedLock(mutex);
    // ... do some locked things
}
```
a safer alternative would be to have something like `void performLocked(alias DEL)(Mutex mutex)` that is used like this:
```
{
    mutex.performLocked!({
        // ... do some locked things
    });
}
```

s-ludwig:
@MartinNowak what exactly is the plan for the object monitor? Will `Object.Monitor` stay as an inheritance target or is that also going to get removed (you mentioned `IMonitor`)?

etcimon:
> What would be the intend of interrupting the connection attempt? Trying to reuse the task more quickly for something else in order to save resources?

Actually, I thought of this just now. The interrupt feature on a task can be a last straw effort recovering from OOM errors on system-critical vibe.d processes.

MartinNowak:
> @MartinNowak what exactly is the plan for the object monitor? Will Object.Monitor stay as an inheritance target or is that also going to get removed (you mentioned IMonitor)?

[`alias IMonitor = Object.Monitor;`](https://github.com/D-Programming-Language/druntime/blob/83bfb3c2f9b3fdbf05a37fc21a68a9e4aba98f8d/src/object_.d#L1769)

Looks like the lowering works out, still need to get the codegen right though.
The idea is to use `lock`/`unlock` if they are available and fall back to the monitor otherwise.
Then `IMonitor` can remain throwing, but classes deriving from it (`core.sync.*`) may implement `nothrow` and we can even use synchronized with struct based locks.
We have to add an assertion that any Mutex-like class doesn't have a different monitor set, but I would be very surprised to find such code.

s-ludwig:
> alias IMonitor = Object.Monitor;

Yeah, I've seen that. Rereading everything, you just meant the `nothrow` annotations when you said "no clear path to deprecate IMonitor", right? I got that wrong and thought that the whole interface is to be removed.

Regarding `Mutex` (etc.), has there been a decision to make those `final` eventually? If yes, has it been considered to add another way to plug in different mutex implementations? There have been a few occurrences where being able to inject a `TaskMutex` instead of a `Mutex` was quite useful for making libraries compatible with both, standard I/O and vibe.d.

I've implemented the deferred throwing of exceptions in a [separate branch](https://github.com/rejectedsoftware/vibe.d/commits/fix_dmd_2067) now. There is still a `performLocked` function in there as a replacement for `synchronized`, but I'll remove that prior to release if the `synchronized` lowering works.

etcimon:
> I've implemented the deferred throwing of exceptions in a separate branch now.

Oh now I see what you meant by exception defer'ing. So the `processDeferredExceptions()` would be called if the fiber yields, I would also check it before the fiber ends. Looks like a durable solution

s-ludwig:
> So the processDeferredExceptions() would be called if the fiber yields, I would also check it before the fiber ends.

Agreed, I've added log messages in 240d46381de6fef4c8956128c248c95fdf127818 (unhandled `InterruptException`s are harmless, so only a debug message).

MartinNowak:
> Regarding Mutex (etc.), has there been a decision to make those final eventually?

I'd like to, but I have no idea how much code it would break. So we won't do it soon.

> If yes, has it been considered to add another way to plug in different mutex implementations?
> There have been a few occurrences where being able to inject a TaskMutex instead of a Mutex was quite useful for making libraries compatible with both, standard I/O and vibe.d.

If you wanted to do this polymorphic, the `nothrow` in Mutex would still enforce `nothrow` in TaskMutex. It would work using a templated Mutex or by passing IMonitor.

MartinNowak:
What a nightmare. So I made the lowering work, and it's actually quite nice (locking on a struct is a bit more efficient than using the monitor thingie).
But I ran into an unexpected problem. Before this, dmd didn't semantically check `synchronized` (_d_monitorenter/exit) and would silently allow to use it in `pure @safe @nogc` code. It was even possible to synchronize on a const mutex.
The correct attributes for `core.sync.*` would be `nothrow @safe @nogc` and we'd need to add all three to `core.sync.*` in order to not break correct code. Synchronized isn't pure, so code relying on that would break. On top of that pthread isn't yet `@nogc` annotated.

While this definitely goes into the right direction (cleans up cruft, makes attribute correct, and is faster), I'm hesitating to change so much stuff at the end of the beta.

https://github.com/MartinNowak/dmd/tree/synchronizedLowering
https://github.com/MartinNowak/druntime/tree/synchronizedLowering
https://github.com/MartinNowak/phobos/tree/synchronizedLowering

Will need to sleep about that. Might be better to undo the whole synchronized nothrow thing and fix it correctly in 2.068 :(.

s-ludwig:
> If you wanted to do this polymorphic, the nothrow in Mutex would still enforce nothrow in TaskMutex. It would work using a templated Mutex or by passing IMonitor.

Sure, there will be *some* breakage no matter what. But I was thinking more about something that let's you change the implementation without inheritance (similar to the drivers in vibe.d). But such a thing probably makes most sense as a holistic approach that also includes I/O and similar operations.

> Before this, dmd didn't semantically check synchronized (_d_monitorenter/exit) and would silently allow to use it in pure @safe @nogc code. It was even possible to synchronize on a const mutex.

Would adding a fallback to the old behavior for arguments that have previously been supported be an option?

MartinNowak:
> Can you link the PR that changed relevant bits? As I have already mentioned in mail-list, bunch of freshly ported code (for 2.066.1) got broken with various references to missing attributes and I have zero idea why.

That's for Tango I assume @Dicebot, right?
If it's for something else I'd like to have more details.
Can I use http://code.dlang.org/packages/tango to reproduce the error?

I guess you have the same problem as Jacob Carlborg.
http://forum.dlang.org/post/62A71AF2-9165-48C0-A60F-8274C3F8D34E@me.com
For some reason he switched that conversation to private email, but here is what I recommended.

> And I just looked at tango, the tango.core.sync modules are identical
> to core.sync, so you should simply forward to druntime, problem solved.

> It's fairly obvious that the druntime implementation comes from tango.
> There might be a few addition in druntime, but you should be able to
> simply use the druntime version.

> https://github.com/SiegeLord/Tango-D2/blob/d2port/tango/core/sync/Mutex.d
> https://github.com/D-Programming-Language/druntime/blob/master/src/core/sync/mutex.d

> https://github.com/SiegeLord/Tango-D2/blob/d2port/tango/core/sync/ReadWriteMutex.d
> https://github.com/D-Programming-Language/druntime/blob/master/src/core/sync/rwmutex.d

We're rolling back this change for 2.067.0 and retackle the issue in 2.068.

MartinNowak:
Sorry, that this cause so much trouble for vibe.d. Once we rolled back the nothrow thing, you might have to undo some code in vibe.d added for 2.067 as well :(.
https://github.com/D-Programming-Language/dmd/pull/4459

We can make the lowering approach work, but at this time it might break too much code because of [Issue 14251](https://issues.dlang.org/show_bug.cgi?id=14251 "synchronized (mtx) doesn't check attributes (pure, const)").

etcimon:
It would be nice to see at least a deprecation message for the throwing mutex and have it work thrown as well since nothrow can be just a suggestion for the moment because it's a restrictive attribute rather than additive. 

I.e. using ` @nothrow` could work inside both nothrow and throwable scope but show deprecation when derived to a throwable scope

MartinNowak:
We'll also need to make sure that a new vibe.d release is ready soon after 2.067 is out.

s-ludwig:
I've tagged 0.7.23-beta.1 now. `performLocked` is kept for internal use only.

MartinNowak:
Looks good, we need to revert a few nothrow annotations for 2.067.0-b4.

```
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/core/drivers/libevent2.d(627): Error: _d_monitorenter is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/core/drivers/libevent2.d(627): Error: _d_monitorexit is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/core/drivers/libevent2.d(618): Error: function vibe.core.drivers.libevent2.Libevent2ManualEvent.emit is nothrow yet may throw
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/http/client.d(395): Deprecation: this is not an lvalue
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/http/client.d(359): Deprecation: this is not an lvalue
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/http/client.d(359): Deprecation: this is not an lvalue
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(131): Error: _d_monitorenter is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(131): Error: _d_monitorexit is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(124): Error: function vibe.utils.memory.LockAllocator.alloc is nothrow yet may throw
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(146): Error: _d_monitorenter is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(146): Error: _d_monitorexit is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(134): Error: function vibe.utils.memory.LockAllocator.realloc is nothrow yet may throw
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(160): Error: _d_monitorenter is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(160): Error: _d_monitorexit is not nothrow
../../../.dub/packages/vibe-d-0.7.23-beta.1/source/vibe/utils/memory.d(149): Error: function vibe.utils.memory.LockAllocator.free is nothrow yet may throw
```

```
../../../.dub/packages/vibe-d-master/source/vibe/core/drivers/libevent2.d(853): Warning: instead of C-style syntax, use D-style syntax 'char[64] buf'
../../../.dub/packages/vibe-d-master/source/vibe/stream/openssl.d(54): Warning: instead of C-style syntax, use D-style syntax 'ubyte[64] m_peekBuffer'
../../../.dub/packages/vibe-d-master/source/vibe/core/concurrency.d(1193): Error: 'vibe.core.task.Task.getThis' is not nothrow
../../../.dub/packages/vibe-d-master/source/vibe/core/concurrency.d(1193): Error: function 'vibe.core.concurrency.VibedScheduler.thisInfo' is nothrow yet may throw
../../../.dub/packages/vibe-d-master/source/vibe/core/concurrency.d(1194): Error: constructor this is not nothrow
../../../.dub/packages/vibe-d-master/source/vibe/core/concurrency.d(1194): Error: function 'vibe.core.concurrency.VibedScheduler.newCondition' is nothrow yet may throw
../../../.dub/packages/vibe-d-master/source/vibe/core/core.d(1085): Deprecation: function core.thread.Fiber.call is deprecated - Please pass Fiber.Rethrow.yes or .no instead of a boolean.
../../../.dub/packages/vibe-d-master/source/vibe/core/net.d(179): Error: cannot cast expression ip[cast(ulong)(i * 2)..cast(ulong)(i * 2 + 2)] of type ubyte[] to ubyte[2]
../../../.dub/packages/vibe-d-master/source/vibe/crypto/passwordhash.d(64): Error: undefined identifier format
../../../.dub/packages/vibe-d-master/source/vibe/data/bson.d(710): Deprecation: alias std.process.getpid is deprecated - Please use thisProcessID instead
../../../.dub/packages/vibe-d-master/source/vibe/http/client.d(319): Deprecation: this is not an lvalue
../../../.dub/packages/vibe-d-master/source/vibe/http/client.d(356): Deprecation: this is not an lvalue
../../../.dub/packages/vibe-d-master/source/vibe/http/common.d(596): Error: escaping reference to local variable c
../../../.dub/packages/vibe-d-master/source/vibe/http/common.d(533): Error: template instance vibe.http.common.CookieValueMap.opBinaryRight!in error instantiating
../../../.dub/packages/vibe-d-master/source/vibe/http/common.d(533): Error: rvalue of in expression must be an associative array, not const(CookieValueMap)
../../../.dub/packages/vibe-d-master/source/vibe/http/common.d(555): Error: rvalue of in expression must be an associative array, not const(CookieValueMap)
../../../.dub/packages/vibe-d-master/source/vibe/http/log.d(180): Error: rvalue of in expression must be an associative array, not CookieValueMap
../../../.dub/packages/vibe-d-master/source/vibe/http/log.d(43): Error: template instance vibe.http.log.formatApacheLog!(FixedAppender!(const(char)[], 2048LU, const(char))) error instantiating
../../../.dub/packages/vibe-d-master/source/vibe/http/server.d(1394): Error: rvalue of in expression must be an associative array, not CookieValueMap
../../../.dub/packages/vibe-d-master/source/vibe/http/server.d(1235): Error: function vibe.http.server.handleRequest no return exp; or assert(0); at end of function
```

s-ludwig:
I'd just extend the `scope (failure) assert(false)` to `__VERSION__ == 2067`, since all operations are still de facto `nothrow`.

s-ludwig:
Errors fixed in 8d73d64c8d67729a5be06e2766ef5c0e38a8430a. However, I don't get the deprecation warnings. `this` is just passed as a normal argument to a constructor in those lines, nothing for which an lvalue should be needed.

MartinNowak:
> However, I don't get the deprecation warnings. this is just passed as a normal argument to a constructor in those lines, nothing for which an lvalue should be needed.

I got them, because my app had other compiler errors, sometimes you got to see gagged internal error from dmd, e.g. from type inference. After I fixed them they were gone, thx for the quick followup.

s-ludwig:
> I got them, because my app had other compiler errors, sometimes you got to see gagged internal error from dmd, e.g. from type inference. After I fixed them they were gone, thx for the quick followup.

Interesting, they are still there for me, though. I'll try to reduce the code.

MartinNowak:
Maybe deprecations aren't correctly gagged, just a guess.

s-ludwig:
The cause is `scoped`'s use of `auto ref`: `@system auto scoped(Args...)(auto ref Args args)`

Passing `this` as an argument triggers the warning. Looks like the compiler should opt for a non-ref argument instead, right?

s-ludwig:
https://issues.dlang.org/show_bug.cgi?id=14283

etcimon:
Yup, got this as well. It works if you `import std.conv : to;` at ln 542

s-ludwig:
See also #1016, definitely seems to be a DMD or Druntime issue (I *think* there is a possibly related bug report already).

Geod24:
Compiling 0.7.23, just FYI. I didn't wanted to open an issue for this, since it concerns us, but isn't really our problem to solve.

I just did `dub init size --type=vibe.d` and built with different compiler.
DMD 2.066.1, DMD 2.067.1, LDC 0.15.1 (2.066.1), GDC 4.9.2 (2.065.0)

````
-rw-r--r-- 1 geod24 users 115M Mar 27 22:50 libevent-debug-linux.posix-x86_64-dmd_2066-45C8210F3493AB0A9BCA7FEB51E586C0/libvibe-d.a
-rw-r--r-- 1 geod24 users 155M Mar 27 22:49 libevent-debug-linux.posix-x86_64-dmd_2067-1EF4C92940F058E654A032934ACB22B6/libvibe-d.a
-rw-r--r-- 1 geod24 users  18M Mar 27 22:34 libevent-debug-linux.posix-x86_64-gdc_2065-7358A80C14565DE56E0B6FB3A72896DB/libvibe-d.a
-rw-r--r-- 1 geod24 users  27M Mar 27 22:37 libevent-debug-linux.posix-x86_64-ldc_0-BE3C855847D82415D6BBB036DABCBA69/libvibe-d.a
-rw-r--r-- 2 geod24 users 218M Mar 27 22:32 libevent-unittest-linux.posix-x86_64-dmd_2066-17D8A0627F9E1D64D5F27E709699802B/libvibe-d.a
-rw-r--r-- 1 geod24 users 279M Mar 27 22:26 libevent-unittest-linux.posix-x86_64-dmd_2067-EDCAF0C6E5DA711FC9EABC3DC58DA5E3/libvibe-d.a
-rw-r--r-- 1 geod24 users  29M Mar 27 22:30 libevent-unittest-linux.posix-x86_64-gdc_2065-13B17EC453B2D4CC95CD8FF174D65BFF/libvibe-d.a
-rw-r--r-- 1 geod24 users  44M Mar 27 22:28 libevent-unittest-linux.posix-x86_64-ldc_0-BB0162C9D289A6400EA2F88AE240F383/libvibe-d.a
````

Stripped:
````
-rw-r--r-- 1 geod24 users  37M Mar 27 22:51 libevent-debug-linux.posix-x86_64-dmd_2066-45C8210F3493AB0A9BCA7FEB51E586C0/libvibe-d.a
-rw-r--r-- 1 geod24 users  52M Mar 27 22:51 libevent-debug-linux.posix-x86_64-dmd_2067-1EF4C92940F058E654A032934ACB22B6/libvibe-d.a
-rw-r--r-- 1 geod24 users 5.6M Mar 27 22:51 libevent-debug-linux.posix-x86_64-gdc_2065-7358A80C14565DE56E0B6FB3A72896DB/libvibe-d.a
-rw-r--r-- 1 geod24 users 7.6M Mar 27 22:51 libevent-debug-linux.posix-x86_64-ldc_0-BE3C855847D82415D6BBB036DABCBA69/libvibe-d.a
-rw-r--r-- 2 geod24 users  71M Mar 27 22:51 libevent-unittest-linux.posix-x86_64-dmd_2066-17D8A0627F9E1D64D5F27E709699802B/libvibe-d.a
-rw-r--r-- 1 geod24 users  94M Mar 27 22:52 libevent-unittest-linux.posix-x86_64-dmd_2067-EDCAF0C6E5DA711FC9EABC3DC58DA5E3/libvibe-d.a
-rw-r--r-- 1 geod24 users 9.1M Mar 27 22:52 libevent-unittest-linux.posix-x86_64-gdc_2065-13B17EC453B2D4CC95CD8FF174D65BFF/libvibe-d.a
-rw-r--r-- 1 geod24 users  13M Mar 27 22:52 libevent-unittest-linux.posix-x86_64-ldc_0-BB0162C9D289A6400EA2F88AE240F383/libvibe-d.a
````

Such a difference can only mean that either GDC / LDC are doing something horribly wrong, or DMD is.

etcimon:
These are the debug symbols, it combines those from phobos as well for every template. It's normal, I think LDC/GDC is in the wrong here. Even comparable C++ libraries compile to 180MB easy on GCC when those debug symbols are enabled. This is for .a static libs of course, if you compile it into an executable it'll go down to 1/5th of that, and with plain/release maybe 1/25th of that.

PR #974, opened 2015-02-02
use parseRFC822DateTime from std.datetime

MartinNowak:
- available since 2.066

### DIFF
--- a/source/vibe/inet/message.d
+++ b/source/vibe/inet/message.d
@@ -143,47 +143,53 @@ string toRFC822DateTimeString(SysTime time)
 	return ret.data;
 }
 
-/**
-	Parses a date+time string according to RFC-822/5322.
-*/
-SysTime parseRFC822DateTimeString(string str)
+static if (__VERSION__ >= 2066)
 {
-	auto idx = str.indexOf(',');
-	if( idx != -1 ) str = str[idx+1 .. $].stripLeft();
+	/// Parses a date+time string according to RFC-822/5322.
+	alias parseRFC822DateTimeString = parseRFC822DateTime;
+}
+else
+{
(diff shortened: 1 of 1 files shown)


MartinNowak:
[Issue 12886 – std.datetime cannot parse HTTP date](https://issues.dlang.org/show_bug.cgi?id=12886)

Geod24:
Good catch. But what's up with the autotester and those frequent 404 ?

0.04scurl: (22) The requested URL returned error: 403
0.24s$ curl http://code.dlang.org/files/dub-${LATEST_DUB}-linux-x86_64.tar.gz | tar -C ~/dmd2/linux/bin64 -xzf -
curl: (22) The requested URL returned error: 404

Is it an issue on code.dlang.org's end ?

MartinNowak:
> Is it an issue on code.dlang.org's end?

I think so, seen them a couple of times now.

Geod24:
> I think so, seen them a couple of times now.

Maybe it would be good to have some redundancy on dlang.org then.
Also, you need to force push this to let Travis have another go at it.

s-ludwig:
I'm getting sporadic segmentation faults since a while that I haven't looked into, yet. But that should produce a 500 error (well, actually it *should* produce a 502, but that's a different story) while the registry process is being restarted and not a 404, so I'm not completely sure if this is actually related.

I'll just merge for now and then will have a look at the reason for the crash.

s-ludwig:
Turns out that `LATEST_DUB` is simply not defined and it tries to download `/files/dub--linux-x86_64.tar.gz`.

MartinNowak:
Would be nice to somehow get the latest version from code.dlang.org.

PR #980, opened 2015-02-11
add convenient status function to set statusCode using method chaining

MartinNowak:


s-ludwig:
I'm still a bit undecided with regards to method chaining in D. In contrast to C++ we have `with`, for example, which is not much longer:
`with (res) { status = HTTPStatus.notFound; writeBody("not found"); }`
`res.status(HTTPStatus.notFound).writeBody("not found");`
well, actually writing it out is even shorter in this case ;)
`res.status = HTTPStatus.notFound; res.writeBody("not found");`

In this specific case, this would also duplicate trivial functionality which most of the time has a bad cost/benefit ratio in the long term due to API clutter. But there is already a precedent with `URLRouter`, so I'm not completely sure how to go about this. But maybe we could instead add another `writeBody` overload? `res.writeBody(HTTPStatus.notFound, "not found");`

etcimon:
I personally prefer method chaining when there's 3 or more methods in an object, which need to be used sequentially as many times, like when populating an object. The typing is fast and smooth with an IDE that supports auto-completion.

In this case, the fact that you'd usually return afterwards would prompt for another function like `res.exit(HTTPStatus, string);`. It mimics the php [exit](http://php.net/manual/en/function.exit.php) function and would be implemented by throwing a CustomHTTPStatus error possibly (with respects to the current API) :-p

MartinNowak:
> It mimics the php exit function and would be implemented by throwing a CustomHTTPStatus error possibly (with respects to the current API) :-p Maybe a exit(HTTPStatus, Path) version of the method could support reading the html from an error file.

__Don't__ abuse exceptions for control flow, but what I want is a single statement to do `return res.writeBody("foo", HTTPStatus.notFound)`. The overload sounds like a good idea.

PR #984, opened 2015-02-14
Updated HTTPRouter.any function to handle any method request

gedaiu:


Geod24:
HEAD / OPTIONS / TRACE / CONNECT should definitively go in, but I don't think WebDAV method should be there by default. It would probably be better to provide another method (`anyWebDAV` ?) to explicitely set those.

gedaiu:
Hi,

When I see ANY, I am expecting to have the routing to any method, even is some method that is not specified in documentation. 

I am ok with adding another function for anyWebDAV, but if I add this method what we route there? Because webdav needs the GET / PUT / OPTIONS ... etc methods to work and someone who want to implement a webdav protocol might be fooled by "anyWebDAV" if it does not route the basic HTTP methods.

Maybe there should be: anyHttp -> that route all the basic http methods and any that routes all the requests.

what do you think?

Thanks,
Bogdan

s-ludwig:
At least the semantics that were originally intended are to really match any request method. A typical use would be to intercept *all* requests with `router.any("*", ...)`. If such an interception doesn't work for some methods that might easily introduce security issues or similar. I'd even go as far and change the implementation for `foreach` over all possible values of `HTTPMethod` to avoid missing anything in the future.

s-ludwig:
@Geod24 BTW, do you have a specific use case in mind for separate matching of usual and WebDAV specific methods?

PR #986, opened 2015-02-15
Add official supported compiler / version in the README

Geod24:
This has been discussed in various place, but there's no official doc that state it ATM.


s-ludwig:
Thanks, that's definitely good to have mentioned!

PR #989, opened 2015-02-24
[WIP] Test libevent, libev and libasync drivers

Geod24:
Test all the drivers on all supported compiler.
The downside of this is that it requires 18 jobs: https://travis-ci.org/Geod24/vibe.d

It would however need a dub improvement: https://github.com/D-Programming-Language/dub/issues/520

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,20 @@ d:
   - ldc-0.15.1
   - gdc-4.9.0
 
+env:
+  - VIBED_DRIVER=libevent
+  - VIBED_DRIVER=libev
+  - VIBED_DRIVER=libasync
+
 script:
-  - dub test --compiler=$DC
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'd say we should just run `dub test` with each config for now. Compiling all examples seems like overkill. The external tests would fail for libev, so we should exclude at least those for that configuration.

Geod24:
On the long run, I think we should just compile the examples with one config, to make sure they are up to date.
We also need way more tests. I'd do the REST part (I already have something, just need to clean it up and make sure it covers everything, but given my current workload that won't happen before 2 weeks).

Geod24:
I edited it to only test it. Should be good to go.

s-ludwig:
OK, I'll have a look at the libev driver.

s-ludwig:
Nevermind ;)

PR #990, opened 2015-02-24
Fix libev build

Geod24:
Depends on https://github.com/D-Programming-Deimos/libev/pull/10 (and you making a new tag ;) ).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ dub.selections.json
 vibe-d
 tests/*/tests
 __test__libevent__
+__test__libev__
 
 # Examples
 examples/app_skeleton/__test__library__
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,14 @@ d:
   - ldc-0.15.1
   - gdc-4.9.0
 
+before_install:
+  - wget http://dist.schmorp.de/libev/Attic/libev-4.19.tar.gz
+  - tar -xf libev-4.19.tar.gz
+  - (cd libev-4.19 && ./configure && make)
+
 env:
   - VIBED_DRIVER=libevent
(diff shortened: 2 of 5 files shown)


s-ludwig:
Updated with 4.0.1 tag, still waiting for code.dlang.org to update. Once it is updated, can you bump the version number in dub.json?

s-ludwig:
Wait a minute.. It'll have to be 5.0.0 due to the changed callback function signatures. I'll change that.

s-ludwig:
OK, 5.0.0 is up.

Geod24:
Should be good to go.

Geod24:
I had to come up with some hacks to get a satisfying solution.
Since Travis requires you to disable `sudo` to use their docker-based infrastructure (which is much faster), we're downloading `libev` and compiling it in place (only adds about 8s of overhead per build).
I'm not aware of a way to provide lflags to dub, is there one ? I tried `LFLAGS="..." dub test ...`, looks like it's not supported.

s-ludwig:
`LFLAGS` isn't supported, but it could be done with `DFLAGS` and `-Llinkerflag` (for DMD). You just need to also specify the other flags explicitly (`-g -debug`).

s-ludwig:
I'll merge as is for now, the `LFLAGS`issue can be resolved later (I'd do it when I get some time later today).

Geod24:
I tried this approach, it doesn't appear to work either, but I didn't had
time to investigate. Maybe related to
https://github.com/D-Programming-Language/dub/issues/500 .

2015-02-25 10:29 GMT+01:00 Sönke Ludwig <notifications@github.com>:

s-ludwig:
You are probably right, forgot about that.

PR #991, opened 2015-02-24
Assume array hold stack construction ordering

etcimon:
I've started writing a pool stack and considered this as problematic if object B's destructor relies on an object owned by object A.


etcimon:
This is probably just confusing

s-ludwig:
Hm, no I think that's actually a valuable goal! FILO construction/destruction order should always be done if possible.

PR #992, opened 2015-02-25
Fix compilation on DMD v2.067

etcimon:
This also adds a fix for HashMap reallocation issue pointed out earlier.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1346,13 +1346,18 @@ static this()
 
 static ~this()
 {
+	version(VibeLibasyncDriver) {
+		import vibe.core.drivers.libasync;
+		if (LibasyncDriver.isControlThread)
+			return;
+	}
 	auto thisthr = Thread.getThis();
 
 	bool is_main_thread = false;
 
 	synchronized (st_threadsMutex) {
 		auto idx = st_threads.countUntil!(c => c.thread is thisthr);
-		assert(idx >= 0);
+		assert(idx >= 0, "No more threads registered");
 		if (idx >= 0) {
 			st_threads[idx] = st_threads[$-1];
 			st_threads.length--;
@@ -1586,4 +1591,11 @@ private template needsMove(T)
 {
 	// FIXME: reverse the condition and only call .move for non-copyable types!
 	enum needsMove = is(typeof(T.init.move));
+}
+
+version(VibeLibasyncDriver) {
+	shared static ~this() {
+		import libasync.threads : destroyAsyncThreads;
+		destroyAsyncThreads(); // destroy threads		
+	}
 }
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -6,6 +6,7 @@
 	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
 */
 module vibe.core.drivers.libasync;
+
 version(VibeLibasyncDriver):
 
 import vibe.core.core;
@@ -14,7 +15,7 @@ import vibe.core.drivers.threadedfile;
 import vibe.core.log;
 import vibe.inet.path;
 
-import libasync.all;
+import libasync;
 import libasync.types : Status;
 
 import std.algorithm : min;
@@ -33,14 +34,15 @@ import core.sync.mutex;
 import std.container : Array;
 
 import vibe.core.drivers.timerqueue;
-import vibe.utils.memory;
+import libasync.internals.memory;
 import vibe.utils.array : FixedRingBuffer;
 import std.stdio : File;
 
 private __gshared EventLoop gs_evLoop;
 private EventLoop s_evLoop;
 private DriverCore s_driverCore;
-EventLoop getEventLoop()
+
+EventLoop getEventLoop() nothrow
 {
 	if (s_evLoop is null)
 		return gs_evLoop;
@@ -48,7 +50,7 @@ EventLoop getEventLoop()
 	return s_evLoop;
 }
 
-DriverCore getDriverCore()
+DriverCore getDriverCore() nothrow
 {
 	assert(s_driverCore !is null);
 	return s_driverCore;
@@ -72,7 +74,7 @@ final class LibasyncDriver : EventDriver {
 		SysTime m_nextSched;
 	}
 		
-	this(DriverCore core)
+	this(DriverCore core) nothrow
 	{
 		//if (isControlThread) return;
 
@@ -99,7 +101,7 @@ final class LibasyncDriver : EventDriver {
 		s_evLoop = getThreadEventLoop();
 		if (!gs_evLoop)
 			gs_evLoop = s_evLoop;
-		logInfo("Loaded event-d backend in thread %s", Thread.getThis().name);
+		logTrace("Loaded libasync backend in thread %s", Thread.getThis().name);
 
 	}
 
@@ -300,7 +302,7 @@ final class LibasyncDriver : EventDriver {
 			m_timers.destroy(timer_id);
 	}
 	
-	bool isTimerPending(size_t timer_id) { return m_timers.isPending(timer_id); }
+	bool isTimerPending(size_t timer_id) nothrow { return m_timers.isPending(timer_id); }
 	
 	void rearmTimer(size_t timer_id, Duration dur, bool periodic)
 	{
@@ -1571,19 +1573,5 @@ void recycleID(size_t id) {
 	}
 	catch (Exception e) {
 		assert(false, "Error destroying Manual Event ID: " ~ id.to!string ~ " [" ~ e.msg ~ "]");
-	}
-}
-version(unittest){
-	static ~this() {
-		import std.c.stdlib : exit;
-		if (thread_isMainThread)
-			exit(0);
-	}
-} else {
-	version(Windows) static ~this() {
-		// these would cause exit code -11 on linux
-		if (thread_isMainThread) 
-			destroyAsyncThreads(); // destroy threads
-
 	}
 }
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -1,5 +1,5 @@
 /**
-	Task synchronization facilities
+	Interruptible Task synchronization facilities
 
 	Copyright: © 2012 RejectedSoftware e.K.
 	Authors: Leonid Kramer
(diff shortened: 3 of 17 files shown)


source/vibe/core/drivers/libasync.d
```diff
@@ -760,7 +761,7 @@ final class LibasyncManualEvent : ManualEvent {
 				auto evloop = getEventLoop();
 				shared AsyncSignal sig = cast(shared AsyncSignal) signal;
 				if (!sig.trigger(evloop))
-					throw new Exception(sig.error);
+					return; //throw new Exception(sig.error);
```
Geod24: In which situation could this happen ? Wouldn't an asset be better over a silent fail ?

etcimon: Oops, that's an artifact when I was experimenting making this method `nothrow`

source/vibe/core/sync.d
```diff
@@ -122,26 +122,30 @@ class TaskMutex : core.sync.mutex.Mutex {
...
-		while (!tryLock()) ecnt = m_signal.wait(ecnt);
+		try {
+			if (tryLock()) return;
+			debug assert(m_owner == Task() || m_owner != Task.getThis(), "Recursive mutex lock.");
+			atomicOp!"+="(m_waiters, 1);
+			version(MutexPrint) writefln("mutex %s wait %s", cast(void*)this, atomicLoad(m_waiters));
+			scope(exit) atomicOp!"-="(m_waiters, 1);
+			auto ecnt = m_signal.emitCount();
+			while (!tryLock()) ecnt = m_signal.wait(ecnt);
+		} catch {}
```
Geod24: We're just hiding the problem here, we need to fix it.

source/vibe/core/core.d
```diff
@@ -1587,3 +1592,10 @@ private template needsMove(T)
 	// FIXME: reverse the condition and only call .move for non-copyable types!
 	enum needsMove = is(typeof(T.init.move));
 }
+
+version(VibeLibasyncDriver) {
+	shared static ~this() {
+		import libasync.threads : destroyAsyncThreads;
+		destroyAsyncThreads(); // destroy threads		
+	}
+}
```
Geod24: IIUC, this (and code at line 1349) is similar to what was in libasync (at line 1575). Why is the change needed ?
IMO driver-specific-code should go into core, unless there's no other way.

etcimon: > IIUC, this (and code at line 1349) is similar to what was in libasync (at line 1575). Why is the change needed ?
> IMO driver-specific-code should go into core, unless there's no other way.

I get a runtime circular reference error in druntime if I leave it in the driver. This is really the only way I've found, and it's a necessity now.

source/vibe/db/redis/redis.d
```diff
@@ -417,6 +417,15 @@ struct RedisDatabase {
...
 
+	/// When all the elements in a sorted set are inserted with the same score, in order to force lexicographical ordering, 
+	/// this command returns all the elements in the sorted set at key with a value between min and max.
+	RedisReply!T zrangeByLex(T = string)(string key, string min = "-", string max = "+", long offset = 0, long count = -1)
+		if(isValidRedisValueType!T)
+	{
+		if (offset > 0 || count != -1) return request!(RedisReply!T)("ZRANGEBYLEX", key, min, max, "LIMIT", offset, count);
+		else return request!(RedisReply!T)("ZRANGEBYLEX", key, min, max);
+	}
+
```
Geod24: Change looks unrelated. I know it's related to http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/24144/ but it should be in a separate P.R.

source/vibe/core/core.d
Geod24: Was it present in 2.066 (and before) or does it comes with 2.067 ?

source/vibe/core/sync.d
etcimon: I didn't want to change the `ManualEvent` interface to nothrow. It's going to create a huge mess, and I only wanted it to run for now. Maybe I could turn any exception into a throwable through an assert in the catch clause?

source/vibe/core/core.d
etcimon: It follows a fix I made in libasync, where it checks that daemon threads are destroyed when exiting to avoid some invalid memory operations. So everything using a newer version of libasync gets an assertion error on exit unless the threads are destroyed properly. I couldn't destroy them in the driver in any DMD version anyways.

source/vibe/http/common.d
```diff
@@ -601,7 +601,7 @@ struct CookieValueMap {
 
 	inout(string)* opBinaryRight(string op)(string name) inout if(op == "in")
 	{
-		foreach(c; m_entries)
+		foreach(ref c; m_entries)
```
Geod24: This shouldn't be needed. I did the same change in #972, but Sonke fixed it in https://github.com/rejectedsoftware/vibe.d/commit/29d3fad3b03b0a65a8d91ad0152d60ad965896d2

etcimon: This gives me an error about escaping local references in the newer DMD. I was working with git head so any fix that was there wasn't working

source/vibe/core/sync.d
Geod24: Well, the change in Druntime itself created a huge mess ;)
There is WIP to get a proper solution for 2.067, but for the time being, it's okay if it doesn't compile with the beta IMO.
I'd prefer if this P.R. focused on getting libasync green on the other supported platform (2.065 and 2.066). We'll get the three drivers working at the same time (that's why I added them to the auto tester, I'd also like libasync to be the default on the long run).

source/vibe/internal/meta/uda.d
```diff
@@ -102,7 +102,7 @@ template findNextUDA(UDA, alias Symbol, long idx, bool allow_types = false)
 	static assert(idx >= 0, "Index given to findNextUDA can't be negative");
 	static assert(idx <= udaTuple.length, "Index given to findNextUDA is above the number of attribute");
 
-	private template extract(size_t index, list...)
+	public template extract(size_t index, list...)
```
Geod24: Looks like a regression to me. I encountered it in #972 but didn't have time to look into this. I'll see if I can repro / submit an issue tomorrow afternoon.

etcimon: It allows everything to compile on my end, so, I'm happy with it :-)

source/vibe/http/common.d
s-ludwig: Does your current code here compile with DMD 2.065? It gave an error about not being able to convert from `inout` to `inout` or similar, which is why the `static if` my fixing commit was still necessary.

s-ludwig:
Still fails, because `libasync.all` doesn't exist.

etcimon:
> Still fails, because libasync.all doesn't exist.

Yes, that one was removed recently in favor of just libasync

source/vibe/http/common.d
etcimon: Travis seems to suggest everything is good all the way down to libevent 2.065 so this seems like the proper way to go

Geod24:
It's in need for a rebase.

I still don't consider 7cca2ba as a good solution though, especially given the fact it turns an `Exception` into an `Error`.

etcimon:
I think the proper solution is to stop inheriting from `Object.Monitor`, stop using synchronized in fibers, and write a `scopedLock!T` for it instead (so that fibers can be interrupted!)

etcimon:
> Into segmentation fault actually (with -release) ;)

I got a strange segmentation fault in -release using dmd master with libasync (Windows). This isn't supposed to happen, what platform are you on.

etcimon:
Nevermind, that's actually what usually happens on assertion failures in release ^^

etcimon:
I hope this commit will settle the question :)

s-ludwig:
Sorry, but in this form this is not a solution. This is all public code and changing the inheritance like this is a breaking change. All breaking changes must at least go through a deprecation phase.

etcimon:
> All breaking changes must at least go through a deprecation phase.

How do you deprecate broken code?

etcimon:
This new fix adds features on top for DMD <= 2.066, and makes vibe.d compatible for DMD 2.067.

Unfortunately, it'll break anything that tries to run on DMD 2.067, but that was already the case.

I don't know how to deprecate it properly, but the changes for DMD >= 2.067 are as follows:

- `TaskMutex`, `RecursiveTaskMutex`, `TaskCondition` are now detached from `Object.Monitor`
- `synchronized` statement is unavailable as a consequence, it must be replaced with `auto l = mutex.locked` which returns a `ScopedMutexLock`
- If using DMD >= 2.067, the `with` statement is also available as a convenience (it doesn't work on <= 2.066): `with(mutex.locked) { // synchronized }`

s-ludwig:
What would have to be deprecated is the old API. This would for example be the whole `TaskMutex` class. A new class would then have to be added as a replacement.  But maybe there is a more elegant path, too... maybe using some `alias this` trickery. But the final direction taken in Druntime needs to be clear first.

BTW, `TaskMutex.locked` is redundant and doesn't save much over `m.scopedLock` (UFCS). I'd remove it to avoid confusion.

etcimon:
Didn't think the UFCS would allow me to remove parenthesis.

So this last fix should be good, if you can merge the travis exclusion I'm sure we can see the whole thing go green.

s-ludwig:
I would have liked to keep the mutex changes separate from the libasync ones (defer it until the Druntime case is settled), but anyway, that can always be changed again before release. But the first commit still seems to contain merge errors (reverts some changes in the libevent driver), or was that intended (Travis still passes)?

etcimon:
> But the first commit still seems to contain merge errors (reverts some changes in the libevent driver)

That was intentional, the changes I submitted are in a different direction and the `nothrow` / `static assert(false)` are unnecessary now. This fix doesn't affect the DMD <= 2.066.1 builds

> I would have liked to keep the mutex changes separate from the libasync ones (defer it until the Druntime case is settled)

They allow libasync to run on 2.067, which is relevant to the pull :)

etcimon:
> (Travis still passes)?

Travis passes on everything except the 4 versions that aren't supported by libasync (yet)

s-ludwig:
So did you make the changes in libevent.d on purpose? Why are they in the commit that fixes the libasync driver?

etcimon:
> So did you make the changes in libevent.d on purpose? Why are they in the commit that fixes the libasync driver?

They're in the libasync driver because it addresses the mutex issues in 2.067, where the proposed solution was to make the drivers nothrow. The libevent driver had a few added nothrows, which I removed because it seemed unnecessary given the fact that the new approach is to let the mutexes and drivers remain throwable.

I can revert the changes if you like.

s-ludwig:
> I can revert the changes if you like.

No no, I think it's most probably a step in the right direction. It just seemed out of place in the commit order back then.

etcimon:
Green!

s-ludwig:
Yay!

s-ludwig:
And thanks of course!

source/vibe/core/sync.d
```diff
@@ -14,30 +14,71 @@ import vibe.core.driver;
...
 	lock,
 	tryLock,
 	defer
 }
 
+interface Lockable {
+	@trusted bool tryLock();
+	@trusted void lock();
+	@trusted void unlock();
+}
```
MartinNowak: This won't work with `synchronized` and really looks like an abysmal solution.

PR #994, opened 2015-02-26
Exclude unsupported builds from travis

etcimon:
The libasync driver is not yet ready to be built on these compilers, and I don't intend to work it out until a little later.

This will allow the vibe.d repo to go green again soon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,16 @@ env:
   - VIBED_DRIVER=libev-travis
   - VIBED_DRIVER=libasync
 
+matrix:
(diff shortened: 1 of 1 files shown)


s-ludwig:
Had to revert the change. I forgot that Travis complains about a malformed travis.yml before merging. Something was wrong, but I don't know enough about the format to be able to say what.

etcimon:
It was this: https://github.com/etcimon/vibe.d/commit/e4e7fc30152b8260a7376b744840c01826955d11

PR #995, opened 2015-02-26
Fix syntax on travis

etcimon:
^^


etcimon:
This travis patch and the `fix-libasync-new-dmd` pull request result in the following travis build:

https://travis-ci.org/etcimon/vibe.d

PR #996, opened 2015-02-27
Fix RedisSessionStore.create()

YusukeSuzuki:
redis doesn't let empty hash. so insert place holder to initial
session key-value.


s-ludwig:
Thanks! Using the session key seemed a bit odd to me at first, but really it should be as good as anything else. I would probably have opted for something like `"__placeholder", ""`, but this is ok, too.

PR #997, opened 2015-03-03
Add example for basic authentication.

redstar:
This is (more or less) the example from the web site pimped with 2 templates.
The directory is named in a way that digest authentication can be easiely added as next directory (in sorted order).

### DIFF
--- a/examples/auth_basic/dub.json
+++ b/examples/auth_basic/dub.json
@@ -1,0 +1,9 @@
+﻿{
+	"name": "auth-basic-example",
+	"description": "Demonstrates basic authentication.",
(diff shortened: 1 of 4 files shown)


s-ludwig:
Thanks, definitely good to have!

aka-demik:
Diet templates in this example has the extension "dl" (`examples/auth_basic/views/index.dl`). In other examples, "dt". Extension does not matter, or is there something special?

s-ludwig:
Oh, didn't notice that. It matters for the blocks/extension feature, because the implied extension is ".dt" there. I'll fix this.

PR #998, opened 2015-03-04
Use Monitor by default

etcimon:
These changes should allow `Interrupt` to be used on a special kind of mutex (`TaskMutexInt`) without breaking compatibility in 2.067 with synchronized Monitor object. There is no specific deprecation path, just an assertion failure when interrupt is used incorrectly.

### DIFF
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -15,6 +15,12 @@ import core.atomic;
 import core.sync.mutex;
 import core.sync.condition;
 import std.stdio;
+
+version(VibeForceInterrupt) {
+	const Interruptible = true;
+}
+else const Interruptible = false;
+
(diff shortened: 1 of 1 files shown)


s-ludwig:
Please, let's wait with any actions until we have reached a conclusion. It takes too much time to make and discuss changes over and over again.

etcimon:
Sure, I did this with hopes of moving the discussion forward with a faster pace, because this is now a bottleneck regarding the release of DMD 2.067

I can't really say if interrupt will be of any use, so I'd keep it by default.

However, the problem doesn't seem to be interrupt. The nothrow propagates through to `yieldForEvent`, which can't be nothrow, so basically the best solution I've seen is to catch and fail. If the only exception can be an InterruptException, this should be fine...

PR #999, opened 2015-03-07
Simplify @path handling

Geod24:
- Make @rootPathAttribute the default: it just scales better, there is rarely a point in having multiples interfaces at the root.
- Kill RootPathAttribute and use PathAttribute only instead: There was no real difference between @rootPath and @path, as they were mutually exclusive (path on method, rootPath on interface) and had the same behavior (e.g. they are always relative). This will obviously make user code less error-prone.
- Communicate the settings to the sub-interface.

### DIFF
--- a/source/vibe/web/common.d
+++ b/source/vibe/web/common.d
@@ -260,7 +260,7 @@ unittest
 
 
 /**
-    UDA to defeine the ContentType for methods returning an InputStream or ubyte[]
+    UDA to define the ContentType for methods returning an InputStream or ubyte[]
 */
 ContentTypeAttribute contentType(string data) 
 {
@@ -291,11 +291,20 @@ MethodAttribute method(HTTPMethod data)
 }
 
 /**
-	User Defined Attribute interface to force specific URL path n REST interface
-	for function in question. Path attribute is relative though, not absolute.
+	UDA to force a specific URL path for REST interfaces.
+
+	This attribute can be applied either to an interface itself, in which
+	case it defines the root path for all methods within it,
+	or on any function, in which case it defines the relative path
+	of this method.
+	Path are always relative, even path on interfaces, as you can
+	see in the example below.
+
+	See_Also: $(D rootPathFromName) for automatic name generation.
 
 	Example:
 	---
+	@path("/foo")
 	interface IAPI
 	{
 		@path("info2") getInfo();
@@ -305,8 +314,10 @@ MethodAttribute method(HTTPMethod data)
 	
 	shared static this()
 	{
+		// Tie IAPI.getInfo to "GET /root/foo/info2"
 		registerRestInterface!IAPI(new URLRouter(), new API(), "/root/");
-		// now IAPI.getInfo is tied to "GET /root/info2"
+		// Or just to "GET /foo/info2"
+		registerRestInterface!IAPI(new URLRouter(), new API());
 	}
 	---	
 */
@@ -319,51 +330,23 @@ PathAttribute path(string data)
 
 
 /**
-	UDA to define root URL prefix for annotated REST interface.
-	Empty path means deducing prefix from interface type name (see also rootPathFromName)
+	Will be deprecated in the next release.
+	Use @$(D path) instead.
  */
-RootPathAttribute rootPath(string path)
+PathAttribute rootPath(string path)
 {
 	if (!__ctfe)
 		assert(false, onlyAsUda!__FUNCTION__);
-	return RootPathAttribute(path);
-}
-///
-unittest
-{
-	import vibe.http.router;
-	import vibe.web.rest;
-
-	@rootPath("/oops")
-	interface IAPI
-	{
-		int getFoo();
-	}
-
-	class API : IAPI
-	{
-		int getFoo()
-		{
-			return 42;
-		}
-	}
-
-	auto router = new URLRouter();
-	registerRestInterface(router, new API());
-	auto routes= router.getAllRoutes();
-
-	assert(routes[0].pattern == "/oops/foo" && routes[0].method == HTTPMethod.GET);
+	return PathAttribute(path);
 }
 
 
-/**
-	Convenience alias
- */
-@property RootPathAttribute rootPathFromName()
+/// Convenience alias to generate a name from the interface's name.
+@property PathAttribute rootPathFromName()
 {
 	if (!__ctfe)
 		assert(false, onlyAsUda!__FUNCTION__);
-	return RootPathAttribute("");
+	return PathAttribute("");
 }
 ///
 unittest
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -79,7 +79,7 @@ void registerRestInterface(TImpl)(URLRouter router, TImpl instance, RestInterfac
 
 	string url_prefix = settings.baseURL.path.toString();
 
-	enum uda = findFirstUDA!(RootPathAttribute, I);
+	enum uda = findFirstUDA!(PathAttribute, I);
 	static if (uda.found) {
 		static if (uda.value.data == "") {
 			auto path = "/" ~ adjustMethodStyle(I.stringof, settings.methodStyle);
@@ -129,10 +129,12 @@ void registerRestInterface(TImpl)(URLRouter router, TImpl instance, RestInterfac
 					ParameterTypeTuple!overload.length == 0,
 					"Interfaces may only be returned from parameter-less functions!"
 				);
+				auto subSettings = settings.dup;
+				subSettings.baseURL = URL(concatURL(url_prefix, url, true));
 				registerRestInterface!RT(
 					router,
 					__traits(getMember, instance, method)(),
-					concatURL(url_prefix, url, true)
+					subSettings
 				);
 			} else {
 				// normal handler
@@ -180,6 +182,7 @@ void registerRestInterface(TImpl)(URLRouter router, TImpl instance, string url_p
 */
 unittest
 {
+	@path("/")
 	interface IMyAPI
 	{
 		// GET /api/greeting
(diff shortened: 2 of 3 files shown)


Geod24:
Ping @Dicebot as we discussed it.
@s-ludwig : Why is `RestInterfaceSettings` a class ? Shouldn't it be a struct ?

s-ludwig:
> Make @rootPathAttribute the default: it just scales better, there is rarely a point in having multiples interfaces at the root.

You mean `@rootPathFromName`? I don't agree here. This would be a silent breaking change and for what it's worth, *all* of my web interfaces currently have the default path of "/". We can't make this change like this without a deprecation path, but I also think that making the path inference explicit is the better idea. It just doesn't seem obvious to me that the type name would automatically become part of the path. Is that possibly different for people coming from certain other web frameworks?

Deprecation of `@rootPath` sounds fine, though.

>  Why is RestInterfaceSettings a class ? Shouldn't it be a struct ?

It's just to be consistent with all other `*Settings` types in the library (I think that `HTTPServerSettings` was the first - not really sure why I made it a class at the time). There are some low priority pros and cons for either side, but in the end none of them matters much, so I wouldn't want to go through the hassle and change everything now.

Geod24:
> You mean @rootPathFromName? I don't agree here. This would be a silent breaking change and for what it's worth, all of my web interfaces currently have the default path of "/". We can't make this change like this without a deprecation path, but I also think that making the path inference explicit is the better idea. It just doesn't seem obvious to me that the type name would automatically become part of the path. Is that possibly different for people coming from certain other web frameworks?

Yes, I meant `@rootPathFromName`.
I'm aware of the breaking change, and that's also a problem I considered. I couldn't find any "perfect" solution, beside deprecating un-marked interface in one release, then removing `@rootPathFromName` in the next, before deprecating it few releases later (but it's rather annoying on a user POV, but probably better).

As per why I want it to be the default :
The REST code generator is, ATM, a kind of gloryfied RPC system IMO.
Usually when you have a REST interface (I think Github's one of the best example), you have 2 kinds of "objects": a ressource, and a collection. Examples of collections are `/users/`, `/posts/` (for a blog), `/comments/`, `/organizationName/` (in case of Github). In those collections, you must have ressources (OFC your collection can be empty).
However, in `vibe.web.rest`, such construction is neither enforced, nor "easy" to do. It's doable, but there's a lot of gotchas. One of them is `@rootPathFromName`, as it's an opt-in strategy to this kind of design.

Again, another possibility is to deprecate un-annotated interface for X releases / X time.

> It's just to be consistent with all other *Settings types in the library [...]

Ok, makes sense.

Geod24:
Rebased.

s-ludwig:
You are right that collection/object handling still needs to be improved. I had some ideas for that, but have to think about this again to remember the details. But please let's leave this as a separate concern for now, because deprecating `@rootPath` is good and doesn't really depend on the by-name change.

One thing regarding `@rootPath`: before marking it as `deprecated` it should just get a doc comment "Scheduled for deprecation - please use $(D @path) instead.` (I need to get that written down somewhere). Directly going to `deprecated` is always annoying during the transition phase from one version to another.

Geod24:
> You are right that collection/object handling still needs to be improved. I had some ideas for that, but have to think about this again to remember the details.

I'd be interested to know, when you have time to write them down (unrelated: did you get my email?).

For the deprecation, I agree, I just didn't like the  double-deprecation (we ask to use rootPathAttribute, then we remove it).

Is this new version better ?

Geod24:
> One thing regarding `@rootPath`: before marking it as deprecated it should just get a doc comment "Scheduled for deprecation - please use $(D `@path`) instead." (I need to get that written down somewhere). Directly going to `deprecated` is always annoying during the transition phase from one version to another.

I'm not sure I understand your point. The point of `deprecated` is to have a nice way to tell your users that something is going to break. Most users won't bother to look the doc up on each release, especially something as basic as `@rootPath.

This policy is similar to what Druntime/Phobos use to do. But now they're going straight to deprecated because you can give a message with `deprecated`.

s-ludwig:
> Is this new version better ?

What I meant was that I'd like to exclude the `@rootPathFromName` change from the PR until we have reached a final conclusion. We can get the `@path`<->`@rootPath` change in now, though (with the `deprecated` <-> "scheduled for deprecation" change).

> I'm not sure I understand your point. 

A typical situation is that you have a project with dependencies to several libraries. Now it may easily be the case that library A only works with vibe.d 0.7.22, but library B requires 0.7.23. If we now immediately deprecate something in 0.7.23, there is no way to get a clean build with up-to-date dependencies. Including an additional "soft deprecation" step at least alleviates this issue.

Geod24:
Updated.

source/vibe/web/common.d
```diff
@@ -291,11 +291,27 @@ MethodAttribute method(HTTPMethod data)
...
+	of this method.
+	Path are always relative, even path on interfaces, as you can
+	see in the example below.
+
+	Note that by default, interfaces without attributes are tied to "/".
+	Since this behavior doesn't scale, it will be deprecated in the next
+	release (0.7.25), and later turned into an error (0.7.26) to allow
+	users to catch any unattributed interface and annotate them with
+	@path("/"). Later on (0.7.27), the unattributed interfaces will be
+	legal again, with a default path generated from their name.
```
s-ludwig: This paragraph should get removed for this PR.

source/vibe/web/common.d
```diff
@@ -305,8 +321,10 @@ MethodAttribute method(HTTPMethod data)
 	
 	shared static this()
 	{
+		// Tie IAPI.getInfo to "GET /root/foo/info2"
 		registerRestInterface!IAPI(new URLRouter(), new API(), "/root/");
-		// now IAPI.getInfo is tied to "GET /root/info2"
+		// Or just to "GET /foo/info2"
```
s-ludwig: Is the "/root" above ignored when `@path` is given? I think the URL should be `/root/foo/info2` when both are given like here.

source/vibe/web/common.d
```diff
@@ -319,51 +337,29 @@ PathAttribute path(string data)
...
+   Convenience alias to generate a name from the interface's name.
 
+   This alias will ultimately be deprecated and made the default.
+   It's only usable on interfaces, and is the default behavior for methods.
 
-/**
-	Convenience alias
- */
-@property RootPathAttribute rootPathFromName()
+   See_Also: $(D path) for rationale and roadmap.
```
s-ludwig: The above lines again should be removed for this PR.

source/vibe/web/rest.d
```diff
@@ -87,6 +87,15 @@ void registerRestInterface(TImpl)(URLRouter router, TImpl instance, RestInterfac
...
 		}
+	} else {
+		// Introduce this for 0.7.25. Turn it into an error in 0.7.26.
+		//pragma(msg, I.stringof~`: Unattributed interface are deprecated and will error on the next release. Use @path("/") instead`);
+
+		// On 0.7.27, it will be legal again, but with a different default (see following code):
+		/*
+		  auto path = "/" ~ adjustMethodStyle(I.stringof, settings.methodStyle);
+		  url_prefix = concatURL(url_prefix, path);
+		*/
```
s-ludwig: ditto for the else block.

source/vibe/web/rest.d
```diff
@@ -287,13 +299,19 @@ class RestInterfaceClient(I) : I
...
 				url.path = Path(concatURL(url.path.toString(), adjustMethodStyle(I.stringof, settings.methodStyle), true));
 			} else {
 				url.path = Path(concatURL(url.path.toString(), uda.value.data, true));
 			}
+		} else {
+			// Introduce this for 0.7.25. Turn it into an error in 0.7.26.
+			//pragma(msg, I.stringof~`: Unattributed interface are deprecated and will error on the next release. Use @path("/") instead`);
+
+			// On 0.7.27, it will be legal again, but with a different default (see following code):
+			//url.path = Path(concatURL(url.path.toString(), adjustMethodStyle(I.stringof, settings.methodStyle), true));
```
s-ludwig: ditto

source/vibe/web/common.d
```diff
@@ -319,51 +337,29 @@ PathAttribute path(string data)
 
 
 /**
-	UDA to define root URL prefix for annotated REST interface.
-	Empty path means deducing prefix from interface type name (see also rootPathFromName)
+	Deprecated. Use @$(D path) instead.
  */
-RootPathAttribute rootPath(string path)
+deprecated(`Use @path("path") directly.`)
```
s-ludwig: `deprecated` still there? Do you still have objections against the three-phase deprecation?

source/vibe/web/common.d
Geod24: It is. Comments describe the line under, not above.
So:
- `registerRestInterface!IAPI(new URLRouter(), new API(), "/root/");` => /root/foo/info2
- `registerRestInterface!IAPI(new URLRouter(), new API());` => /foo/info2

But maybe my comments are not clear. I'll rework that.

source/vibe/web/common.d
Geod24: I misread your comment (https://github.com/rejectedsoftware/vibe.d/pull/999#issuecomment-86201613), and thought you were okay with deprecating `@rootPath` but not `@rootPathFromName`.
So this will become nothing but a documentation commit. If I had knew, I would have pushed to get this into 0.7.23 ;)

source/vibe/web/common.d
s-ludwig: No sorry, you are right, I've skimmed to fast over this and mistook the following line for something else. Please ignore.

source/vibe/web/common.d
s-ludwig: It's not just the documentation, it also changes the return types of `rootPath` and `rootPathFromName` and adjusts the examples.

Geod24:
Update.

s-ludwig:
Thanks, LGTM!

PR #1000, opened 2015-03-07
Add HTTP digest authentication.

redstar:
This is a minimal implementation of RFC 2069.

Possible future enhancements:
- Add the client IP to the nonce
- Implement qpop
- Retrieve user passwords from .htaccess file


s-ludwig:
Great, thanks! I always wanted to implement that for a little project, but never got around to it.

Also congratulations of course to the one thousandth opened ticket ;-)

Issue #949, opened 2015-01-03
[REST] @path doesn't check for placeholders


Geod24:
The following code:
````
@rootPath("/repos/")
interface IGithubPR {
  @path(":owner/:repo/pulls")
  GithubPR getPullRequests(string owner, string repo);
}
````
Is an easily made typo at this time of the day :)
In this specific case, it results in an assertion failure:
`core.exception.AssertError@../../.dub/packages/vibe-d-master/source/vibe/inet/path.d(186): Trying to append absolute path.`

The REST generator can and should detect it, and error accordingly.

Geod24:
Fixed in #1001 (but I screwed up the title :-( )

PR #1001, opened 2015-03-07
[REST] Reorganize error handling + fix issue #949

Geod24:
A common problem I was encountering with REST was how painful it was to test failing cases (as the module should warn on interfaces error), as well as the duplication of code to have error both in client and server.
This centralize the error checking in a single CTFE function that is called from 2 static asserts. If the function returns non-null, it means there is an error (and the return is the error message).
Hence, it makes it quite easy to test issues, such as #949 which I originally intended to fix when I started this branch.

Note that `static assert (ParamNames[i].length, ..)` could never be false, as `std.traits.ParameterIdentifierTuple` would not compile (unless this has to do with an older release?). I fixed the test.
EDIT: Actually, this static assert now works in 2.067.

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -71,11 +71,13 @@ void registerRestInterface(TImpl)(URLRouter router, TImpl instance, RestInterfac
 	import std.traits : MemberFunctionsTuple, ParameterIdentifierTuple,
 		ParameterTypeTuple, ReturnType;
 
+	alias I = baseInterface!TImpl;
+
+	static assert(getInterfaceValidationError!(I) is null, getInterfaceValidationError!(I));
+
 	if (!settings) settings = new RestInterfaceSettings;
 
 	string url_prefix = settings.baseURL.path.toString();
-
-	alias I = baseInterface!TImpl;
 
 	enum uda = findFirstUDA!(RootPathAttribute, I);
 	static if (uda.found) {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Interesting idea with the `string` return. I just think it looks a bit too clever like this. I'd at least change the name to something like `getInterfaceValidationError`, because reading `static assert(!validateInterface!(I), validateInterface!(I));` had a slight WTF moment ;)

Maybe returning a small struct or tuple would be even cleaner (with the potential to safe one CTFE run), but the name change enough would be clean enough for me for an internal function.

Geod24:
> Interesting idea with the string return. I just think it looks a bit too clever like this. I'd at least change the name to something like getInterfaceValidationError, because reading static assert(!validateInterface!(I), validateInterface!(I)); had a slight WTF moment ;)
>
> Maybe returning a small struct or tuple would be even cleaner (with the potential to safe one CTFE run), but the name change enough would be clean enough for me for an internal function.

It started as a template with a `static assert` inside it... But returning a message helps alot (because `__traits(compiles, ...)` is an awful way to check for correctness). And it needs to be a function because, well, CTFE and template are rarely compatible.

I changed the name, and add `is null` for extra verbosity. You're right, `getInterfaceValidationError` is much more suited as a name :)

Geod24:
Ping

s-ludwig:
Okay, looks good. It's definitely an interesting idiom.

Just out of interest, why did you need `__traits(compiles)` with the template+`static assert` solution, but not for the CTFE string return version?

Geod24:
For testing.
If you put the assert directly inside the template, then how do you test that the generator correctly rejects invalid interface ? You're bound to use `__traits(compiles)`, and that's an horrible solution because if it breaks for a different reason, the test will pass.
Check line ~1200 for an example.

Note that my first attempt was to write an external test, but those should usually be 'black box' testing, so it required me to duplicate every single test (1 for client and 1 for server).

s-ludwig:
Okay, makes sense.

PR #1002, opened 2015-03-07
Enable default block content in Diet templates

redstar:
With Jade templates, you can define a default block content which is used if no content for a block is provided.
This is also possible with vibe.d - but an assertion prevents this.
IMHO the assertion is a cut'n'paste error. Removing it enables use of default block content.

### DIFF
--- a/source/vibe/templ/diet.d
+++ b/source/vibe/templ/diet.d
@@ -605,7 +605,6 @@ private struct DietCompiler(TRANSLATE...)
 						output.pushNode("<![endif]-->");
 						break;
 					case "block": // Block insertion place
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm, interesting. In fact this was never implemented (the `// output defaults` comments are supposed to read `// TODO: output defaults`) and it's pure coincidence that it works for the default case. I'll add back an assertion for the append/prepend cases and add some unit tests for the default one.

s-ludwig:
Okay should be working correctly now. With only the assertion removed it would always append the default block contents, even if the block was overridden. "prepend" also works, but "append" is disabled again, as that is a little more involved to implement.

PR #1004, opened 2015-03-08
Fix output of block comments

redstar:
A block comment

    body
      //
        Line 1
        Line 2

should be rendered as

    <body>
        
    </body>

Currently, the block content is treated as normal template code.

This PR fixes the output behaviour.


s-ludwig:
Thanks, you are right. When implementing this, my idea was to avoid accidentally leaking source code to the output, but I've never actually looked at what Jade does, and that makes a lot of sense, too.

PR #1007, opened 2015-03-09
Implement plain text starting with < in Diet templates.

redstar:
This is the Jade way to emit conditional comments but works
with all plain text starting with <, too.

See http://jade-lang.com/reference/comments/.


s-ludwig:
Hm, interesting, did they remove the `//if` syntax? I'm pretty sure that it came from Jade when I implemented it. Anyway, supporting this shouldn't hurt! Can you add a short unit test around line 1455 to ensure that this stays functional?

redstar:
Sure!

redstar:
According to the Jade changelog, version 1.0.0 removed support for conditional comments.

s-ludwig:
Okay, thanks! Definitely makes sense to try and stay compatible with the 1.0.0 spec.

PR #1010, opened 2015-03-12
Remove deprecation notice for index handling in vibe.web.rest

Geod24:
This had been deprecated in fe6a5de7433140660b14af28e9540b22c198c1c7 and so this message is there since July 2013 (and not relevant anymore). Time to remove it ?

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -32,22 +32,22 @@ import std.typetuple : anySatisfy, Filter;
 
 	The following table lists the mappings from prefix verb to HTTP verb:
 
-	<table>
-		<tr><th>Prefix</th><th>HTTP verb</th></tr>
-		<tr><td>get</td><td>GET</td></tr>
-		<tr><td>query</td><td>GET</td></tr>
-		<tr><td>set</td><td>PUT</td></tr>
-		<tr><td>put</td><td>PUT</td></tr>
-		<tr><td>update</td><td>PATCH</td></tr>
-		<tr><td>patch</td><td>PATCH</td></tr>
-		<tr><td>add</td><td>POST</td></tr>
-		<tr><td>create</td><td>POST</td></tr>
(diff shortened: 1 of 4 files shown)


s-ludwig:
Instead of removing just the deprecation message, the whole `index` special case should be removed. For that we'd need something like a `disable_index_special_case = false` parameter for `extractHTTPMethodAndName`, because unfortunately the `vibe.web.web` module still needs to support it (although it was never supposed to be supported there in the first place).

Geod24:
Since there was no other code, I though it was already removed. I'll take a look later this week then.

Geod24:
Updated (and added small cosmetic changes).

source/vibe/web/rest.d
```diff
@@ -32,22 +32,22 @@ import std.typetuple : anySatisfy, Filter;
...
+		$(TR $(TD add)    $(TD POST))
+		$(TR $(TD create) $(TD POST))
+		$(TR $(TD post)   $(TD POST))
+	)
 
 	If a method has its first parameter named 'id', it will be mapped to ':id/method' and
-    'id' is expected to be part of the URL instead of a JSON request. Parameters with default
-    values will be optional in the corresponding JSON request.
+        'id' is expected to be part of the URL instead of a JSON request. Parameters with default
+        values will be optional in the corresponding JSON request.
```
s-ludwig: These should be tabs instead of spaces. I'll fix that after merge.

s-ludwig:
There still seems to be a missing false template parameter at rest.d line 813 according to Travis.

s-ludwig:
Nevermind. Your fix came in just in the second I commented.

Geod24:
Yup, and that means the unittest don't cover that line. Reminds me that I never checked the coverage on Vibe.
It should be fixed now (and tabs are in).

Geod24:
Green :)

PR #1012, opened 2015-03-13
Fix parsing of hidden comments

redstar:
The hidden comment from the web site

    //-title there is no title

is actually processed as

    <//-title>there is no title<///-title>

because there is no whitespace between - and t. Jade treats //-title
as a hidden comment.

This PR fixes the parsing problem. I moved all the comment handling
code into one place, too.


s-ludwig:
Looks good, thanks!

PR #1014, opened 2015-03-15
Revert "Support const(Json) serialization and add pretty serialization support."

Geod24:
This reverts commit 87c0d41e3381fe5f15fe74e19b4e60cf12722332.

Looks like someone forget to `dub test` ;)


s-ludwig:
Arrgh. Yeah, I somehow forgot how problematic that spot used to be and naively pushed after the specific application compiled. Good that this is now covered by the unit tests...

Geod24:
Thanks for fixing ;)

PR #1017, opened 2015-03-17
Added length check to re-enabled `@path("/")`

schuetzm:
This fixes a regression introduced by #1001. The following triggered an out of bounds error:

```
interface APIv1Countries {
    import vibe.web.rest;
    @path("/") @property Bson get();
}
```

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -1168,7 +1168,7 @@ private string getInterfaceValidationError(I)() {
 			// splitter doesn't work with alias this ?
 			auto sp = pathAttr.value.data.splitter('/');
 			foreach (elem; sp) {
-				if (elem[0] ==  ':') {
+				if (elem.length > 0 && elem[0] ==  ':') {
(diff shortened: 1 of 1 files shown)


Geod24:
Oops. But wouldn't that also allow `@path("///")` ?
Maybe we should special case "/" and add a check for empty entries:

`static if (pathAttr.found && pathAttr.value != "/") {`

````
auto sp = pathAttr.value.data.splitter('/');
foreach (elem; sp) {
  if (!elem.length)
    return "%s: Path '%s' contains empty entries".format(FuncId, pathAttr.value)
  if (elem[0] == ':') {
````

This would be consistent with what `vibe.inet.path` does.
Also, it could use an unittest.

s-ludwig:
@Geod24 Sounds like a good idea. I'll merge like this and add the additional check/unit tests so that I can quickly tag a new RC.

schuetzm:
Thanks @s-ludwig, I would have added it later this evening, but I'm busy now.

PR #1019, opened 2015-03-17
Update libasync dependency with memutils

etcimon:
This update adds memutils as a dependency. I intend to use it where libasync and its driver use HashMap, Array, FreeListRef and other manual memory operations. I've already started separating the vibe.d manual memory management from the libasync memory management with the last pull regarding DMD 2.67.0 compatibility fixes.

The new memory management is mostly based on your code, but it adds Phobos containers including red black tree and array, along with improvements in user experience for reference counting and object embedding (`alias this T` => `mixin Embed!T`). This will probably be used to implement the timers directly in libasync. It also implements scoped pools which are used to override the GC in tree-like structures such as [DOM, Json and XML parsing](https://github.com/etcimon/spd). This currently works using `alloc!T` instead of `new T` and pools are created in scope using `auto p = ScopedPool();`, freeing the memory as they go out of scope and calling destructors properly. As for appenders, they are effectively replaced by a `Vector!T` in Botan, and what's left is to make them copied on the scoped pools or GC with `.copy()`. I'm still unsure how I'm going to handle "Deep Copy", and overriding the duplication process.

I'd like to eventually use this for the redis client, hibernated and other protocols that currently rely on the GC.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -38,7 +38,7 @@
 		},
 		{
 			"name": "libasync",
-			"dependencies": {"libasync": "0.7.0" },
+			"dependencies": {"libasync": "0.7.1" },
 			"versions": ["VibeUseNativeDriverType", "VibeLibasyncDriver"]
 		},
 		{
--- a/dub.json
+++ b/dub.json
@@ -38,7 +38,7 @@
 		},
 		{
 			"name": "libasync",
-			"dependencies": {"libasync": "0.7.1" },
+			"dependencies": {"libasync": "0.7.1"},
 			"versions": ["VibeUseNativeDriverType", "VibeLibasyncDriver"]
 		},
 		{


s-ludwig:
Thanks!

Can you give a short summary of `Embed!T`? I'd *really* like to get most of the memutils stuff in Phobos before using it extensively (especially in the public API), because doing the switch later will probably be a nightmare. It's a shame that Andrei's allocator code is still not pushed to Phobos (at least the low level part).

etcimon:
> Can you give a short summary of Embed!T?

It basically pipes the calls to the operators through to the underlying object, because the main problem with `alias this T` is the inability to use it as a foreach loop subject, to slice it, to use indexes on it, etc. The `RefCounted(T)` object is implemented with the `mixin Embed!T` in its definition, which allows e.g. `RefCounted!Vector arr` to have support for `arr[idx]` or `foreach(char c; arr)`.

> I'd really like to get most of the memutils stuff in Phobos before using it extensively (especially in the public API), because doing the switch later will probably be a nightmare.

I'd like to have it there as well, I'm just a bit worried about the size of it and the impact that it would have on Phobos as a whole. I see phobos/druntime as "too big to change" right now. E.g., the `alloc!T` function would probably be better off with a solution in DMD or druntime, because it touches elementary allocation concepts. 

So, I wouldn't be too hopeful. For now I'd keep it delicately on the surface (out there with the small libraries) to avoid making too many waves with it unless there's some community pressure to push it in.

> It's a shame that Andrei's allocator code is still not pushed to Phobos (at least the low level part).

I created memutils with the idea of having only 3-4 allocators, so that the interface remains really simple to use. I think Andrei's allocator code couldn't be used with it, because it allows to shape it at compile-time which is very difficult to tweak in low-level code.

s-ludwig:
>> Can you give a short summary of Embed!T?

> It basically pipes the calls to the operators through to the underlying object, because the main problem with alias this T is the inability to use it as a foreach loop subject, to slice it, to use indexes on it, etc. The RefCounted(T) object is implemented with the mixin Embed!T in its definition, which allows e.g. RefCounted!(Vector!char) arr to have support for arr[idx] or foreach(char c; arr).

I see. But hm, I'm pretty sure that  `alias this` is supposed to forward all operators. Is there a DMD bug report for this/those?

> I see phobos/druntime as "too big to change" right now.

Too big for fundamental, breaking changes, but additions and backwards compatible changes (such as adding optional `allocator` parameters to certain functions shouldn't be a problem.

> I think Andrei's allocator code couldn't be used with it, because it allows to shape it at compile-time which is very difficult to tweak in low-level code.

It has a class based generic wrapper as far as I recall. That could be used to hide any compile time fiddling.

etcimon:
> I see. But hm, I'm pretty sure that alias this is supposed to forward all operators. Is there a DMD bug report for this/those?

I didn't think so :-p so I guess this should count as a workaround then!

> Too big for fundamental, breaking changes, but additions and backwards compatible changes (such as adding optional allocator parameters to certain functions shouldn't be a problem.

I think that would be a good suggestion.

> It has a class based generic wrapper as far as I recall. That could be used to hide any compile time fiddling.

Yes, that's probably the best way to go, I think std.allocator would be a great backend. 

I won't have time to enhance the memutils library for a few months because I'm yet a few months behind schedule on some proprietary deliverable I'm trying to achieve, but I think after a bit more work I would love to merge it into phobos along with libasync (you also have some MIT licenses in there so you'd have to give an OK to switch to Boost).

I'm going to be writing some serious additions to vibe.d now, the [libhttp2](https://github.com/etcimon/libhttp2) is about 1-2 days from being wholly tested and then it's integration time for HTTP/2 and Botan (with probably ALPN).

After that, I need vibe.d to compile into a shared library with a DLL router (something in the global namespace that allows to load/unload lib-grouped handlers).

PR #1027, opened 2015-03-20
Add rawResume

etcimon:
The HTTP/2 driver manages stream lifetime similar to a TCP connection. I have a read/write event loop and tasks opened for every subsequent inter-connection streams. When these tasks need to write through the HTTP/2 protocol, they might yield just like within the TCP Connection.

See here to see how the task is stopped: 
https://github.com/etcimon/vibe.d/blob/c064f3bafc43ea824206d2ecbe2eba4ea0163012/source/vibe/http/http2.d#L250

And how the task is resumed once the buffers are flushed:
https://github.com/etcimon/vibe.d/blob/c064f3bafc43ea824206d2ecbe2eba4ea0163012/source/vibe/http/http2.d#L88

I want the read/write loops to leave those tasks untouched, but I don't want to acquire locks in the process of context switching because those are guaranteed to be single-threaded operations.

For this reason, I'd like to add the `rawResume` feature which schedules the stream resumption before the event loop blocks in the OS.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -512,6 +512,17 @@ void rawYield()
 }
 
 /**
+	Resumes a task that was paused with $(D rawYield) on the next run of the
+	event loop.
+*/
+void rawResume(Task t, Exception ex = null)
+{
+	CoreTask ct = cast(CoreTask)t.fiber;
+	ct.m_exception = ex;
+	s_yieldedTasks.insertBack(ct);
+}
+
+/**
 	Suspends the execution of the calling task for the specified amount of time.
 
 	Note that other tasks of the same thread will continue to run during the


s-ludwig:
The usual primitives for these kinds of tasks are `TaskMutex`, `TaskCondition`, `ManualEvent`, or `vibe.core.concurrency.send`. `TaskPipe` may also be a possibility. But generally, `rawYield` and `resumeTask` should never appear outside of the event driver and the driver code code.

etcimon:
> Edit: replied a bit too fast, w.r.t locks. But still this is something that should be done using a higher level primitive.

It seems a little wrong to use a high-level primitive for this purpose, because these `HTTP2Stream`s are essentially the same as a `TCPConnection`. Considering the slow process in sending a notification through the OS event loop, I don't really think ManualEvent is the right way to go.

I'm essentially trying to achieve a ThreadLocalManualEvent like we talked earlier, but I'm not sure how you would implement that.

s-ludwig:
BUt this reminds me that `ManualEvent` should have an optimization to not use the event loop at all if the respective tasks are in the same thread, but use `DriverCode.yieldAndResumeTask` instead (in `.emit()`).

etcimon:
> but use DriverCode.yieldAndResumeTask instead (in .emit()).

I can't find this function, did you write an implementation for it or is it an idea at the moment? 

I would implement it by pushing the ManualEvent waiter in `vibe.core.core.s_yieldedTasks` and then calling `vibe.core.core.yield()`, is this right?

s-ludwig:
Yes, that's about it. It's currently still in a branch, because I want to defer the related changes to the next release: https://github.com/rejectedsoftware/vibe.d/blob/466751cd581bdef90e9fcbf97a580e2d495152a5/source/vibe/core/core.d#L1083

etcimon:
Ok, seems the same as `rawResume` except that you're forcing emit() to yield the task. Wouldn't that count as an unexpected side effect?

Also, it seems that HTTP/2 and Botan TLS could provide their own buffer to TCPConnection. I'm thinking of making the TCPConnection buffer initialize lazily in libasync driver, so that `read(ubyte[] dst)` used with a `dst.length >= WindowSize` can set the buffer in the object for the kernel to write to while the task is waiting. If this buffer goes over capacity, it would fall back on initializing the circular buffer.

Also, I would need a setWindowSize/getWindowSize in the `TCPConnection` because HTTP/2 has its own flow control implementation. You think it would be a good addition?

etcimon:
Also, I don't think the presence of high-level primitive should restrict users from using low-level "Hacks" in an effort of optimizing its code. It's like inline assembly, it should be there even for the slight chances that it could improve the program when properly used.

s-ludwig:
The idea for vibe.d is to present an interface to the library user, where all the details of the underlying I/O and threading model are abstracted away. It should, for example still be possible to replace fibers by threads and synchronous I/O with synchronous I/O without breaking any user code. This could be important for certain platforms or in certain library environments.

Anything that semantically relies on the fiber model is thus nothing for the public API. `rawYield` is an exception that was necessary for some hacks in the past, but I'd even go as far to remove that function before 1.0.0. Basing such kinds of inter-task control on more generic primitives makes the code more general and more robust.

The only reason that I can see to not do this would be performance issues, but those can be resolved. Using `yieldAndResume` in `emit()` would be absolutely OK from a semantic perspective, but you are right that it may in certain cases not be the intended behavior from a performance point of view, and something that doesn't yield, like your `rawResume`, might be more appropriate. However, such a function should be part of `DriverCore` and not part of the user facing API. Maybe like the following:
 - `DriverCore.resumeTask` - directly resumes the task, can only be called by the event loop
 - `DriverCore.yieldAndResumeTask` - directly resumes the task and in turn yields the current one
 - `DriverCore.scheduleResumeTask` - schedules the task for resumption, but continues the current one

or possibly using a control enum instead:
```
DriverCore.resumeTask(Task task, Exception exception = null, TaskResumeMode mode = TaskResumeMode.immediate);
enum TaskResumeMode {
    immediate,
    yield,
    schedule
}
```

etcimon:
The new resumeTask would seem good for any other kind of user protocol. However, HTTP/2 is currently the most important solution towards scaling up as much as possible on web servers. A decent implementation would have to make the HTTP/2 backend lock-less, and the mutex would definitely make a difference in benchmarks. We could add it to the core and allow it to access the `DriverCore` by annotating the accessor with a `package` visibility? I'd need it to provide its own buffer as well to TCPConnection, to save a 64kb per connection.

s-ludwig:
If you directly use `ManualEvent` instead of a task mutex/condition and `ManualEvent` uses `scheduleResumeTask` instead of pushing an event to the event loop, there should be no difference in performance, AFAICS. BTW, what do you mean by making the HTTP/2 backend lock-less? You have to implement *some* kind of wait operation, which could easily qualify as a kind of lock. However, vibe.d is always lock-less in the classical thread based sense. `TaskMutex` only uses a CAS operation to be thread-safe.
Making the TCP receive window size configurable would be good to have. But is there a reason why you can't use the buffer that is already provided to `TCPConnection.read` (i.e. what kind of API change do you need)?

etcimon:
> However, vibe.d is always lock-less in the classical thread based sense. TaskMutex only uses a CAS operation to be thread-safe.

I think it's worth a try. I was mostly concerned about this: https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/core/drivers/libasync.d#L760

> But is there a reason why you can't use the buffer that is already provided to TCPConnection.read (i.e. what kind of API change do you need)?

I only need a `void read(ubyte[] dst, ref int len)`, which yields like `leastSize` but also asks the event loop to fill the `dst` before resuming. The task will either yield here or in the HTTP/2 event loop, because there's a "pause" feature that blocks the connection (which is why `ref int` is used). However, the buffer should never be exceeded because the protocol has a flow control, there are `WINDOW_UPDATE` frames sent to the remote peer to indicate how much of the buffer was read and trigger more data be sent.

The reason I need this is because it's possible to make a save here, the reader task will have to create its own buffer anyways with the exact window size and has nothing else to wait for. Might as well make the kernel write to it directly. I'll make the same optimization in the TLS library, and with the growing buffer I can guarantee a minimal size of 2kb buffer per connection which grows up to windows_size.

> Making the TCP receive window size configurable would be good to have

Good, I'll add that, it will allow the flow control to work better (the WINDOW_UPDATE frame works on connection and on stream levels)

etcimon:
Closed because I opened a new one more relevant

PR #1029, opened 2015-03-22
Fix exitEventLoop and synchronization unittest issues in libasync

etcimon:
The `assert(!m_queue)` assumes that the task wasn't initialized yet if the event loop isn't running. Now that the event loop can be turned on and off, this isn't true anymore.

The `yield()` was added in the test because the Task unlocking mechanism doesn't generate yielding in the libasync driver, nor does it trigger a kernel signal to pause the thread and process the locking queue after an unlock. This is voluntary behavior, the TaskMutex lock waiter queue will be processed at the next run of the event loop and any lock-holding tasks will resume at that time. Yielding at ManualEvent.emit() would (maybe) process those locks instantly (depending on the ordering of tasks to be resume) but may also create unnecessary performance issues.

Finally, the `getEventLoop().exit()` was in `exitEventLoop()` by mistake, I was under the impression that it would only be called when the program must exit, thus the underlying file descriptors were being destroyed there. Vibe applications can turn on and off the event loop safely now.

### DIFF
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -354,6 +354,7 @@ private void runMutexUnitTests(M)()
 		assert(m.m_impl.m_locked);
 		t1.join();
 		assert(!t1.running && t2.running);
+		yield();
 		assert(m.m_impl.m_locked);
 		t2.join();
 		assert(!t2.running);
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -964,7 +964,6 @@ private class CoreTask : TaskFiber {
 					debug if (s_taskEventCallback) s_taskEventCallback(TaskEvent.start, handle);
 					if (!s_eventLoopRunning) {
 						logTrace("Event loop not running at task start - yielding.");
-						assert(!m_queue);
 						.yield();
 						logTrace("Initial resume of task.");
 					}
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -120,7 +120,6 @@ final class LibasyncDriver : EventDriver {
 		while(!m_break && getEventLoop().loop()){
 			getDriverCore().notifyIdle();
 		}
-		getEventLoop().exit();
 		m_break = false;
 		logInfo("Event loop exit", m_break);
 		return 0;
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -540,6 +540,7 @@ final class SyslogLogger : Logger {
 		}
 
 		assert(msg.level >= LogLevel.debug_);
+		import std.conv : to; // temporary workaround for issue 1016 (DMD cross-module template overloads error out before second attempted module)
 		auto priVal = (m_facility * 8 + syslogSeverity).to!string();
 
 		alias procId = NILVALUE;


source/vibe/core/sync.d
```diff
@@ -354,6 +354,7 @@ private void runMutexUnitTests(M)()
 		assert(m.m_impl.m_locked);
 		t1.join();
 		assert(!t1.running && t2.running);
+		yield();
```
s-ludwig: Can you add a comment akin to the one for the other yield here? `// give t2 the chance to acquire the lock`

s-ludwig:
Thanks for making the PR. Regarding the `assert(m_queue);`, it doesn't really have anything to do with running the event loop (multiple times), but can also occur within a single loop. I'll replace this part later with code to remove the old entry from `m_queue`, because this can have unwanted effects on task execution order (even if that is deliberately not specified, but we need it for the unit tests).

We'll have to decide how to go about `ManualEvent.emit` at some point. Maybe we can circumvent the possible performance issues introduced by noop-`yield`s and can in fact make `yield`ing the default.

etcimon:
The comment would be:  // Gives time for the event loop to notify t2 (TaskMutex.lock waiters) that m (the TaskMutex) was unlocked

Not sure if I should make a pull for it though =)

> We'll have to decide how to go about ManualEvent.emit at some point. Maybe we can circumvent the possible performance issues introduced by noop-yields and can in fact make yielding the default.

I don't think hidden context switching is an intuitive behavior. In HTTP/2 for example, if `emit()` causes a yield on every write, the stream buffers will flush every write and the stream's WINDOW_UPDATE behavior will be broken.

etcimon:
Maybe the `yield` could happen in the `TaskMutex.unlock()`, and I would use `ManualEvent` directly to schedule the r/w loops.

s-ludwig:
> The comment would be: // Gives time for the event loop to notify t2 (TaskMutex.lock waiters) that m (the TaskMutex) was unlocked
>
> Not sure if I should make a pull for it though =)

Sorry, I completely forgot about that before merging ;) I'll add that.

WRT wording, I'd put the comment there in a more abstract way - give t2 some execution time to actually perform the lock acquisition - instead of the more technical and driver dependent way. I'll make a quick commit.

s-ludwig:
> Maybe the yield could happen in the TaskMutex.unlock(), and I would use ManualEvent directly to schedule the r/w loops.

My thought was to maybe add a boolean parameter to `emit()` to control if the waiters should be readily resumed. `TaskMutex` could set this to true, while `TaskCondition` would possibly be better off using `false` by default. I'd then also use `ManualEvent` directly for the r/w loops.

etcimon:
> My thought was to maybe add a boolean parameter to emit() to control if the waiters should be readily resumed.

Awesome. Last obstacle before I have the full HTTP/2 module figured out is the custom buffers in `TCPConnection`, I'd either wait in `leastSize` or in a new `read(ubyte[] buffer, ref int)` which always resumes immediately when data is available. Should it be part of `InputStream` altogether?

s-ludwig:
Another `read` variant sounds good. The question is just which is the optimal name and signature. I'd tend to choose a different name to make the different semantics explicit, maybe `readChunk`? The signature is more tricky..

- `void (scope ubyte[] buffer, ref int size)` is safe and well contained, but it would be nice to directly return a slice
- `size_t (scope ubyte[] buffer)` looks a bit nicer, but is also slightly easier to misuse
- `ubyte[] (ubyte[] buffer = null)` is nicely general and makes the supplied buffer optional, but is not safe (the supplied buffer could be allocated on the stack and looses it's scope information upon return)
- `void (ref ubyte[] dst)` kind of works, but feels a little unintuitive
- Extending DIP25 for arrays would be nice, but even if it would, we can't count on that for at least a year
- anything else?

`ubyte[] (ubyte[] buffer = null)` would be my favorite if it weren't for the un`@safe`ty.

etcimon:
Keep in mind that there's a `pause` option in the protocol that allows the reader to block when consuming data. This means that once the `DataProvider` will decide to resume the session minutes or hours later, the buffer will be checked and consumed for `ref int size` data.

In this case, maybe a `ref ubyte[] read(ubyte[] buffer = null)` could do? The TCPConnection would read the buffer into its struct and refer to it at any moment. If the buffer overflows while the reader has paused (control frames are not flow controlled), this would allow to move the data into a growing circular buffer and reference a `peekDst` into this member buffer array.

etcimon:
What we can also do is register a buffer using `ref ubyte[] setBuffer(ubyte[] buffer);`, and then call `waitForData()` expecting the referenced array to contain the slice.

etcimon:
Here's an idea of what I'm looking for:
https://github.com/etcimon/vibe.d/blob/2d1d657b800ed1e8c4e5256fca60dd3931fb2eda/source/vibe/http/http2.d#L459

Everything is still a big draft so don't pay attention to the rest. The SSL stream would be checked through casts to see if it supports custom buffers as well.

s-ludwig:
The problem with `setBuffer` is that not only the interface is not safe, but also the functionality itself. Passing a stack local buffer would be complete `@system` land in that case, while the `readChunk` version with a returned slice could in theory be fixed by the language.

etcimon:
I would opt for a delegate in that case.

`ubyte[] delegate() read(ubyte[] buffer);`

Would it be a little safer, if the delegate provides the slice when `waitForData` returns?

etcimon:
Meh, it looks a lot like your proposal. If buffer is null, just return the last slice?

PR #1039, opened 2015-03-26
Allocate arrays on the heap to avoid slower GC marking

etcimon:
It might sound contradictory but this is an actual issue until we get a precise GC

The objects affected are used fairly frequently. Having an array inlined in an object with pointers will not allow NO_SCAN to be marked on them, which forces the GC to scan their contents and make scanning very slow compared to not having them there. Yes, scanning takes orders of magnitude longer with those buffers there, especially for benchmarks that make a copy on every connection.

Moving those arrays in a separate allocation fixes this quite well.

Lastly, the peer address in libevent has the lifetime of the object. This is unsafe in the context of an API assumed to be garbage collected, as it produces a segmentation fault if the user needs to access it after the GC collects TCPConnection and gives back the memory to the OS.

### DIFF
--- a/source/vibe/core/drivers/libevent2_tcp.d
+++ b/source/vibe/core/drivers/libevent2_tcp.d
@@ -58,16 +58,15 @@ package final class Libevent2TCPConnection : TCPConnection {
 		bool m_timeout_triggered;
 		TCPContext* m_ctx;
 		string m_peerAddress;
-		ubyte[64] m_peekBuffer;
 		bool m_tcpNoDelay = false;
 		bool m_tcpKeepAlive = false;
 		Duration m_readTimeout;
-		char[64] m_peerAddressBuf;
 		NetworkAddress m_localAddress, m_remoteAddress;
 	}
 
 	this(TCPContext* ctx)
 	{
+		static char[64] peer_address_buf = void;
 		m_ctx = ctx;
 
 		assert(!amOwner());
@@ -78,9 +77,9 @@ package final class Libevent2TCPConnection : TCPConnection {
 		void* ptr;
 		if( ctx.remote_addr.family == AF_INET ) ptr = &ctx.remote_addr.sockAddrInet4.sin_addr;
 		else ptr = &ctx.remote_addr.sockAddrInet6.sin6_addr;
-		evutil_inet_ntop(ctx.remote_addr.family, ptr, m_peerAddressBuf.ptr, m_peerAddressBuf.length);
-		m_peerAddress = cast(string)m_peerAddressBuf[0 .. m_peerAddressBuf.indexOf('\0')];
-
+		evutil_inet_ntop(ctx.remote_addr.family, ptr, peer_address_buf.ptr, peer_address_buf.length);
+		m_peerAddress = cast(string)peer_address_buf[0 .. peer_address_buf.indexOf('\0')].idup;
+		
 		bufferevent_setwatermark(m_ctx.event, EV_WRITE, 4096, 65536);
 		bufferevent_setwatermark(m_ctx.event, EV_READ, 0, 65536);
 	}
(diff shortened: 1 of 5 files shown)


s-ludwig:
You are right about the safety issue for `TCPConnection`, there is a severe need for something like [DIP69](http://wiki.dlang.org/DIP69) to make this safe without being forced to resort to the GC. 

But for the scanning performance issue, putting back all of the GC uses that have been carefully avoided just like that can't be the solution. It should instead be based on `FreeListRef` or similar, so that we don't again get GC allocations (and thus collections) in the core code.

Wasn't there an initiative to integrate the partially precise GC by Rainer Schütze months(years?) ago?

etcimon:
> Wasn't there an initiative to integrate the partially precise GC by Rainer Schütze months(years?) ago?

Yep, essentially it's this https://github.com/rainers/druntime/commit/1125f441cdf1f9b2ea881c0c31f993709cbbdb0e#diff-6f1ab0423fff9dcd084ecf9a677dc426R2243

Although I can't find any of it in the current GC code. It still has a few performance issues I think.

As for the avoided GC allocations, I think the savings are lost, they're even worsening performance by *a lot* from the flooding of data scheduled to be marked. The solution I'm now thinking of would be to use `allocArray` in `this()` and `freeArray` in `~this()` with the manual allocator, because the GC now seems pretty good at calling destructors for classes. Would that be preferable?

s-ludwig:
> As for the avoided GC allocations, I think the savings are lost, they're even worsening performance by a lot from the flooding of data scheduled to be marked.

At least if the GC has to run often because the GC pressure from other places is high... yeah but it obviously looks like we have to do something. BTW, did you do any comparisons of how much this change actually gains under which circumstances (i.e. once tested for a simple static page with `VibeManualMemoryManagement` and once for a normal full blown application that GC allocates itself)?

It would be great add a simple performance test that gets executed for every commit and PR, is there any service that provides this? Maybe we'd have to come up with an own solution based on an EC2 instance that is fired up when needed (does that have reliable performance at all?). Then we could also add an additional error message to PRs to catch performance regressions.

> The solution I'm now thinking of would be to use allocArray in this() and freeArray in ~this() with the manual allocator, because the GC now seems pretty good at calling destructors for classes.

Should be OK for the address buffer, but for the buffers contained in structs I'd use a RC based solution to make it more difficult to introduce memory allocation bugs later.

etcimon:
> It would be great add a simple performance test that gets executed for every commit and PR, is there any service that provides this?

I'd think it's possible but a good benchmarking metric would be more useful like this one: https://github.com/s-ludwig/FrameworkBenchmarks/tree/master/vibe.d

If it reports in a mysql/redis/mongodb server it's easy to build a website with the results.

Has there been a vibe.d github API library so far? The basic idea is for a periodic timer to check the latest commit ID, when different rebuild vibe.d and the benchmarks utility through dub and execute it.

It seems like a few weeks of work, I would have waited until done with more optimizations, I'm currently refactoring the HTTPClient and adding HTTP/2, CookieStore support and much more settings (ping intervals, client-specific SSLContext, etc).

> Should be OK for the address buffer, but for the buffers contained in structs I'd use a RC based solution to make it more difficult to introduce memory allocation bugs later.

Sounds good

s-ludwig:
Actually, I have almost everything necessary already there. I'd just have to put things together and somehow push the results to GitHub. The only issue is that my CI system currently works directly using git. I'd have to add additional polling of the GitHub API to get the pull requests.

So far I've only needed maybe two or three different requests to GitHub, so I didn't start a library for that. There is also no existing library that I know of.

But let's not rush this and better concentrate on actual improvements for now. A simple manual benchmark should be sufficient for now, just to get a feel for the actual numbers. A more systematic approach only becomes important when things get close for a 1.0.0 release.

Geod24:
On the Github API topic: I started something few month ago but never
finished it because I had more pressing issues. It should be possible to
build it using classes that inherit from RestInterfaceClient, for
convenience. But I think most of that can be automatized (generation of the
interface and the data structure), so I'd rather build the tools to do it
than do a by-hand convertion.

WebFreak001:
are these changes still relevant? We also have a precise GC we could optionally select now too, though not sure what breaking changes that would come with.

PR #1047, opened 2015-03-31
Vibe.internal.meta.codegen fixes

Geod24:
- Fixed ParameterTuple to have a saner API and take default parameters of different types, and even functions! But you won't be able to access it :-(
- Added CloneFunctionDecl, which clones a declaration (think CloneFunction without body).
- Added ability to CloneFunction to clone UDAs as well.

### DIFF
--- a/source/vibe/internal/meta/codegen.d
+++ b/source/vibe/internal/meta/codegen.d
@@ -189,7 +189,7 @@ unittest
 
 /// Returns a Tuple containing a 1-element parameter list, with an optional default value.
 /// Can be used to concatenate a parameter to a parameter list, or to create one.
-template ParameterTuple(T, string identifier, TUnused = void)
+template ParameterTuple(T, string identifier, DefVal : void = void)
 {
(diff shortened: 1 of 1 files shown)


Geod24:
@s-ludwig : Anything holding this back ?

s-ludwig:
Sorry, just me being held back by other things. But apart from the `CloneFunction` question, LGTM (as well as #1048).

Geod24:
> Is there any reason why `CloneFunction` does not use `CloneFunctionDecl` internally?

If you're thinking about factoring out the string generation and just adding the semicolon / body, I'd rather not do that: it's unlikely that those will change in the future, and if they do, they're close enough and trivial enough to make changes straightforward, and errors easy to spot.

Or do you have an idea to implement that using the function type ?

s-ludwig:
Sounds reasonable, they're short enough. Merging.

Geod24:
Thanks ! :)

Issue #1045, opened 2015-03-29
[REST](14375) registerRestInterface fails with templated class / interface


Geod24:
I tried to implement policy-based REST interface for userman, but:
https://issues.dlang.org/show_bug.cgi?id=14375

We *could* roll out our own MemberFunctionTuple (and just exclude empty string) for the time being. I'll investigate next week.

Geod24:
@s-ludwig : Maybe we can create `depends-dmd` / `depends-druntime` / `depends-phobos` label for those kind of bug. It'll be easy to check for resolution every ~6 months.

Geod24:
Note: This is fixed in DMD 2.067.1.

PR #1048, opened 2015-03-31
Workaround #1045 / 14375

Geod24:
Depends on #1047


Geod24:
Damn you 2.065 ! I disabled the test for this version: I get forward reference errors, and I doubt there's much I can do. Since no one reported the issue before, I assume no one is interested in templated interfaces.

Geod24:
@s-ludwig : I know it's a hack, and I'm looking to fix the cause in DMD, but that could be cool to have this.

PR #1057, opened 2015-04-03
Allows chaining methods without repeating url_match each time.

smcmurray:
This allows code like so:
````
router.route("/branch").get(&getHandler).put(&putHandler);
````

Caveat: It would allow the following:
````
router.get("/branch", &getHandler).put(&putHandler);
````
which might behave differently than expected.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -114,12 +114,9 @@ interface HTTPRouter : HTTPServerRequestHandler {
 
 		return this;
 	}
+
 	/// Adds a new subroute without a handler for requests matching the specified pattern, regardless of their HTTP verb.
-	final HTTPRouter route(string url_match) {
-		HTTPRouter r = new HTTPRouter(url_match);
-		any("*", r);
-		return r;
-	}
+	HTTPRouter route(string url_match);
 }
 
 
@@ -232,6 +229,13 @@ final class URLRouter : HTTPRouter {
 				r = m_routes.getTerminalData(i);
 			return routes;
 		} else return m_routes;
+	}
+
+	/// Adds a new subroute without a handler for requests matching the specified pattern, regardless of their HTTP verb.
+	URLRouter route(string url_match) {
+		URLRouter r = new URLRouter(url_match);
+		any("*", r);
+		return r;
 	}
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by s-ludwig** [comment]
I see two issues with this approach:
1. `get(&handler)` (to me) intuitively seems to mean `get("*", &handler)` instead of `get("/", &handler)`
2. Since `route()` creates a new router instance, it can have has performance implications

I can see how the `route(...).get(...)` syntax can be nice though, so I'd propose to use a solution similar to this one:
```
class HTTPRouter {
    // ...

    auto route(string url_match)
    {
        static struct Route {
            private HTTPRouter router;
            private string match;
            ref Route get(HTTPServerRequestHander handler) { router.get(match, handler); }
            // ...
        }
        return Route(this, url_match);
    }
}
```

**Comment by smcmurray** [comment]
@s-ludwig Thanks. I've only spent about 10 minutes in vibe.d code, so I was improvising a lot, hoping somebody in the know would catch my drift. It looks like you caught it very well. I considered creating a new struct or class along the lines of your sample, but I didn't know what already exists in the codebase and I didn't want to cause a conflict. I also thought it would be best to tie into an existing concept.

I'm not sure how Route: HTTPRouter is better than what I did.

Also, you could drop the get,put,etc overloads that lack a url_match. They are not a huge convenience, and can cost clarity.

The second example in http://vibed.org/api/vibe.http.router/URLRouter shows:
````
auto c1router = new URLRouter("/component1");
c1router.get("/", &showComponentHome);  //paraphrasing the contents of registerComponent()
c1router.get("/users/:user", &showComponentUser);  //ditto
auto mainrouter = new URLRouter;
mainrouter.get("/", &showHome);
mainrouter.any("*", c1router);
```

In this example, the author is creating a new router instance subject to your second issue listed above. route() is just a convenience method to do this inline:
````
auto router = new URLRouter;
router
  .get("/", &showHome)
  .route("/component1")
    .get("/", &showComponentHome)
    .get("/users/:user", &showComponentUser);
````

s-ludwig:
Chaining routers like that works as expected and is generally not an issue. However, the `route(...).get(...)` syntax would facilitate creating a multitude of chained routers, which then can become a performance issue.

In my proposal, there is no router chaining going on, but all routes are registered directly with the main router. But the idea about using a `class` instead of a `struct` for `Route` indeed wasn't really that good after all. Better to go with the struct and only define the methods without a `url_match` parameter.

s-ludwig:
Implemented the `struct` approach in 23742d065c05f37eb87f7bd6c37026356784420f (commit message is off by accident due to squashing commits).

Issue #1061, opened 2015-04-07
Inadequate documentation on routing


smcmurray:
http://vibed.org/docs#http-routing has almost no information about how to construct a proper pattern for route matching. The sum total is two sentences:

>  It supports static path matching, variable placeholders and wild-cards. Any matched variable will be available as an entry in the params dictionary.

What are variable placeholders? What are wild-cards?

Can I match "/users?:name"?
Can I match "user/:last,:first"?
Can I match regex(r"/user/(\d+)")?
Can I match "/user/\*/dept"?
Can I match "/user/\*/dept/\*"?

PR #1062, opened 2015-04-07
Allow regex in routes

smcmurray:
Adds regex routes as suggested in issue #1061.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -366,41 +366,42 @@ private struct Route {
 	string pattern;
 	HTTPServerRequestDelegate cb;
 
+  this(HTTPMethod method, string url_match, HTTPServerRequestDelegate cb){
+    /***
+    * Convert the url_match into a regex before storing to minimize compiles.
+    * The regex should preserve variable names that start with ":".
+    * The regex will be terminated by any "*" within the string,
+    * so if the "*" was intended as a regex "*",
+    * use the regex version of this method instead.
+    ***/
+    this.method = method;
+    this.cb = cb;
+    replaceAllInto!(p => "(P<"~p[1]~">(?!/).*)")(this.pattern, "^" ~ replaceFirst(url_match, regex("\*.*"), ""), regex(":((?!/).+)"));
+    this.pattern = regex(this.pattern);
+  }
+
+  this(HTTPMethod method, Regex url_match, HTTPServerRequestDelegate cb){
+    this.method = method;
+    this.cb = cb;
+    this.pattern = url_match;
+  }
+
 	bool matches(string url, ref string[string] params)
 	const {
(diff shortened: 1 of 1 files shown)


smcmurray:
I know this is broken code. I'm hoping that somebody who knows D will understand the intent and fix it or replace it with better.

s-ludwig:
The problem is that while the implementation in the old naive router implementation is straightforward, implementing this for the new DFA based router is a lot more involved. Apart from that it doesn't mix with the wildcard support, so there would have to be something that tells regex based routes apart from normal ones.

Another possibility, which other frameworks offer, is to only support the regex for placeholder variables:
```
router.get("/items/:(var:[0-9]+)/something", ...);
```
It would then be implemented as step after the basic route matching.

smcmurray:
@s-ludwig As indicated in issue #1061, I don't have much to go on for how routes are supposed to work. I think I imitated the behavior that was in the code for wildcards, if by "wildcard" you mean "*". I don't know anything about a new DFA-based router. Is it documented somewhere?

I did note issues with "*" in the comments of the code. The Route constructor is overloaded to accept both a string and a regex.
````
router.get(regex(r"/items/(?P<var>[0-9]+)/something"), ...);
```
should work just fine.

s-ludwig:
Okay, using separate overloads would indeed avoid the issues in mixing the two approaches.

The DFA (erroneously called "tree" in the code, it's really a directed graph) is used by default instead of the linear `Route` probing: https://github.com/rejectedsoftware/vibe.d/blob/1908505af05154b172f97baca8624c2ff953cf60/source/vibe/http/router.d#L110

I'll fix the documentation later, but currently wildcards and placeholder variables are relatively limited (but useful enough for all usual needs):
1. Wildcards may only be at the end of the match string and match *any* string: `/some/path*` for example matches `/some/pathxyz` or `/some/path/asdasd`)
2. Placeholders must be at the end of the match string, or have to be followed by a slash: `/foo/:placeholder/bar` or `/foo:placeholder` are valid, but `/foo/:placeholder1-:placeholder2` doesn't work.

Geod24:
Le 7 avr. 2015 14:05, "Sönke Ludwig" <notifications@github.com> a écrit :
>
> Another possibility, which other frameworks offer, is to only support the
regex for placeholder variables:
>
> router.get("/items/:(var:[0-9]+)/something", ...);
>
> It would then be implemented as step after the basic route matching.
>
> —
> Reply to this email directly or view it on GitHub.

The syntax will more likely be Python's named group, which `std.regex`
supports (-> "/items/:(?P<var>[0-9]+)/something/").
That would indeed be very cool to have.

s-ludwig:
Closing this, since the implementation approach doesn't work anymore. I'll open a separate ticket instead.

PR #1063, opened 2015-04-07
Use explicit checks for arrays in boolean contexts

schuetzm:
Newest DMD is more picky about implicit conversion of arrays to bool, see this PR:
https://github.com/D-Programming-Language/dmd/pull/2885

### DIFF
--- a/source/vibe/data/bson.d
+++ b/source/vibe/data/bson.d
@@ -1428,7 +1428,7 @@ struct BsonSerializer {
 			m_type = tp;
 		}
 
-		if (m_entryName) {
+		if (m_entryName !is null) {
 			m_dst.put(tp);
 			m_dst.putCString(m_entryName);
 			m_entryName = null;
--- a/source/vibe/web/validation.d
+++ b/source/vibe/web/validation.d
@@ -76,7 +76,7 @@ struct ValidUsername {
 		ret.nullify();
 
 		auto err = appender!string(); // TODO: avoid allocations when possible
-		if (validateUserName(str)) ret = ValidUsername(str);
+		if (validateUserName(err, str)) ret = ValidUsername(str);
 		else *error = err.data;
 		return ret;
 	}
@@ -120,7 +120,7 @@ struct ValidPassword {
 		ret.nullify();
 
 		auto err = appender!string(); // TODO: avoid allocations when possible
-		if (validatePassword(str, str)) ret = ValidPassword(str);
+		if (validatePassword(err, str, str)) ret = ValidPassword(str);
 		else *error = err.data;
 		return ret;
 	}
--- a/source/vibe/core/args.d
+++ b/source/vibe/core/args.d
@@ -47,7 +47,7 @@ import core.runtime;
 bool readOption(T)(string names, T* pvalue, string help_text)
 {
 	// May happen due to http://d.puremagic.com/issues/show_bug.cgi?id=9881
-	if (!g_args) init();
+	if (g_args is null) init();
 
 	OptionInfo info;
 	info.names = names.split("|").sort!((a, b) => a.length < b.length)().array();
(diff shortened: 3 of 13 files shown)


source/vibe/data/bson.d
```diff
@@ -1428,7 +1428,7 @@ struct BsonSerializer {
 			m_type = tp;
 		}
 
-		if (m_entryName) {
+		if (m_entryName != "") {
```
s-ludwig: Still not perfect, but this should better be `!is null`, because dictionary keys can in theory be `""`, too.

s-ludwig:
Otherwise LGTM.

schuetzm:
Added a bunch of others that turned up when I removed a now unnecessary cast (PR coming). There are still a few left that I can't locate:

```
/home/marc/d/phobos/std/exception.d(351): Warning: implicit conversion of dynamic arrays to bool can be ambiguous and will be deprecated. Use one of: value !is null, value.length, or value.ptr instead
__diet_code__(11): Warning: implicit conversion of dynamic arrays to bool can be ambiguous and will be deprecated. Use one of: "foo" !is null, "foo".length, or "foo".ptr instead
__diet_code__(11): Warning: implicit conversion of dynamic arrays to bool can be ambiguous and will be deprecated. Use one of: cast(string)null !is null, cast(string)null.length, or cast(string)null.ptr instead
__diet_code__(11): Warning: implicit conversion of dynamic arrays to bool can be ambiguous and will be deprecated. Use one of: "foo" !is null, "foo".length, or "foo".ptr instead
__diet_code__(11): Warning: implicit conversion of dynamic arrays to bool can be ambiguous and will be deprecated. Use one of: cast(string)null !is null, cast(string)null.length, or cast(string)null.ptr instead
```

schuetzm:
Ok, found the remaining ones.

source/vibe/web/validation.d
```diff
@@ -76,7 +76,7 @@ struct ValidUsername {
 		ret.nullify();
 
 		auto err = appender!string(); // TODO: avoid allocations when possible
-		if (validateUserName(str)) ret = ValidUsername(str);
+		if (validateUserName(str) != "") ret = ValidUsername(str);
```
s-ludwig: Good that these lines appear now in a diff. They are using the wrong overload. Should be:
`if (validateUserName(err, str)) ret = ValidUsername(str);`

I wouldn't have expected the DMD change to catch these kinds of bugs, too :)

source/vibe/web/validation.d
```diff
@@ -120,7 +120,7 @@ struct ValidPassword {
 		ret.nullify();
 
 		auto err = appender!string(); // TODO: avoid allocations when possible
-		if (validatePassword(str, str)) ret = ValidPassword(str);
+		if (validatePassword(str, str) != "") ret = ValidPassword(str);
```
s-ludwig: Same here.

s-ludwig:
The `getInterfaceValidationError` comparisons should probably use `!is null` instead to be consistent with the rest of the module. Although in reality both should always be equivalent. @Geod24, any preference?

schuetzm:
Changed the validation functions to use the output range overloads.

I replaced the `!= ""` by `!is null`. For public functions, I would prefer the string comparison, but as `getInterfaceValidationError` is private, we're in control of all uses and can make sure that it actually never returns `""`.

s-ludwig:
Thanks! BTW, I agree in principle with avoiding `!is null`, it would be better to use `Nullable!string` in cases where a "null" is expected. I suspect that `!= ""` is less efficient than `.length` or `!str.empty`, though (has anyone actually measured that by chance?), so I'd usually rather use the latter (but no need to do that for the changes here).

schuetzm:
It still compiles to way too much code:
https://issues.dlang.org/show_bug.cgi?id=14436

Before yebblies fixed that, testing for `null` was just as bad.

But this is clearly something that should be fixed in the compiler.

PR #1064, opened 2015-04-07
Remove cast that no longer works with DMD master

schuetzm:
Compare this corresponding fix in Phobos:
https://github.com/D-Programming-Language/phobos/pull/3143


schuetzm:
(Crossing fingers that it will still work with older DMDs...)

PR #1066, opened 2015-04-10
fix usage of fqn name with selective imports

MartinNowak:


s-ludwig:
Nice to see progress on those issues!

PR #1067, opened 2015-04-13
vibe.web.i18n: Fix bug (Range violation) in wrapText

AndreyMZ:
The bug prevented extraction of strings with escape sequence, e.g. "\n".


s-ludwig:
Thanks! LGTM, just waiting for Travis to finish.

PR #1068, opened 2015-04-14
Added REPORT method for webDAV

gedaiu:
Some webDav extensions use the REPORT method to get information about
the existing resources


s-ludwig:
I'll also add all of those mentioned in https://www.ietf.org/rfc/rfc3253.txt when I get the chance.

gedaiu:
thanks!

PR #1069, opened 2015-04-14
Add version Have_vibe_d

etcimon:
I must have removed it during a previous pull


Geod24:
dub add these automatically.
Though it strips special chars (I guess it strips the dash then): https://github.com/D-Programming-Language/dub/search?utf8=%E2%9C%93&q=Have_

PR #1071, opened 2015-04-15
More compatibility fixes for changes in DMD & Phobos

schuetzm:
Adding a missing `package` protection that erroneously wasn't checked before (https://github.com/D-Programming-Language/dmd/pull/4558).

Also, https://github.com/D-Programming-Language/phobos/pull/3172 introduced a regression in `indexOf` that requires explicit slicing to work around.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -201,7 +201,7 @@ private Task runTask_internal(ref TaskFuncInfo tfi)
 
 	f.m_taskFunc = tfi;
 
-	atomicOp!"+="(f.m_taskCounter, 1);
+	f.bumpTaskCounter();
 	auto handle = f.task();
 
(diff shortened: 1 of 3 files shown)


s-ludwig:
Where is the place for which `package` is needed? I couldn't find any reference to it outside of task.d.

schuetzm:
```
source/vibe/core/core.d(204): Error: class vibe.core.task.TaskFiber member m_taskCounter is not accessible from module core
```

s-ludwig:
Oh okay, missed that. Then let's fix that by adding a `CoreTask.bumpTaskCounter()` method instead. That's cleaner and `package` class members also somehow seem like a strange thing.

schuetzm:
Done.

s-ludwig:
Thanks!

PR #1076, opened 2015-04-25
urlEncode() should only copy if necessary

schuetzm:


s-ludwig:
Could `allOf` make use of a sorted input array of characters and use a binary search (i.e. using `assumeSorted`)? Otherwise I'd use a `switch` range statement or an AA instead for efficiency reasons. Also for efficiency, using `str.representation.allOf` is preferred to avoid unnecessary UTF decoding.

schuetzm:
I changed it to a switch statement. Both DMD and LDC compile it into a lookup table.

s-ludwig:
Thanks, that should work.

PR #1077, opened 2015-04-25
Fix unittest failures caused by changed iteration order of AAs

schuetzm:


s-ludwig:
Thanks! Out of interest: Do you know which change in the AA implementation has changed the order?

schuetzm:
Haven't tested it, but most likely this one:
https://github.com/D-Programming-Language/druntime/pull/1229

PR #1078, opened 2015-04-26
Fix isOpApplyDg to accept non-ref args.

s-ludwig:
@Geod24 Just a quick verification, if there is anything subtle that would preclude non-ref args.

### DIFF
--- a/source/vibe/internal/meta/traits.d
+++ b/source/vibe/internal/meta/traits.d
@@ -360,13 +360,20 @@ template isOpApplyDg(DG, TKEY, TVALUE) {
 		private alias STC = ParameterStorageClass;
 		// Just a value
 		static if (PTT.length == 1) {
-			enum isOpApplyDg = (is(PTT[0] == TVALUE) && PSCT[0] == STC.ref_);
+			enum isOpApplyDg = (is(PTT[0] == TVALUE));
 		} else static if (PTT.length == 2) {
-			enum isOpApplyDg = (is(PTT[0] == TKEY) && PSCT[0] == STC.ref_)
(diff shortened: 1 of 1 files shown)


Geod24:
The specs (kinda) says they are ref http://dlang.org/statement.html#ForeachStatement , but from experience, non-ref is just fine. Actually, if you look at the specs, they are even more permissive with templated opApply.
Did that triggers any bug ?

source/vibe/internal/meta/traits.d
```diff
@@ -360,13 +360,20 @@ template isOpApplyDg(DG, TKEY, TVALUE) {
...
 		enum isOpApplyDg = false;
 	}
 }
+
+unittest {
+	static assert(isOpApplyDg!(int delegate(int, string), int, string));
+	static assert(isOpApplyDg!(int delegate(ref int, ref string), int, string));
+	static assert(isOpApplyDg!(int delegate(int, ref string), int, string));
+	static assert(isOpApplyDg!(int delegate(ref int, string), int, string));
+}
```
Geod24: Missing newline.

s-ludwig:
Yeah, it used to be the case that `ref` was required and the examples are probably just still from that time. I'll try to find out when this was changed.

The problem is that usually the compiler allows to use non-ref variables in the `foreach` statement, even if `opApply` only takes `ref` arguments. However, when a template type is used for the delegate, it tries to match a delegate without `ref`, so that the `isOpApplyDg` failed. I need to dig up the actual example I had, though, since the unit tests do seem to pass...

BTW, I just noticed now that the unit test changes in rest.d fail for 2.065. I'll fix that and rebase.

PR #1081, opened 2015-04-28
Added HTTP/2 and a new TLS engine - Botan

etcimon:
Looks like things are nearing a stable state for this development. I'd like to put it up here as a candidate for a pull, of course there will always be improvements to be made b/c adding features will always add potential zones of failure as well.

There's a lot of stuff in these changes, it needs some more testing but I think I've resolved most of the stuff. Right from the top of my mind I'll have to fix the libev, win32 drivers for the new waitLocal event method, test the new semaphore for a max connections setting in the ConnectionPool, try using the push promise on a server, and test for leaks using `valgrind --tool=massif`.

But I added the h2_request, h2_server examples to test this new server/client out. There's really a lot of stuff available here and more there if you consider that Botan exposes its API pretty much directly as well (we can setup PSK or SRP handshakes too without altering vibe/botan).

So, have fun testing it ;)


Geod24:
You got an extra commit, looks like you need to rebase on master.
In addition such a HUGE change (we're talking 8k LoCs) should be splitted into smaller chunk if you want a realistic review.

dub.json
```diff
@@ -1,73 +1,118 @@
...
+	"description": "Asynchronous I/O framework",
+	"license": "MIT",
+	"copyright": "Copyright © 2012-2014 rejectedsoftware e.K.",
+	"homepage": "http://vibed.org/",
+	"authors": [
+		"Sönke Ludwig",
+		"Jan Krüger",
+		"Matthias Dondorff",
+		"see github for more"
+	],
```
Geod24: Looks like indentation change.

dub.json
```diff
@@ -1,73 +1,118 @@
...
+		{
+			"name": "libasync_botan",
+			"dependencies": {
+				"botan": {"version": "~>1.11.12" },
+				"libasync": {"version": "~>0.7.3" }
+			},
+			"versions": ["Botan", "VibeUseNativeDriverType", "VibeLibasyncDriver"]
+		},
+		{
+			"name": "libasync_botan_win32",
```
Geod24: User shouldn't have to specify it's platform in the configuration.

examples/h2_request/source/app.d
```diff
@@ -0,0 +1,99 @@
...
+import std.stdio;
+import vibe.http.client;
+import vibe.stream.operations : readAllUTF8;
+import std.datetime;
+import core.thread;
+import vibe.stream.botan;
+void main()
+{
+	StopWatch sw;
+	sw.start();
```
Geod24: You can move the StopWatch directly in runTask, can't you ?

examples/h2_request/source/app.d
```diff
@@ -0,0 +1,99 @@
...
+								if (tls.x509Certificate) logDebug("%s", tls.x509Certificate.toString());
+							}
+						}
+					},
+					(scope res) {
+						logInfo("Response: %d", res.statusCode);
+						foreach (k, v; res.headers)
+							logInfo("Header: %s: %s", k, v);
+						result = res.bodyReader.readAllUTF8(true);
+						sw.stop();
```
Geod24: Shouldn't you read, stop the watch, then print the result ? Won't the `logInfo` skew the results ?

examples/h2_request/dub.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+	"name": "h2-request-example",
+	"description": "Performs a custom HTTP/2 request.",
+	"authors": [ "Sönke Ludwig", "Etienne Cimon" ],
+	"dependencies": {
+		"vibe-d": {"version": "~master", "path": "../../"}
```
Geod24: You can drop the `version` here

examples/h2_server/dub.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+	"name": "h2-server-example",
+	"description": "Uses the HTTPS and HTTP/2 protocol for serving requests.",
+	"authors": [ "Sönke Ludwig" ],
+	"dependencies": {
+		"vibe-d": {"version": "~master", "path": "../../"}
```
Geod24: Ditto for version

examples/h2_server/source/app.d
```diff
@@ -0,0 +1,27 @@
...
+import vibe.stream.tls;
+import vibe.stream.botan;
+import vibe.core.log;
+import vibe.core.core;
+import libasync.threads;
+import std.datetime;
+
+void handleRequest(scope HTTPServerRequest req, scope HTTPServerResponse res)
+{
+	res.writeBody(cast(ubyte[])"Hello, World!", "text/plain");
```
Geod24: `.representation`

source/vibe/http/server.d
```diff
@@ -730,44 +897,133 @@ final class HTTPServerRequest : HTTPRequest {
...
 	bool isHeadResponse() const { return m_isHeadResponse; }
 
-	/** Determines if the response is sent over an encrypted connection.
-	*/
-	bool ssl() const { return m_ssl; }
+	/// Determines if the response is sent over an encrypted connection.
+	bool tls() const { return (!m_isTest && m_conn.stack.tls) ? true : false; }
+
+	deprecated("Use tls")
+	bool ssl() const { return tls(); }
```
Geod24: Any deprecation should **really** be in a separate commit.

source/vibe/http/server.d
```diff
@@ -730,44 +897,133 @@ final class HTTPServerRequest : HTTPRequest {
...
+	}
+
+	private @property bool isHTTP2() { return (!m_isTest && m_conn.stack.http2 !is null)?true:false; }
+
 	@property SysTime timeFinalized() { return m_timeFinalized; }
 
-	/** Determines if the HTTP header has already been written.
-	*/
+	/// Determines if compression is used in this response.
+	@property bool hasCompression() { return (m_compressionStream.gzip !is null) ? true : false; }
```
Geod24: `return m_compressionStream.gzip !is null;` would be enough.

source/vibe/http/server.d
```diff
@@ -730,44 +897,133 @@ final class HTTPServerRequest : HTTPRequest {
...
+			}
+			else if (m_bodyStream.counting) {
+				FreeListObjectAlloc!CountingOutputStream.free(m_bodyStream.counting);
+				m_bodyStream.counting = null;
+			}
+		}
+
+	}
+
+	private @property bool isHTTP2() { return (!m_isTest && m_conn.stack.http2 !is null)?true:false; }
```
Geod24: SUperfluous ternary

source/vibe/http/server.d
```diff
@@ -730,44 +897,133 @@ final class HTTPServerRequest : HTTPRequest {
...
+			if (!hasCompression) {
+				if (isHeadResponse) return cast(OutputStream) m_bodyStream.none;
+				else if (m_isChunked) return cast(OutputStream) m_bodyStream.chunked;
+				else return cast(OutputStream) m_bodyStream.counting;
+			}
+			else if (m_isGzip) return cast(OutputStream) m_compressionStream.gzip;
+			else return cast(OutputStream) m_compressionStream.deflate;
+		}
+
+		@property ConnectionStream topStream() { return m_isTest ? cast(ConnectionStream) m_conn.test : (m_conn.stack.http2 ? cast(ConnectionStream) m_conn.stack.http2 : ( m_conn.stack.tls ? cast(ConnectionStream) m_conn.stack.tls : cast(ConnectionStream) m_conn.stack.tcp)); }
```
Geod24: That's some serious abuse of ternary operator here :D
Can you make this nested `if` please ?

Geod24:
You made the `setup-*.sh` non executable. Working on Windows ? ;)

examples/h2_server/source/app.d
```diff
@@ -0,0 +1,27 @@
...
+	auto settings = new HTTPServerSettings;
+	settings.port = 8080;
+	settings.bindAddresses = ["::1", "127.0.0.1"];
+	version(Botan) settings.tlsContext = new BotanTLSContext(TLSContextKind.server/*, createCreds()*/); // see stream/botan for more options
+	settings.tlsContext.useCertificateChainFile("server.crt");
+	settings.tlsContext.usePrivateKeyFile("server.key");
+
+	listenHTTP(settings, &handleRequest);
+
+}
```
Geod24: Newline ;)

source/vibe/core/core.d
```diff
@@ -1169,12 +1176,14 @@ private class VibeDriverCore : DriverCore {
...
-			assert(!s_eventLoopRunning, "Event processing outside of a fiber should only happen before the event loop is running!?");
-			m_eventException = null;
-			try if (auto err = getEventDriver().runEventLoopOnce()) {
-				logError("Error running event loop: %d", err);
-				assert(err != 1, "No events registered, exiting event loop.");
-				assert(false, "Error waiting for events.");
+			try {
+				assert(!s_eventLoopRunning, "Event processing outside of a fiber should only happen before the event loop is running!?");
+				m_eventException = null;
+				if (auto err = getEventDriver().runEventLoopOnce()) {
```
Geod24: Why the change ? `assert` and a member assignement should be `nothrow`, aren't they ?

source/vibe/core/drivers/libasync.d
```diff
@@ -105,8 +105,13 @@ final class LibasyncDriver : EventDriver {
 
 	}
 
-	static @property bool isControlThread() {
-		return Thread.getThis().isDaemon && Thread.getThis().name == "CmdProcessor" ;
+	static @property bool isControlThread() nothrow {
+		bool is_ctrl_thr;
+
+		try is_ctrl_thr = Thread.getThis().isDaemon && Thread.getThis().name == "CmdProcessor";
+		catch {}
```
Geod24: `scope (failure) assert(false);` instead of just ignoring the error.

etcimon:
> should be splitted into smaller chunk if you want a realistic review.

Honestly it takes a lot of efforts to split it into branches so I don't mind if you pull out the parts and merge, I'll rebase. I'm in a bit of a rush and the community sharing task part of my work is going to be lower priority for a couple months (1 day/week vs 5). Thanks for taking the time to comment it's always good to improve the quality of software

s-ludwig:
I propose to retarget the pull against a non-master branch (I'll create one now). I'll then pull after a basic review and we can continue to refine the implementation in the branch before merging into master.

s-ludwig:
Okay, there is now a "http2-botan" branch here as well.

examples/h2_server/source/app.d
schuetzm: Yes, strictly speaking the cast is even wrong, because the result is mutable.

schuetzm: Ah, I see, `writeBody()` expects mutable data. I opened a PR here: https://github.com/rejectedsoftware/vibe.d/pull/1072

etcimon: I actually have the `write` methods implemented as a `const(ubyte)[]` in botan, which makes a lot more sense. I haven't seen any implementations of the `write` method mutating the data, in fact they're usually copied to a buffer before being mutated, for this reason this cast doesn't produce any bugs.

examples/h2_server/source/app.d
etcimon: You mean the file should end with a new line?

dub.json
etcimon: Here the -m32mscoff option isn't available as a dub architecture choice and the dflag propagate in the dependencies. I'd need to have some kind of shell script or a new feature in dub. So, this is a temporary workaround (unless someone teaches me something new)

source/vibe/core/core.d
etcimon: My bad, I was probably looking for a segfault somewhere. When the program catches Throwable, it makes it impossible to find an InvalidMemoryOperation because writing to logs uses an appender and the GC, and even the call stack won't hold any information about where it failed. And b/c it can happen in worker tasks using std.concurrency, that part also catches Throwable to append the thread to the scheduler again and the GC stays locked ie. you have an impossible debugging experience. Ugh XD

etcimon:
> You made the setup-*.sh non executable. Working on Windows ? ;)

I'm pulling my hair out about this, whatever I do, the permissions are changed back when committing on Windows..

Geod24:
To ignore filemode changes:
git config core.fileMode false

Or to edit it yourself: git update-index --chmod=+x <file>

2015-04-29 16:23 GMT+02:00 Etienne Cimon <notifications@github.com>:

examples/h2_server/source/app.d
s-ludwig: Definitely right about the missing `const` (PR is merged). But there is also a `string` overload of `writeBody`, so that the cast still shouldn't be necessary.

PR #1083, opened 2015-04-29
strip() DigestAuth directives and make them case-insensitive.

aka-demik:
The current implementation of Digest authentication does not work in FireFox. He uses comma+space as a directives delimiter.
Plus added toLower () as in [RFC 2617] (https://tools.ietf.org/html/rfc2617) said:
> The realm directive (**case-insensitive**) is required for all authentication schemes that issue a challenge. The realm value (case-sensitive), in combination with the canonical...


s-ludwig:
Thanks! I'll also change the `if`-list to a single `switch` statement.

PR #1084, opened 2015-04-29
Make exception handling for REST servers optional

schuetzm:


source/vibe/web/rest.d
```diff
@@ -533,12 +533,21 @@ class RestInterfaceSettings {
 	*/
 	bool stripTrailingUnderscore = true;
 
+	/** By default, vibe.d will handle $(D HTTPStatusException)s and turn other
+	    exceptions into an Internal Server Error.
```
Geod24: I'd suggest to rewrite to 'this module will handle ...' (because when they pass through, it's still Vibe.d that handle them), and 'turn exception into JSON exception object with an error code of 500 and a statusMessage/statusDebugMessage [blabla]', or something of that vein, describing what is being sent on the wire.

Geod24:
Not 100% sure if there could be a better way to do it, but it the mean time, if it gets the job done for you, LGTM.

Out of interest, what's the main motivation behind the change ? Better error handling or avoid leaking server internals, or something else ?

schuetzm:
Yes, I had doubts about that delegate, but according to my tests it doesn't affect performance negatively. But thinking about it again, I could alternatively catch the exceptions unconditionally and rethrow them if desired. Not sure whether that's better or worse.

My motivation is to handle additional exceptions and turn them into the appropriate HTTP error code. For example, an "record not found" exception throw by my ORM should result in 404, and a failed UTF8 validation or an invalid record (e.g. some required fields are left empty) in 401 Bad Request. This change allows uniform exception handling for REST and other requests alike.

s-ludwig:
Other possible approaches may be to
 - register an exception callback (`void delegate(HTTPResponse res, Exception e) RestInterfaceSettings.exceptionHandler`)
 - define a `@exceptionHandler!func` UDA that could even be translated to the proper `catch` clause (with specific exception types), avoiding the need for explicit dynamic casts

Thougts? I find the usage of these possibly a bit more obvious than what is required with the `handleExceptions` feature.

schuetzm:
A UDA is not enough for my purposes, because it applies only to specific methods. I don't want to repeat it over and over again, it needs to apply to the entire interface, ideally including sub-interfaces. Also, one of my goals is to unify exception handling of REST interfaces and others, so that both can be used side-by-side.

But you're right that the current way is not the most obvious one. Any arguments against supporting both UDAs _and_ a global exception handler?

s-ludwig:
> A UDA is not enough for my purposes, because it applies only to specific methods. I don't want to repeat it over and over again, it needs to apply to the entire interface, ideally including sub-interfaces.

```
class MyAPIImpl : MyAPI {
    @exceptionHandler!func:

    void getSomething() {}
    void getSomethingElse() {}
}
```

> Also, one of my goals is to unify exception handling of REST interfaces and others, so that both can be used side-by-side.

Do you mean `registerWebInterface` with (which I think should support the same mechanism), or something else? It seems like other kinds of interfaces *might* require at least a different content type in the reply.

> Any arguments against supporting both UDAs and a global exception handler?

Mostly just that ideally there should ideally be only one way to achieve a certain goal, but this is not *that* important. I just want avoid feature creep as long as there is no good reason.

schuetzm:
Thanks, I always forget that you can use `:` with UDAs.

> Do you mean registerWebInterface with (which I think should support the same mechanism), or something else?

Actually all routes, not restricted to any specific type of interface. I can simply set `HTTPServerSettings.errorPageHandler`, it applies to all of them.

> It seems like other kinds of interfaces might require at least a different content type in the reply.

Generally yes, but for my app it's fine. I'm implementing a hybrid interface that is usable both with Ajax-Requests and with normal form posts/gets _to the same URLs_. That's arguably weird, but I like to have that flexibility. Fall-through route and `@before` handlers work really well for that.

schuetzm:
Took me some time; I now rewrote it to use UDAs, as suggested by Sönke. It turned out to be quite simple. You can even specify the type of exception to catch.

@s-ludwig and @Geod24 : Is this ok?

source/vibe/web/rest.d
```diff
@@ -744,6 +747,20 @@ private HTTPServerRequestDelegate jsonMethodHandler(T, alias Func)(T inst, RestI
 				auto ret = handler(&__traits(getMember, inst, Method), params);
 				res.writeJsonBody(ret);
 			}
+		}
+
+		try {
+			static if (error_attributes.length == 1) {
+				auto Attr = error_attributes[0];
```
Geod24: This isn't used

Geod24:
It could also use an example :)
@s-ludwig had some WIP for the REST generator to make it more user-extensible. Could it cover this use case ?

source/vibe/web/rest.d
schuetzm: Actually, it is, see next line. It's a workaround for a DMD bug: https://issues.dlang.org/show_bug.cgi?id=14889

schuetzm:
Closing this PR, as it did something completely different that what the current one does.

schuetzm:
@Geod24 Can you please point me to the REST generator work you're talking about?

Geod24:
@schuetzm : Sure. There's a P.R. that add a JS client and a significant refactoring of the code ( #1209 ), and you can have more insight here : https://github.com/rejectedsoftware/vibe.d/pull/1202#issuecomment-127150505

PR #1085, opened 2015-04-29
Added HTTP/2 and a new TLS engine - Botan

etcimon:
This is a targeted pull for http2-botan branch of the pull request at https://github.com/rejectedsoftware/vibe.d/pull/1081

I read the review and made a few corrections and will keep doing so as long as I see them coming. Since this is a pretty big project (> 115k LOC incl. libasync, botan, libhttp2, memtuils), some of those bugs may take roots deeper into dmd/dub/druntime/etc, for example 

- The use of 32mscoff configuration is necessary on win32 because -m32mscoff doesn't propagate and dub doesn't have a --arch=x86coff option. Also, the `dub fetch` packages don't seem to generate the dub files correctly so the repositories need to be cloned and the dependency "path"s manually adjusted.
- This branch is also not compatible with DMD < 2.067.1 on windows because of the amount of symbols, I implemented COFF Big Tables in DMD for this reason.
- This is not compatible with DMD < 2.067.0 because Botan uses the `asm pure` attributes
- The LDC/GDC compilers are thus not compatible because they haven't reached 2.067 yet and don't seem to be able to ignore the DMD-specific code
- It also requires DMD64 when building win64 because it uses too much ram to compile.
- The linker for posix is very slow (>13 seconds) unless you use the Gold linker (~2 seconds) because of the amount of symbols. 

I added appropriate error messages where possible. I think everything runs very stable right now nonetheless once you have the platform correctly setup


Geod24:
Thanks a lot for pushing this forward.
I'll try to review stuff here frequently, and we'll see if we can integrate it step by step into master.

etcimon:
> Thanks a lot for pushing this forward.
> I'll try to review stuff here frequently, and we'll see if we can integrate it step by step into master.

My pleasure (I needed it). vibe.d will be an absolute powerhouse in terms of web development with this. I still wish there were more native libraries (dns resolvers, distributed balancers, ssh, git, etc) but I think this will give us an edge over other libraries, whereas it could show off the low-level & high-level combined abilities of the platform to help convince some highly skilled contributors to hop on eventually.

s-ludwig:
Since this is too big for a meaningful review, I'll pull as it is now and then move things bit by bit into master until nothing is left (I'll try to let this be the main focus of my work next week).

etcimon:
Thanks! And I hope you'll like it. I'm going to be concentrating more on vibe.d because I need more features related to it in my developments now, e.g. I'll need to implement the `HTTPClient.writePart` method this week

s-ludwig:
A question regarding licensing: Usually all contributions have to be made under the rejectedsoftware copyright as noted in the contribution guidelines. But they definitely have to be MIT licensed and they definitely may not violate any external license terms. Since some of the code looks like it's ported from C++ and copyright/license notes are missing for new files (`cookiejar` and `http2`), this still needs some clarification.

etcimon:
> Since some of the code looks like it's ported from C++ and copyright/license notes are missing for new files (cookiejar and http2), this still needs some clarification.

The http2.d file was completely written from scratch and I'm licensing it MIT

The cookiejar.d file was completely written from scratch except for the date parsing, which was heavily inspired by: https://github.com/icefox/parse_cookie_date - The Nokia license doesn't apply here because nothing from Qt was used so I removed it. But the original author licensed it 3-BSD which is why I tried to rewrite it as differently as possible. If I infringed on anything I might as well start over, right?

s-ludwig:
Can you make a quick pull adding the module headers to the http2-botan branch? I'd say the date parsing code should better be rewritten (that was actually also what I was looking at when referring to ported code). Kind of stupid because the BSD license is so close to MIT, but this really needs to be handled correctly. I'll just keep it in mind for now though, no need for action before the code is ready to be pulled into master.

etcimon:
> I'll just keep it in mind for now though, no need for action before the code is ready to be pulled into master.

What about dependencies though? The libhttp2 library has MIT but with a clause from curl license. Can vibe.d rely on that kind of dependency or should it be removable through versioning?

s-ludwig:
Should be fine. Only GPL and similar licenses have additional restrictions on dependent code. It's just that their license terms will apply to any compiled software (e.g. if there is a clause that dictates showing the copyright string(s) in the final application).

etcimon:
So I might as well move the cookie date parser to libhttp2 ?

s-ludwig:
Yeah, that would just mean that you'll get a mixed license library. Since it's also used for HTTP/1.x, libhttp2 doesn't seem to be the most fitting place though. But let's wait with that until things are close to getting merged. The HTTP2 code could even be merged before the cookie code.

etcimon:
> The HTTP2 code could even be merged before the cookie code.

A nice ordering I think would be :
semaphore, libasync fixes, client refactoring & features, server refactoring & features, botan, waitLocal, http/2, cookie jar

The memutils library is mixed license though, you have the secure pool there that is 2-BSD under Botan license but it's really important for the flow of data. It maps to a memory region that doesn't ever touch peristent memory on any occasion.

s-ludwig:
I'm currently undoing as much of the refactorings and optimizations in the HTTP client/server as possible. Otherwise it won't be possible to merge the HTTP/2 additions at all (usually the whole PR would have had zero chances of being merged due to its size and due to mixing so many things, but in this case it would be a shame). `waitLocal` is removed, too, I think we need a solution that is implemented as an optimization under the hood instead. But otherwise I'd chose the same order.

I'll get back with a list of all things that I've removed/reverted when I'm ready, so that we can go through them one by one to see what makes sense to be put back and in which form.

etcimon:
A lot of the refactoring was necessary from my perspective. Are you sure its going to be possible to remove it?

s-ludwig:
Quoting some further fixes that need to be applied to the http2 branch:
> - The outputStream can be null in which case the freeing shouldn't happen and the streams are still flushed (currently this fails the assertion): if (!outputStream) skip freeing
> - The outputStream doesn't necessarily exist when writing headers (the current code may trigger a segfault): should be StreamOutputRange(topStream)
> - The libasync driver has an assertion that slows everything down a lot when many writes are involved: should be removed
> - The libasync AsyncFile was sluggish for large files which is fixed in 0.7.4 (I tagged it just now).

s-ludwig:
> A lot of the refactoring was necessary from my perspective. Are you sure its going to be possible to remove it?

Some of it seems to be removable without any issues. Other things may have a performance/memory impact. Maybe others are necessary in a more fundamental way, but I'll *have* to get everything to a point where the actual changes are small enough to keep a mental model of them. When everything is done so far, the refactorings can be discussed one-by-one and possibly added back (or maybe a suitable alternative).

But using it as-is would basically mean that the previously tested implementation gets swapped by an almost completely new implementation that could introduce any number of bugs that are more or less impossible to detect by pure inspection. Since there is no serious HTTP protocol test coverage, this is really not an option at this stage.

etcimon:
> Since there is no serious HTTP protocol test coverage, this is really not an option at this stage.

I'd really like to have some more high-level tests. ie. phantomjs with a vibe.d proxy - so that some dynamic websites with real json events can be tested through the vibe.d server using a browser and http/2 multiplexing enabled/disabled in the proxy.

PR #1086, opened 2015-05-01
Added keys method for Json objects

gedaiu:


s-ludwig:
We unfortunately can't add any methods to `Json` due to `opDispatch`, as that would be a breaking change (and a silently breaking change in this case: `foreach (k; json.keys) writefln("Key: %s", k);` could work before and after the change, but with different outcomes). `opDispatch` seemed to be a nice idea on the surface, but it turned out to be definitely a design mistake.

I'm not sure if it actually does, but it should automatically work in [stdx.data.json](https://github.com/s-ludwig/std_data_json) though, through the use of `Algebraic`.

gedaiu:
I did not know that... I hoped that I can easily add the `keys` method :P

PR #1087, opened 2015-05-01
Switch statement without a default is deprecated

schuetzm:


s-ludwig:
Ugh, sorry. I just pushed a fix before noticing the pull request here. Really need to get that libasync driver issue sorted out, so that the build goes green again.

PR #1090, opened 2015-05-03
Update dub.json

etcimon:
Try to go green

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -39,7 +39,7 @@
 		},
 		{
 			"name": "libasync",
-			"dependencies": {"libasync": "0.7.1"},
+			"dependencies": {"libasync": "0.7.3"},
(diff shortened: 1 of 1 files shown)


s-ludwig:
Starting to look good. Glad I didn't put more work into debugging this on 0.7.1. I really should get libasync better integrated into my development workflow.

etcimon:
Yes, you should also know that I've fixed a bunch of driver issues too on the http2-botan branch.

etcimon:
Also, I refactored the client/server a lot, and found some case sensitive header comparisons (e.g. for content-encoding, == "gzip")

s-ludwig:
Okay, good to know. My plan is to move things bit by bit into master anyway, so these are good candidates.

PR #1091, opened 2015-05-04
[REST] Add ability to write to the headers

Geod24:
This patch solve one common issue: the inability to write to write to the headers from the server code.
I thought the best way to do it was to use the existing semantic, namely ref and out parameters.
This P.R. also includes some cleanups.

### DIFF
--- a/examples/rest/source/app.d
+++ b/examples/rest/source/app.d
@@ -340,6 +340,16 @@ unittest
  * - @headerParam : Get a parameter from the query header;
  * - @queryParam : Get a parameter from the query URL;
  * - @bodyParam : Get a parameter from the body;
+ *
+ * In addition, @headerParam have a special handling of 'out' and
+ * 'ref' parameters:
+ * - 'out' are neither send by the client nor read by the server, but
+ *	their value (except for null string) is returned by the server.
+ * - 'ref' are send by the client, read by the server, returned by
+ *	the server, and read by the client.
+ * This is to be consistent with the way D 'out' and 'ref' works.
+ * However, it makes no sense to have 'ref' or 'out' parameters on
+ * body or query parameter, so those are treated as error at compile time.
  */
 @rootPathFromName
 interface Example6API
@@ -347,7 +357,12 @@ interface Example6API
 	// The first parameter of @headerParam is the identifier (must match one of the parameter name).
 	// The second is the name of the field in the header, such as "Accept", "Content-Type", "User-Agent"...
 	@headerParam("auth", "Authorization")
-	string getResponse(string auth);
+	@headerParam("tester", "X-Custom-Tester")
+	@headerParam("www", "WWW-Authenticate")
+	string getPortal(string auth,
+					 ref string tester,
+					 out Nullable!string www);
+
 	// As with @headerParam, the first parameter of @queryParam is the identifier.
 	// The second being the field name, e.g for a query such as: 'GET /root/node?foo=bar', "foo" will be the second parameter.
 	@queryParam("fortyTwo", "qparam")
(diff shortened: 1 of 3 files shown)


Geod24:
Ping @Dicebot @s-ludwig

s-ludwig:
(Still need to review the last two commits for client and server side ref/out implementation.)

Just a general note, because it is frequently the reason why it takes ages for me to fully review a PR: It would really be good to limit pull requests to small chunks and especially not to mix refactorings with functional changes (even if they are properly split into distinct commits).

s-ludwig:
One thing missing is a set of (unit)tests that test for actual data correctness (reading/writing of the actual HTTP headers). I had started an attempt to refactor the code to be more typesafe, more `@nogc` and less string mixin ([web_rest_revamp](https://github.com/rejectedsoftware/vibe.d/commits/web_rest_revamp)). And although I'll have to redo everything from scratch now anyway (too much has changed over time), we should have a decently solid test suite before making such big changes. So at least for new features, I'd like to make a decent set of tests a requirement

s-ludwig:
Otherwise, LGTM!

Geod24:
> Just a general note, because it is frequently the reason why it takes ages for me to fully review a PR: It would really be good to limit pull requests to small chunks and especially not to mix refactorings with functional changes (even if they are properly split into distinct commits).

Sorry, got a little carried away on this one. I restrain myself for doing P.R. for just refactorings, since they are harder to justify. But I'll try to next time.

> One thing missing is a set of (unit)tests that test for actual data correctness (reading/writing of the actual HTTP headers).

Yeah, the example does that - so far -, but I'd really like to move it to an unittest. I also have some somewhere, but I need to update & P.R. them.

Geod24:
Thanks for reviewing !

Geod24:
Rebased on master (phew).
It can't be merged now because:
- I need to add the `Nullable` handling, but I'd like to consider it module-wise first, so I'll dive back into #1127 (also, comment of interest: https://github.com/rejectedsoftware/vibe.d/pull/1127/files#r32380883 ).
- There is a bug in `funcattr`: #1138 .

Geod24:
Oh, it doesn't fail because we're not running the examples... But it should.

Geod24:
Almost 2 months later... This is ready to go !

I worked around the `@before` bug by not creating an handler when we have no `@before` / `@after`.
This might seem a bit too high of a requirement, but I'd really like to have this in the next release so I can start building on it, and you know my position on `@before`. In addition, I really think this will lower the need for `@before` even more, and the next documentations update will clearly reflect this limitation.

@s-ludwig : Sorry for the additional P.R. to review ;-) I take the release won't happen before 2.068.0, will it ? I doubt there will be any adjustement needed this time, but just to be sure.

s-ludwig:
I'll put this next on my list to try and get it into 0.7.24 (will be released in lock-step with DMD 2.068).

Geod24:
There's a slight bug right now (spotted yesterday), but it only shows up when the client is used independently of the server.
So the client side will change a little. Basically instead of doing a check in `scope (success)` and trying to `throw`, I'll either move the `throw` in the handler or use a `try` / `catch`.

Geod24:
Good to go (and I'm definitely rewritting the client to something less stringy once 0.7.24 is behind us).

s-ludwig:
LGTM, except that several Travis runs seem to hang while compiling the REST example. Any idea?

s-ludwig:
Seems like #1161 didn't work (the libev tests also build the examples now).

Geod24:
Yikes, that's weird, I'm sure I tested it and it was working as expected.
I'll submit a patch ASAP, but currently I can't test anything since Travis report it cannot access " http://code.dlang.org/packages/libevent/2.0.1+2.0.16.zip ". Anything wrong with the dub-repository ?

s-ludwig:
Hm must be some regression on vibe.d master. I've recompiled the registry to get the SDL documentation online - it must have happened there. Looking into it...

s-ludwig:
Fixed. See https://github.com/rejectedsoftware/vibe.d/pull/1148#discussion_r33888248

Geod24:
@s-ludwig 
> LGTM, except that several Travis runs seem to hang while compiling the REST example. Any idea?

My guess is that 2.065 can't handle something in the string mixin of the client and it locks itself up (or consume so much memory it starts swapping like crazy).
I've no interest in investigating this, given we're going to drop 2.065 . It can just wait for another release, given there are already a couple of REST changes / bugfixes inside. However the first 2 commits (well, 3 now, I've rebased on top of #1166 in order to give it another shot with Travis) can go in IMO (they just forbid something that doesn't make sense), which is why I opened #1165

s-ludwig:
Okay, merged #1165. We can drop the 2.065 compilers right after release and merge the rest then.

Geod24:
Rebased on master, and green at last.

s-ludwig:
:+1:

PR #1093, opened 2015-05-04
Make it possible to log unsafe objects

schuetzm:
Specifically, `Variant` erroneously had a `@trusted` destructor, which has been corrected here:
https://github.com/D-Programming-Language/phobos/pull/3227

(The affected functions are templates, and get their `@safe`-ty inferred anyway.)

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -109,7 +109,7 @@ nothrow {
 		args = Any input values needed for formatting
 */
 void log(LogLevel level, /*string mod = __MODULE__, string func = __FUNCTION__,*/ string file = __FILE__, int line = __LINE__, S, T...)(S fmt, lazy T args)
-	nothrow @safe if (isSomeString!S)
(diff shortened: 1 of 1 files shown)


s-ludwig:
I actually consider that `Variant` change a regression, at least in case of an `Algebraic` where all types are `@safe`, that really needs to be fixed properly instead of just breaking valid code (std.data.json). But `Algebraic` is a different story of course.

For this PR, it would be good to include a unit test that makes sure to still stay `nothrow @safe` for `nothrow @safe` parameters. Also, the `nothrow` annotation can stay in the interface, as it's really intended to always be `nothrow`, even for possibly throwing parameter types.

schuetzm:
Re-added `nothrow`. There was already a unittest testing `@safe`, I just added `nothrow` to it.

s-ludwig:
Ah ok thanks, didn't see/remember that. Merging.

PR #1101, opened 2015-05-12
Use Content-Length Header in Form Uploads

etcimon:
This optimization takes advantage of Content-Length if provided and avoids many calls to read/write, which gives me a 20% performance improvement.

### DIFF
--- a/source/vibe/inet/webform.d
+++ b/source/vibe/inet/webform.d
@@ -190,8 +190,7 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 			long to_read;
 			do {
 				to_read = min(len, buf.length);
-				stream.read(buf[0 .. to_read]);
-				file.write(buf[0 .. to_read]);
+				file.write(stream, to_read);
 				len -= to_read;
 			} while (to_read == buf.length && len > 0);
 		}
--- a/source/vibe/inet/webform.d
+++ b/source/vibe/inet/webform.d
@@ -183,7 +183,19 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 
 		auto file = createTempFile();
 		fp.tempPath = file.path;
-		stream.readUntil(file, cast(ubyte[])boundary);
+		if (auto plen = "Content-Length" in headers) {
+			import std.conv : to;
+			long len = (*plen).to!long;
+			ubyte[1024] buf = void;
+			long to_read;
+			do {
+				to_read = min(len, buf.length);
+				stream.read(buf[0 .. to_read]);
+				file.write(buf[0 .. to_read]);
+				len -= to_read;
+			} while (to_read == buf.length && len > 0);
+		}
+		else stream.readUntil(file, cast(ubyte[])boundary);
 		logDebug("file: %s", fp.tempPath.toString());
 		file.close();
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Instead of a manual buffered read loop, it should use the `OutputStream.write` overload that takes an `InputStream` and a length. At least in theory this opens up some optimization opportunities and reduces code size.

etcimon:
Ah, yes of course. I'm a little unsure about how to deal with any newline there could be before the next boundary.

etcimon:
Ok the boundary includes a newline. One less thing to worry about

etcimon:
I'm currently writing an ecryption-enabled cookie jar in sqlite, which for the moment simply decrypts/decompresses/extracts a blob into a file, calls the file cookie jar, and puts it back in the db. I think a lot of session stuff could also be added this way which is ideal for user applications that deal with a web UI. Would you consider embedding the sqlite library with vibe.d or should this belong to another repository?

s-ludwig:
I'd opt for a separate repository. I'll eventually drag out the MongoDB/Redis stuff, too, so it doesn't make much sense to include more database code now. Another thing is that sqlite doesn't really fit well with vibe.d, because it doesn't use AIO - at least not well enough to advertise it as part of the core distribution.

s-ludwig:
Regarding the read loop, I meant simply `if (auto plen = ...) file.write(stream, (*plen).to!long);`

etcimon:
> I'd opt for a separate repository. I'll eventually drag out the MongoDB/Redis stuff, too, so it doesn't make much sense to include more database code now. Another thing is that sqlite doesn't really fit well with vibe.d, because it doesn't use AIO - at least not well enough to advertise it as part of the core distribution

Sure, do you think it could appear as a use of "dub features" though? For example the way Botan was built, I kind of hacked my way through but the principle is to opt-in/opt-out of certain features and get a nice mix. The compile time and exec size seems to adjust very well to it too and it doesn't require building a separate repo which also requires finding ways of seeing if the right "mix" is being used together.

s-ludwig:
I think we'll definitely need to have a "features" feature, but in this case a separate package seems like the logical choice. You can make dependencies optional, which often serves the same purpose as a "feature". I've already changed the code in the http2-botan branch to use `version (Have_botan)` to conditionally include the Botan based TLS implementation, so that it will hopefully work out without increasing the number of configurations.

s-ludwig:
Oh, I wouldn't want to add it to vibe.d as a "feature" or optional dependency, though. It's simply a small library layered on top. At least that's how it should be in an ideal world.

etcimon:
Btw, I've also found some bugs when testing some more on my end (with a file server) 

- The outputStream can be null in which case the freeing doesn't happen but the streams are flushed: https://github.com/rejectedsoftware/vibe.d/blob/http2-botan-cleanup/source/vibe/http/server.d#L1270 => if (!outputStream) skip freeing
- The outputStream doesn't necessarily exist when writing headers: https://github.com/rejectedsoftware/vibe.d/blob/http2-botan-cleanup/source/vibe/http/server.d#L1330 => should be StreamOutputRange(topStream)
- The libasync driver has an assertion that slows everything down *a lot* when many writes are involved: https://github.com/rejectedsoftware/vibe.d/blob/http2-botan-cleanup/source/vibe/core/drivers/libasync.d#L549 => should be removed
- The libasync AsyncFile was sluggish for large files which is fixed in 0.7.4 (I tagged it just now).

I think these quick fixes belong in the http2-botan-cleanup so I'm pointing them out here

s-ludwig:
Okay, I'll paste these points to the http2 pull request, so that they won't get lost.

PR #1102, opened 2015-05-13
Header write being skipped on clients causes a connection freeze

etcimon:
I'm currently writing more high-level tests (reverse proxy) and finding a lot of new issues with HTTP/2 and the client. Seems like I'll be adding more soon enough, it seems like a good way to validate the code.

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -989,6 +989,8 @@ final class HTTPClientRequest : HTTPRequest {
 
 		// http/2
 		if (isHTTP2) {
+			if (auto pka = "Connection" in headers)
+				headers.remove("Connection");
 			logDebug("Writing HTTP/2 headers");
 			http2Stream.writeHeader(requestURL, tlsStream ? "https" : "http", method, headers, m_cookieJar, m_concatCookies);
 			return;
--- a/source/vibe/http/http2.d
+++ b/source/vibe/http/http2.d
@@ -1865,16 +1865,18 @@ private:
 							stream.m_rx.ex = new Exception("Push Promise being sent by client stream... and with no headers.");
 						// header request/response/push-response
 						} else {
+							bool is_push;
 							// Clients: Is possible that we received this request as a push response?
 							if (!isServer) {
 								// To verify, we "pop" the push response that corresponds to this stream's headers, or null if none
 								if (HTTP2Stream push_stream = popPushResponse(stream)) {
+									is_push = true;
 									assert(!isServer);
 									stream.readPushResponse(push_stream);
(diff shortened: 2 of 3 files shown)


etcimon:
The reverse proxy test I'm doing has this setup:

Firefox <-- HTTP/1.1 --> vibe.http.proxy <-- HTTP/2 --> vibe.http.fileserver

I tried to test to a full extend most HTTP protocol features this way. I managed to make it stable for secure http/2. There was even a [severe bug in nghttp2](https://github.com/tatsuhiro-t/nghttp2/pull/232) that made it impossible to transfer large files.

Here is a commit log just to give an idea (most of it is writeln debug info):
https://github.com/etcimon/vibe.d/commit/71752440bab77e2146fe62d44686586d9f7a0130

Most of the changes are in http2.d btw. I'm going to be testing the h2c (cleartext http/2 upgrade) and the push response over the next day and start merging. Which of your branches would you rather have it based on? Is there some way I can help merge it in the cleanup branch?

s-ludwig:
Ideal would be to keep all new changes based on the cleanup branch. I currently slowly work through the http2 commit and integrate parts into master (or revert small cosmetic bits to make the diff smaller) and constantly rebase onto master during that time. Pulling from the original http2-botan branch should still be possible without big issues, but they do start to diverge.

Regarding the test setup, that indeed seems to be very well suited for testing, I remember how a lot of bugs were caught that way in the early days when the first vibe.d version got released (which were not caught by testing the server or client components in isolation).

etcimon:
> Regarding the test setup, that indeed seems to be very well suited for testing, I remember how a lot of bugs were caught that way in the early days when the first vibe.d version got released (which were not caught by testing the server or client components in isolation).

Yes, I'm preparing to write it for some test server in the `test/` subdirectory with this in travis syntax: http://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-PhantomJS

Still need to figure out what test pages would be most suitable

etcimon:
I was wondering about a feature. It could be useful to have a HTMLPushStream in the fileserver, which reads the `src=""` attributes and pushes them if the file isn't a 304. What do you think?

s-ludwig:
Hm, just going through the open PRs, this isn't in the http2-botan-cleanup branch, yet, is it? Can you retarget the PR onto it?

etcimon:
This can be closed, I marged many changes except one that I made a pull request for.

I haven't merged my changes to the reverse proxy and file server so there's a few additions there too. I also fixed clearSession in the cookie jar recently.. I just need some time to finish a project and I can merge everything then

PR #1105, opened 2015-05-14
REST: Provide a better error message when user has no / multiple interfaces

Geod24:
http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/25414/


Geod24:
Green

PR #1106, opened 2015-05-14
Deprecate HTTPRouter

Geod24:
The removal notice has been around for more than a year (24 april 2014). It's time to stick a deprecate attribute on this. Based on #1104 to be green.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -31,6 +31,7 @@ else version = VibeRouterTreeMatch;
 	string format) must be considered part of the interface. However, this removes the
 	prime argument for having an interface in the first place.
 */
+deprecated("Will be removed in 0.7.25. See removal notice for more information.")
 interface HTTPRouter : HTTPServerRequestHandler {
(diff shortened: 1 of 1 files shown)


s-ludwig:
I thought about how to deprecate the interface (or rather deprecate using the interface as a base interface to `URLRouter`) without breaking code that relies on it (using `alias this` tricks and similar means). But I didn't find any that worked in the end (without issuing a deprecation message when building vibe.d itself). So unless we do find a way to make this happen, we can instead as well remove the interface altogether.

Geod24:
I would have went for a straight removal, because I doubt anyone use the interface itself.
Some people might use it in their signature, so maybe we could just provide an alias to avoid breaking signature, and just breaks everything that relied on it being an interface. How does the updated code sound ?

s-ludwig:
There could indeed be a few people deriving from `HTTPRouter`, because it was added by request back then. But I guess it should be few enough that the alias solution is a good alternative.

PR #1107, opened 2015-05-15
Add optional reason message and code to WebSocket.close

Yoplitein:
Not sure why the payload length is limited to 125 bytes. Regular messages can be longer, so it doesn't seem to be an issue with vibe code. I found no mention of such a limit for close frames in the RFC.

### DIFF
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -298,12 +298,12 @@ final class WebSocket {
 		Actively closes the connection.
 
 		Params:
-			reason = Message describing why the connection was terminated.
 			code = Numeric code indicating a termination reason.
+			reason = Message describing why the connection was terminated.
 	*/
-	void close(string reason = "", short code = 0)
+	void close(short code = 0, string reason = "")
 	{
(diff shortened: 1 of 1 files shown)


s-ludwig:
According to the RFC, the status code should match the codes in section 7.4. For that reason and to avoid a dynamic memory allocation in the default case, I'd wrap the new code in an `if (code != 0)`. I'd also change the order of parameters, because the code must always be set properly if a message is given, but not vice-versa. Otherwise looks good.

s-ludwig:
The 125 byte limit is mentioned in [Section 5.5](https://tools.ietf.org/html/rfc6455#section-5.5):
> All control frames MUST have a payload length of 125 bytes or less
> and MUST NOT be fragmented.

Yoplitein:
> The 125 byte limit is mentioned in Section 5.5:

Ah. I guess I didn't look well enough.

> I'd wrap the new code in an if (code != 0). I'd also change the order of parameters

Done and done. Also rebased, so the builds should pass now.

s-ludwig:
Thanks!

PR #1109, opened 2015-05-15
JsonSerializer.writeValue should accept immutable/const values

japplegame:


s-ludwig:
Since `T` is a template type, it should usually be inferred to `const`/`immutable` just fine. If you have any failing case, can you add that as a unit test?

japplegame:
serialization fails if immutable/const class has `toJson/fromJson` methods.
Caused by this line: https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/data/serialization.d#L308
Perhaps it will be better to change serialization.d instead json.d:
`serializer.writeValue!TU(value)` -> `serializer.writeValue!T(value)`.

s-ludwig:
Okay, I've taken another look and I think this is the safest fix. I'll merge and add a unit test.

PR #1110, opened 2015-05-18
Fix zlib output stream

etcimon:
The finalize instruction will ensure that `ChunkedOutputStream` prints `0\r\n\r\n`


etcimon:
You will need to do the same in `CountingOutputStream` though

etcimon:
I think this would only fix the HTTP Client on your end because it calls finalize once on the m_bodyWriter, whereas the server calls it on each stream..

s-ludwig:
Correct, the fix needs to be in the HTTP client. `finalize` is defined to not finalize underlying streams. If it would do that, it would be impossible to run multiple finalized streams in sequence over the same underlying stream (like it happens for keep-alive connections).

PR #1112, opened 2015-05-19
Add http/2 over cleartext, fix buffers

etcimon:
I tested on my browser using the following configuration:

`firefox <-- HTTP/1.1 --> vibe.d reverse proxy <-- HTTP/2 --> vibe.d file server`

The connection between the reverse proxy and the file server was both secure, or insecured (ALPN & HTTP/1.1 Upgrade), and gzip / no-gzip.

These tests allowed me to find many bugs and glitches in the HTTP/2 driver, and some in other places as well.

I tried to keep your changes as much as possible, and I tried to mirror mine, but sorry in advance if there is something I overlooked.

I didn't try to compile it, but should it already be working on this branch?

ps. I didn't test the push requests yet, I'm pushing it off for later. I got the most important parts working for now


etcimon:
Will start over with the head of branch

PR #1116, opened 2015-05-21
Zlib memory leak and leastsize assertion

etcimon:
This fixes an issue where zlib is not guaranteed to call deflateEnd or inflateEnd.

It also fixes an assertion failure that occurs when the underlying stream is closed prematurely.

### DIFF
--- a/source/vibe/stream/zlib.d
+++ b/source/vibe/stream/zlib.d
@@ -63,6 +63,11 @@ class ZlibOutputStream : OutputStream {
 		zlibEnforce(deflateInit2(&m_zstream, level, Z_DEFLATED, 15 + (type == HeaderFormat.gzip ? 16 : 0), 8, Z_DEFAULT_STRATEGY));
 	}
 
+	~this() {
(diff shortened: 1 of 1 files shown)


etcimon:
Bump

s-ludwig:
Sorry for having forgot about this. I meant to comment on this for a while.

s-ludwig:
I still need to understand the issue with that assertion. This should really have a unit test, but I think that it would be a little safer to, instead of removing the assertion, to set the `m_finished` flag when a premature end of input is detected.

etcimon:
The issue is when the connection closes, `m_in.leastSize` sees a premature end of input, and the zlib stream fails on this assertion, so it stands out from other streams where you can end up with a closed connection and fail an enforcement at a higher level rather than having the task end unexpectedly as a whole.

s-ludwig:
I've fixed this locally by adding a check for empty input in `readChunk`. I'll pull and then commit my changes on top.

PR #1118, opened 2015-05-21
support ulong values for Json

IgorStepanov:
Hello, Vibe.D!

I've written about my problem on forum: http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/25482/

I use vibe.d on small commercial product, and two days ago I got this trouble.
One REST server (api.shodan.io) sent me a JSON like `{..., serial: 17559991181826658461, ...}`
I read the official JSON standard: http://json.org/
Unfortunately, it doesn't define bounds for a number value, thus `99999999999999999999999999` is a correct JSON value.
However the most real systems uses 64-bit (as maximum) signed and unsigned integers for describe integral values.
Thus, supporting of ulong is very needed for users.

I've added a new `Type` value: `uint_`, which represents `ulong` values.
I prefer `int_` type instead of `uint_` when it possible.
If the integral value x <= long.max, I write is to Json as `int_` even if it hasn't the sign.

If this way is unacceptable, please suggest another way to parse ulong values.
We may convert it to long or to string.
Get an exception during parsing of correct JSON answer of a REST server is very, because I can't get the REST answer as string (using `vibe.web.rest`) and parse it manually.
Thanks.

### DIFF
--- a/examples/json/source/app.d
+++ b/examples/json/source/app.d
@@ -1,6 +1,7 @@
 ﻿import vibe.data.json;
 
 import std.stdio;
+import std.bigint;
 
 void main()
 {
@@ -17,4 +18,8 @@ void main()
 	Json parent = obj;
 	parent.remove("item1");
 	foreach (i; obj) writeln(i);
+
+	auto obj2 = parseJsonString(`{"serial":17559991181826658461}`);
+	writeln("serial: ", obj2["serial"]);
+	assert(obj2["serial"] == BigInt(17559991181826658461UL));
 }
--- a/source/vibe/data/json.d
+++ b/source/vibe/data/json.d
@@ -44,7 +44,7 @@ unittest {
 public import vibe.data.serialization;
 
 public import std.json : JSONException;
-import std.algorithm : equal, min;
+import std.algorithm;
 import std.array;
 import std.conv;
 import std.datetime;
@@ -53,7 +53,7 @@ import std.format;
 import std.string;
 import std.range;
 import std.traits;
-
+import std.bigint;
 
 /******************************************************************************/
 /* public types                                                               */
@@ -69,14 +69,41 @@ import std.traits;
 	a JSONException. Additionally, an explicit cast or using get!() or to!() is
 	required to convert a JSON value to the corresponding static D type.
 */
+
 struct Json {
 	private {
 		// putting all fields in a union results in many false pointers leading to
 		// memory leaks and, worse, std.algorithm.swap triggering an assertion
 		// because of internal pointers. This crude workaround seems to fix
 		// the issues.
-		void*[2] m_data;
+		void*[max((BigInt.sizeof+(void*).sizeof-1)/(void*).sizeof, 2)] m_data;
 		ref inout(T) getDataAs(T)() inout { static assert(T.sizeof <= m_data.sizeof); return *cast(inout(T)*)m_data.ptr; }
+
+		/**
+         * BigInt is a struct, and it have a special BigInt.init value,  differs the null.
+		 * m_data has no special initializer and when we tries to first access to BigInt
+		 * via m_bigint(), we should explicitly initialize m_data with BigInt.init
+         */
+		void initBigInt()
+		{
+			BigInt init_;
+			(cast(ubyte*)m_data.ptr)[0 .. BigInt.sizeof] = (cast(ubyte*)&init_)[0 .. BigInt.sizeof];
+		}
+
+		void finiBigInt()
+		{
+			BigInt init_;
+			/**
+             * After swaping, init_ contains the real number from Json, and it
+			 * will be destroyed when this function is finished.
+			 * m_bigint now contains static BigInt.init value and destruction may
+			 * be ommited for it.
+             */
+			swap(init_, m_bigint);
+		}
+
+
+		@property ref inout(BigInt) m_bigint() inout { return getDataAs!BigInt(); }
 		@property ref inout(long) m_int() inout { return getDataAs!long(); }
 		@property ref inout(double) m_float() inout { return getDataAs!double(); }
 		@property ref inout(bool) m_bool() inout { return getDataAs!bool(); }
@@ -90,8 +117,14 @@ struct Json {
 			uint m_magic = 0x1337f00d; // works around Appender bug (DMD BUG 10690/10859/11357)
 			string m_name;
 		}
-	}
 
+		long bigIntToLong() inout
+		{
+			assert(m_type == Type.bigint);
+			enforceJson(m_bigint >= long.min && m_bigint <= long.max, "Unable to convert BigInt("~format("%d", m_bigint)~") to long.");
+			return m_bigint.toLong();
+		}
+	}
 	/** Represents the run time type of a JSON value.
 	*/
 	enum Type {
@@ -99,6 +132,7 @@ struct Json {
 		null_,      /// Null value
 		bool_,      /// Boolean value
 		int_,       /// 64-bit integer value
+		bigint,     /// BigInt values
 		float_,     /// 64-bit floating point value
 		string,     /// UTF-8 string
 		array,      /// Array of JSON values
@@ -113,7 +147,6 @@ struct Json {
 		Array = array,          /// Compatibility alias - will be deprecated soon
 		Object = object         /// Compatibility alias - will be deprecated soon
 	}
-
 	/// New JSON value of Type.Undefined
 	static @property Json undefined() { return Json(); }
 
@@ -146,6 +179,8 @@ struct Json {
 	/// ditto
 	this(long v) { m_type = Type.int_; m_int = v; }
 	/// ditto
+	this(BigInt v) { m_type = Type.bigint; initBigInt(); m_bigint = v; }
+	/// ditto
 	this(double v) { m_type = Type.float_; m_float = v; }
 	/// ditto
 	this(string v) { m_type = Type.string; m_string = v; }
@@ -159,12 +194,20 @@ struct Json {
 	*/
 	ref Json opAssign(Json v)
 	{
+		auto old_type = m_type;
 		m_type = v.m_type;
+		if (old_type == Type.bigint && m_type != Type.bigint)
+			finiBigInt();
 		final switch(m_type){
 			case Type.undefined: m_string = null; break;
 			case Type.null_: m_string = null; break;
 			case Type.bool_: m_bool = v.m_bool; break;
 			case Type.int_: m_int = v.m_int; break;
+			case Type.bigint:
+				if (old_type != Type.bigint)
+					initBigInt();
+				m_bigint = v.m_bigint;
+				break;
 			case Type.float_: m_float = v.m_float; break;
 			case Type.string: m_string = v.m_string; break;
 			case Type.array: opAssign(v.m_array); break;
@@ -173,20 +216,72 @@ struct Json {
 		return this;
 	}
 	/// ditto
-	void opAssign(typeof(null)) { m_type = Type.null_; m_string = null; }
+	void opAssign(typeof(null))
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.null_;
+		m_string = null;
+	}
 	/// ditto
-	bool opAssign(bool v) { m_type = Type.bool_; m_bool = v; return v; }
+	bool opAssign(bool v)
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.bool_;
+		m_bool = v;
+		return v;
+	}
 	/// ditto
-	int opAssign(int v) { m_type = Type.int_; m_int = v; return v; }
+	int opAssign(int v)
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.int_;
+		m_int = v;
+		return v;
+	}
 	/// ditto
-	long opAssign(long v) { m_type = Type.int_; m_int = v; return v; }
+	long opAssign(long v)
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.int_;
+		m_int = v;
+		return v;
+	}
 	/// ditto
-	double opAssign(double v) { m_type = Type.float_; m_float = v; return v; }
+	BigInt opAssign(BigInt v)
+	{
+		if (m_type != Type.bigint)
+			initBigInt();
+		m_type = Type.bigint;
+		m_bigint = v;
+		return v;
+	}
 	/// ditto
-	string opAssign(string v) { m_type = Type.string; m_string = v; return v; }
+	double opAssign(double v)
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.float_;
+		m_float = v;
+		return v;
+	}
+	/// ditto
+	string opAssign(string v)
+	{
+		if (m_type == Type.bigint)
+			finiBigInt();
+		m_type = Type.string;
+		m_string = v;
+		return v;
+	}
 	/// ditto
 	Json[] opAssign(Json[] v)
 	{
+		if (m_type == Type.bigint)
+			finiBigInt();
 		m_type = Type.array;
 		m_array = v;
 		version (VibeJsonFieldNames) { if (m_magic == 0x1337f00d) { foreach (idx, ref av; m_array) av.m_name = format("%s[%s]", m_name, idx); } else m_name = null; }
@@ -195,6 +290,8 @@ struct Json {
 	/// ditto
 	Json[string] opAssign(Json[string] v)
 	{
+		if (m_type == Type.bigint)
+			finiBigInt();
 		m_type = Type.object;
 		m_object = v;
 		version (VibeJsonFieldNames) { if (m_magic == 0x1337f00d) { foreach (key, ref av; m_object) av.m_name = format("%s.%s", m_name, key); } else m_name = null; }
@@ -221,6 +318,7 @@ struct Json {
 			case Type.null_: return Json(null);
 			case Type.bool_: return Json(m_bool);
 			case Type.int_: return Json(m_int);
+			case Type.bigint: return Json(m_bigint);
 			case Type.float_: return Json(m_float);
 			case Type.string: return Json(m_string);
 			case Type.array:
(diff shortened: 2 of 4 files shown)


Geod24:
Using ulong just pushes the limit, it doesn't really fix the problem. Me think supporting `BigInt` could definitely fix this. @s-ludwig ?

source/vibe/data/json.d
```diff
@@ -108,6 +110,7 @@ struct Json {
 		Null = null_,           /// Compatibility alias - will be deprecated soon
 		Bool = bool_,           /// Compatibility alias - will be deprecated soon
 		Int = int_,             /// Compatibility alias - will be deprecated soon
+		Uint = uint_,           /// Compatibility alias - will be deprecated soon
```
Geod24: I doubt we need to add a compatibility alias for a new feature.

etcimon:
I also vote for a BigInt here

s-ludwig:
`BigInt` would also have the advantage of better forward compatibility with [std_data_json](https://github.com/s-ludwig/std_data_json/). So I'd  also be in favor of that solution.

IgorStepanov:
So, you are all prefer `BigInt`? Ok. Should it replace the long value, or there should be both types?
In the second case, how should work parsing of int values?

s-ludwig:
It should work similar to the current `ulong` solution. So all integral numbers fitting in `long` should still be stored as a `long` and anything bigger as a `BigInt`. The only issue is that the naive implementation of this (convert `string` to `BigInt` and the `BigInt` to long if possible) will probably have a non-trivial performance impact. But for cases where that matters, we could introduce a `version (VibeJsonDisableBigInt)` that uses the old `string` -> `long` parsing mode.

IgorStepanov:
Ok, I'll do it.

IgorStepanov:
@s-ludwig done

source/vibe/data/json.d
```diff
@@ -92,28 +117,59 @@ struct Json {
 		}
 	}
 
+	version (VibeUseJsonBigInt) {
+		long bigIntToLong() inout
```
s-ludwig: Missing `private`

source/vibe/data/json.d
```diff
@@ -69,20 +68,46 @@ import std.traits;
...
+		{
+			version (VibeUseJsonBigInt)
+				BigInt bigint;
+			long long_;
+			double double_;
+			bool bool_;
+			string string_;
+			Json[string] object;
+			Json[] array;
+		}
```
s-ludwig: We need to keep the workaround unless the issues mentioned in the comment don't apply anymore.

source/vibe/data/json.d
```diff
@@ -181,6 +253,20 @@ struct Json {
 	/// ditto
 	long opAssign(long v) { m_type = Type.int_; m_int = v; return v; }
 	/// ditto
+	version (VibeUseJsonBigInt) {
+		BigInt opAssign(BigInt v)
+		{
+			if (v >= long.min && v <= long.max) {
```
s-ludwig: I think making this choice only in the parser is a better idea. It could be surprising to have code like this trigger an error: `void test(BigInt num) { Json j = num; assert(j.type == Json.Type.bigInt); }`

s-ludwig:
Thanks @IgorStepanov! I'm leaning towards making the `version` only apply to the parsing stage, but to keep general support for `BigInt` anyway. The current approach just seems to make it too likely to break foreign code (such as a `final switch (json.type)`). Thoughts?

Apart from my comments above, seems to look fine. I'll do a more thorough review later.

source/vibe/data/json.d
IgorStepanov: fixed

source/vibe/data/json.d
IgorStepanov: fixed

IgorStepanov:
>I'm leaning towards making the version only apply to the parsing stage, but to keep general support for BigInt anyway. The current approach just seems to make it too likely to break foreign code (such as a final switch (json.type)). Thoughts?

Ok, I've changed the code. See the second commit.
However, If we talk only about parsing, we may simply check the number length in `skipNumber`, and construct BigInt only if it needed.
Note, JSON doesn't support non-trivial integer literals like octal, or hex thus parsing of JSON ints is so easy.
We may simply compute result while parsing a number (`num *= 10; num += (s[idx]) = '0');`) and check the overflow at each step. And no special version.

marcioapm:
Sorry for being late to the party, but what about supporting cent/ucent, instead? :)
At first sight it seems like a more natural extension than BigInt.

I'm not sure what the state of this is in DMD but I remember it was worked on recently for LDC.

Seems like a bad spot to be in for performance sensitive scenarios - deciding between not supporting values >= 2^63 and < 2^64 or using BigInts which require extra heap storage currently, whereas the cent would fit in the Json value's buffer - converting cent back to ulong is also pretty fast.

etcimon:
https://github.com/D-Programming-Language/dmd/blob/895eb9caf2e76208b52a48231410ce27bf37f996/src/mtype.c#L318

IgorStepanov:
@marcioapm `cent` isn't supported in dmd and it doesn't solve the general problem: json standard allows an unlimited numbers. 
For the floating point numbers a trouble is less essential: all float types allow to represent any real number (-inf, +inf) with some error.
For an integer number json, current parser may reject a correct value which greater then long.max. I is not good.

IgorStepanov:
@s-ludwig I've add an overflow check to `skipNumber`, and removed `version (VibeJsonDisableBigInt)`:
now I build a `BigInt` value only if number is greater then `long.max`.

source/vibe/data/json.d
IgorStepanov: I missed this comment, sorry. And I don't understand it. Do you tell about replacing old void*[2] payload with this union?
We don't know the BigInt size (and we can't be sure, that it will not be changed in future), thus a union, I think, is the safest choise.

source/vibe/data/json.d
```diff
@@ -108,12 +128,12 @@ struct Json {
 		Null = null_,           /// Compatibility alias - will be deprecated soon
 		Bool = bool_,           /// Compatibility alias - will be deprecated soon
 		Int = int_,             /// Compatibility alias - will be deprecated soon
+		BigInt = bigint,       /// Compatibility alias - will be deprecated soon
```
schuetzm: This is unnecessary.

IgorStepanov: Ok, removed.

source/vibe/data/json.d
s-ludwig: Yes exactly. I did the inverse some time ago to fix those issues.

You can do `void*[max((BigInt.sizeof+(void*).sizeof-1)/(void*).sizeof, 2)] m_data;` to guarantee correctness. We could try switching back to a `union`, but I know that I had severe issues with it in the past, so that we should be really sure that it doesn't reintroduce those.

IgorStepanov: Why `void*`?
Why not `ubyte[max(BigInt.sizeof, (void*).sizeof * 2)] m_data;`?

IgorStepanov: Ok, I've changed it as you say. Is it ok now?
BTW, what troubles with union do you have met?

IgorStepanov: ldc raises a strange error with `void*[]` inside biguintcore. See test results.
I'm trying `ubyte[]`, but I think, it will fail too.

IgorStepanov: Ok, I've added explicit initialization of m_bigint before assigning. Now all works.
Rebased.
Is it ok now?

Geod24:
Could you squash your commits into a single one ?

examples/json/source/app.d
```diff
@@ -17,4 +17,12 @@ void main()
 	Json parent = obj;
 	parent.remove("item1");
 	foreach (i; obj) writeln(i);
+
+	version (VibeJsonDisableBigInt) {
+	} else {
```
Geod24: This seem to be the only occurence of the version left.

examples/json/source/app.d
```diff
@@ -17,4 +17,12 @@ void main()
 	Json parent = obj;
 	parent.remove("item1");
 	foreach (i; obj) writeln(i);
+
+	version (VibeJsonDisableBigInt) {
+	} else {
+		import std.bigint;
+		auto obj2 = parseJsonString(`{"serial":17559991181826658461}`);
+		writeln("serial: ", obj2["serial"]);
+		assert(obj2["serial"] == BigInt(17559991181826658461UL));
```
Geod24: Also, it'd be good to have a test that test number outside of a size_t boundary, to ensure that this part is also working.

source/vibe/data/json.d
```diff
@@ -44,7 +44,7 @@ unittest {
 public import vibe.data.serialization;
 
 public import std.json : JSONException;
-import std.algorithm : equal, min;
+import std.algorithm : equal, min, max;
```
Geod24: Minor: I would suggest that we reduce, and even remove, any selective import at the top. Until 313 and 314 are fixed, those cause symbols leakage, and upgrading a version of Vibe.d which removes one of the import might break code. And when 313 / 314 are fixed, it'll definitely breaks a lot of code...

source/vibe/data/json.d
```diff
@@ -69,14 +69,23 @@ import std.traits;
...
 		// because of internal pointers. This crude workaround seems to fix
 		// the issues.
-		void*[2] m_data;
+		void*[max((BigInt.sizeof+(void*).sizeof-1)/(void*).sizeof, 2)] m_data;
 		ref inout(T) getDataAs(T)() inout { static assert(T.sizeof <= m_data.sizeof); return *cast(inout(T)*)m_data.ptr; }
+
+		void initBigInt()
+		{
+			BigInt init_;
+			(cast(ubyte*)m_data.ptr)[0 .. BigInt.sizeof] = (cast(ubyte*)&init_)[0 .. BigInt.sizeof];
```
Geod24: So that's the LDC bug I guess. Do you have an issue number, or something ? I think anyone reading this in a 6+ months time would be like "WHAT? WHY?".

IgorStepanov:
>Could you squash your commits into a single one ?

All commits represents different approaches. If the my last approach will be considered best, I squash commits. Otherwice, it will be much easy to rollback to previous version.

source/vibe/data/json.d
```diff
@@ -181,6 +199,8 @@ struct Json {
 	/// ditto
 	long opAssign(long v) { m_type = Type.int_; m_int = v; return v; }
 	/// ditto
+	BigInt opAssign(BigInt v) { m_type = Type.bigint; initBigInt(); m_bigint = v; return v; }
```
Geod24: IDK how `BigInt` are implemented, but I can see the cycle: 'blitting the default value' -> 'assigning data' -> 'Do it again' as potentially problematic: if `BigInt` use `malloc`ed memory, it's more likely that on assignement, either `malloc` or `free` (or both) happen.

source/vibe/data/json.d
IgorStepanov: Do you suggest to move `std.algorithm : max` inside struct Json?
I can't move/remove `import std.algorithm : equal, min` because it irrelevant with this PR.

examples/json/source/app.d
IgorStepanov: Ok. Please wait few minutes.

examples/json/source/app.d
IgorStepanov: My bad. I'll remove it.

Geod24:
> All commits represents different approaches. If the my last approach will be considered best, I squash commits. Otherwice, it will be much easy to rollback to previous version.

That request is just to make the history easier to understand. I often find myself `blame`ing a file, to see when the code was introduced, and why. Having to go through various modifications for the same P.R. just eats a lot of time (same goes for history).

As per the rollback, with git, you can easily checkout a branch that points to your current commit, and squash them together. `git` never destroys old objects - unless nothing points to it, and they are gc'ed, which is ~ 2 weeks later, but won't happen with another branch -.

source/vibe/data/json.d
Geod24: I'll just make it `import std.algorithm` and call it a day.

source/vibe/data/json.d
IgorStepanov: I think it is not a bug. Simply dmd and gdc uses another BigInt implementation: with inline asm.
The general idea: BigInt.init contains a BigUInt instance, which initially should contatin a [0] array.
When we get BigInt via `getDataAs`, it may contain the garbarge (null or something worse) (and BigUInt invariant is fault). Thus, we should explicitly write BigInt.init to storage before the first using.

source/vibe/data/json.d
IgorStepanov: >IDK how BigInt are implemented, but I can see the cycle

Default value may be only static value, because BigInt is a struct, and BigInt.init should be known at compile time.
Thus everytime, `a`, `b` and `c` in the following example has a identical (bitwise) value. Thus I simply write BigInt.init to the m_data.
```
    BigInt a;
    BigInt b;
    BigInt c;
```

source/vibe/data/json.d
IgorStepanov: >I think anyone reading this in a 6+ months time would be like "WHAT? WHY?".

I should add the comment, yes?

source/vibe/data/json.d
IgorStepanov: done

examples/json/source/app.d
IgorStepanov: done

examples/json/source/app.d
IgorStepanov: done

source/vibe/data/json.d
Geod24: What I was refering to was, if someone calls `opAssign` with a `BigInt` value multiple time, the finalizers of all but the last 'instance' will not be run.

IgorStepanov: This case is fixed.
`initBigInt()` is called only if old `m_type` in assign is not `Type.bigint`

source/vibe/data/json.d
IgorStepanov: Comments have been added.

source/vibe/data/json.d
IgorStepanov: `/tests/redis/source/app.d` used equal from `json`. I've added an import to app.d

source/vibe/data/json.d
```diff
@@ -181,6 +209,15 @@ struct Json {
 	/// ditto
 	long opAssign(long v) { m_type = Type.int_; m_int = v; return v; }
 	/// ditto
+	BigInt opAssign(BigInt v)
+	{
+		if (m_type != Type.bigint)
+			initBigInt();
```
Geod24: IIUC, in the following code:
````
Json bi;
bi = BigInt("42");
bi = 42L;
bi = BigInt("42");
````

The bug is still present, isn't it ?

s-ludwig:
@IgorStepanov WRT `union` vs `void*[...]`, the issues I had were memory leaks, but also it broke using `std.algorithm.swap` (and thus `std.sort` etc.). `void*` needs to be used instead of `ubyte` to tell the GC to look for pointers within the array. If we'd store for example a `Json[]` within an `ubyte[]` array, it may simply get collected even if it's still referenced.

s-ludwig:
@Geod24 The destructor issue is actually a good point. In fact, the same issue is in [std_data_json/`JSONNumber`](https://github.com/s-ludwig/std_data_json/blob/066eacbbeeec13dd437466ad62ea3d92e7f3ee46/source/stdx/data/json/lexer.d#L1159) and nobody has noticed it yet (probably because it looks so natural with the `union` syntax). Maybe it would be worth to at least add a warning when a type with destructor is used within a `union` and that union is stored in a struct without destructors (same for postblit and possibly opAssign).

Also, I'll have to find the tagged union implementation for Phobos again that I had started some time ago and submit a PR. `Algebraic` is (not so) nice for many things, but often it's simply desirable to work with a type `enum` instead of `TypeInfo`s and one shouldn't have to mess with this kind of low level stuff in regular code.

source/vibe/data/json.d
IgorStepanov: Ok, I've fixed it. (see the below description)

IgorStepanov:
@s-ludwig @Geod24 I've fixed BigInt destruct issue.
Now when we try to call `opAssign` for BigInt objects, and save non-BigInt to it, `finiBigInt` is called.
`finiBigInt` simply creates an automatic `BigInt init_` variable, and `swap`s it with the stored BigInt value.
When `finiBigInt` is finished, old BigInt value will be destroyed with `init_`, and `m_bigint` now contains a `BigInt.init` value, which can be rewritten with any non-BigInt value without any possible leaks.

s-ludwig:
Looks good to me now. I'll just make some minor, mostly cosmetic, adjustments after merge. Does anyone else still see any issue?

Geod24:
Well... I look forward to `std.data.json` :D

s-ludwig:
Good thing is that this change improves compatibility with `std.data.json`.

Merging now. Thanks Igor for your work!

s-ludwig:
Okay, got the changes in. The only interface change has been to rename `Type.bigint` to `Type.bigInt`.

IgorStepanov:
>Okay, got the changes in. The only interface change has been to rename Type.bigint to Type.bigInt.

Thanks! 
BTW, when do you plan to release the next vibe.d version (with this patch)? I mean, that when I'll able to fetch it with `dub`.
I use vibe.d in a commercial product, and if the next release will not be soon, I'll write temporary code to handle this issue until the next version will be released.

s-ludwig:
I could tag an alpha/beta release right away.

s-ludwig:
Okay 0.7.24-beta.1 is tagged now.

IgorStepanov:
>Okay 0.7.24-beta.1 is tagged now.

Thanks a lot! :)

MartinNowak:
???
Javascript doesn't have integer types at all, it uses double-precision (64-bit) float for any Number.
[math - What is JavaScript's highest integer value that a Number can go to without losing precision? - Stack Overflow](http://stackoverflow.com/questions/307179/what-is-javascripts-highest-integer-value-that-a-number-can-go-to-without-losin)
Try `alert(01234567890123456789);`.

s-ludwig:
I don't think that this really matters, since this is JSON and not JavaScript. JSON doesn't specify anything about the numerical representation, so the implementation is basically free to choose whatever fits. Since large integers are used in practice, it would be quite arbitrary to not support them.

PR #1120, opened 2015-05-22
Exit event loop after WM_QUIT message

aka-demik:
This allow stop event loop from non-D thread without calling dll_thread_attach().


aka-demik:
Any chance of get this in v0.7.24?
There are only four lines of code. But thanks to them, I have an opportunity of a stronger separation of threads.

s-ludwig:
Okay finally managed to take a look and LGTM!

PR #1121, opened 2015-05-22
Fix infinite connection attempt

etcimon:
The connectTCP function was returning correctly even though the connection attempt had failed.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -249,8 +249,8 @@ final class LibasyncDriver : EventDriver {
 
 		enforce(conn.run(&tcp_connection.handler), "An error occured while starting a new connection: " ~ conn.error);
 
-		while (!tcp_connection.connected) getDriverCore().yieldForEvent();
-		
+		while (!tcp_connection.connected && !tcp_connection.m_error) getDriverCore().yieldForEvent();
+		enforce(!tcp_connection.m_error, tcp_connection.m_error);
 		tcp_connection.m_tcpImpl.localAddr = conn.local;
 		
 		if (Task.getThis() != Task()) 
@@ -960,6 +960,7 @@ final class LibasyncTCPConnection : TCPConnection {
 
 		bool m_closed = true;
 		bool m_mustRecv = true;
+		string m_error;
 
 		// The socket descriptor is unavailable to motivate low-level/API feature additions 
 		// rather than high-lvl platform-dependent hacking
@@ -1027,8 +1028,11 @@ final class LibasyncTCPConnection : TCPConnection {
 		scope(exit) releaseReader();
 		
 		while( m_readBuffer.empty ){
-			checkConnected();
+			if (!connected)
+				return 0;
+			m_settings.reader.noExcept = true;
 			getDriverCore().yieldForEvent();
+			m_settings.reader.noExcept = false;
 		}
 		return m_readBuffer.length;
 	}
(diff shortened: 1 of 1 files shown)


s-ludwig:
Seems like the change in log.d slipped in. Not sure if that workaround is safe to remove now (DMD 2.065).

etcimon:
The writeln in core.d? Yes I'll remove that. It took me hours to find why I coudn't run my application with `/SUBSYSTEM:WINDOWS`, turns out writeln is going to throw an invalid descriptor exception in std.stdio and `Throwable.toString` refuses to attach a stack trace to it

s-ludwig:
Okay, I suspected that when reading this... it's been a while since I last tested a `/SUBSYSTEM:WINDOWS` app (using console during development). Can you check if `write()` instead of `write("")` also fixes the issue? The idea is just to access anything from `std.stdio` to cause it to get initialized.

etcimon:
Yes I can try that, the error came from LockingTextWriter actually

s-ludwig:
I've tried now and even with the `write("");` I don't get an error for pure GUI applications. On which OS exactly did you have the error? Do you still get it now with the `write` call?

s-ludwig:
Or maybe this was fixed in Phobos in the meantime?

etcimon:
I got it with 2.068 on windows, using /SUBSYSTEM:WINDOWS in the lflags. I got around it by removing this enforcement: https://github.com/D-Programming-Language/phobos/blob/d9f44820953e33315712b08a7cf3fd5c49b3064b/std/stdio.d#L2377

And using `if (fp !is null)` here: https://github.com/D-Programming-Language/phobos/blob/d9f44820953e33315712b08a7cf3fd5c49b3064b/std/stdio.d#L197

s-ludwig:
I just can't reproduce it at all (Using DMD 2.068/32-bit)... hm.. did you maybe use the VisualC++ runtime instead of the DigitalMars one (COFF)?

etcimon:
> I just can't reproduce it at all (Using DMD 2.068/32-bit)... hm.. did you maybe use the VisualC++ runtime instead of the DigitalMars one (COFF)?

AH yes, I'm building with COFF on x86_64

s-ludwig:
Okay, I'll try with that then.

etcimon:
My application is a combined installer/service/launcher, it first evaluates if it has admin rights - if not it elevates, then it checks if the webserver is on - if not it checks if the service is installed - if not it installs a service otherwise runs it, then it launches a browser. I might release it once I have the launchd Mac version running, there's a new objective-c wrapper which is not currently merged here:  http://diveframework.com/

s-ludwig:
Using `File f; f.close();` as a dummy operation instead of `write("");` fixed it. Can you rebase and remove the change here?

So your application would be useful as a generic management tool for server style applications? Sounds like it could be useful for things like vibelog or similar, too, at least for GUI-affine users. I didn't know about the Dive framework, it sounds great to have. But I was really waiting for someone to finally merge https://github.com/D-Programming-Language/dmd/pull/4321

etcimon:
> I didn't know about the Dive framework, it sounds great to have. But I was really waiting for someone to finally merge D-Programming-Language/dmd#4321

Well, the dive framework was developed in association with the objective C attributes in that pull.

> So your application would be useful as a generic management tool for server style applications? Sounds like it could be useful for things like vibelog or similar, too, at least for GUI-affine users. 

It's really just a new branch of desktop applications. At least anything you'd do with Qt will be do-able with vibe.d, with the help of javascript user interfaces. Why learn Qml or embed a browser in a Window when you can interface with a client-side server? 

It would seem to have a promising future to me at the very least. Everyone knows javascript, so there's no learning curve, so instantly you've destroyed the "learning curve" argument against D.

etcimon:
> Using File f; f.close(); as a dummy operation instead of write(""); fixed it. Can you rebase and remove the change here?

yep

s-ludwig:
Thanks!

etcimon:
Hmm, can you remove the `processTimers` in `TCPConnection.waitForData`? It is called from within a fiber and I got a few errors on my fork with it

s-ludwig:
Done.

PR #1123, opened 2015-05-29
Use swap in FreeListRef.opAssign to fix #1122

### LINKED ISSUE #1122: memory leak example
i have created a vibe application which calls a few backend services (it shovels data into amazon kinesis). unfortunately once put into production on a heavy traffic website it starts to eat all my memory.

maybe i'm doing something completely wrong or it may be related to #1117 .
I suspect that the http client leaks memory (too).

example is here: https://github.com/yannick/vibed-crash-example/tree/memleak
(i just put it into a branch and reused the example for #1099  ). more or less it just uses requestHTTP within a server to call backend services.

any help appreciated

etcimon:
This is an attempt only, more valgrind would be needed.

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -1496,7 +1496,8 @@ private bool handleRequest(Stream http_stream, TCPConnection tcp_connection, HTT
 	scope(exit) request_allocator.reset();
 
 	// some instances that live only while the request is running
-	FreeListRef!HTTPServerRequest req = FreeListRef!HTTPServerRequest(reqtime, listen_info.bindPort);
+	HTTPServerRequest req = FreeListObjectAlloc!HTTPServerRequest.alloc(reqtime, listen_info.bindPort);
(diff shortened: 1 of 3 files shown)


s-ludwig:
We definitely need to investigate why `FreeListRef` causes a memory leak (I'd actually doubt that `FreeListRef` is responsible directly since it works without signs of a memory leak for `VibeManualMemoryManagement`)! AFAICS, the original `opAssign` implementation should work as well as the `swap` solution. For `swap`, I'm just hoping that it doesn't accidentally call `opAssign` recursively due to `swap`'s implementation, at least on older DMD versions (it used to be a trivial `T tmp = a; a = b; b = tmp;` if I remember right).

etcimon:
This was unnecessary after all. The issue didn't repeat itself, I was looking at 2 requests of data in valgrind and the buffers use a FreeListRef so I was quick to accuse it ;p

PR #1127, opened 2015-06-06
REST header serialization, optional REST parameters

fast4shoot:
This fixes the way REST headers are serialized (converting to literal values or JSON, instead of just calling `.toString()`), while also adding the ability to have optional parameters in REST methods, akin to the web interface.

Also added is a `const` version of `toRepresentation` of `DictionaryList`.

### DIFF
--- a/source/vibe/utils/dictionarylist.d
+++ b/source/vibe/utils/dictionarylist.d
@@ -59,6 +59,12 @@ struct DictionaryList(VALUE, bool case_sensitive = true, size_t NUM_STATIC_FIELD
 		foreach (k, ref v; this) ret ~= FieldTuple(k, v);
 		return ret;
 	}
+	/// ditto
+	const(FieldTuple)[] toRepresentation() const {
+		const(FieldTuple)[] ret;
+		foreach (k, ref v; this) ret ~= const(FieldTuple)(k, v);
+		return ret;
+	}
 
 	/** Removes the first field that matches the given key.
 	*/
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -433,18 +433,22 @@ class RestInterfaceClient(I) : I
 			auto query = appender!string();
 
 			foreach (idx, p; params) {
-				static if (!(idx % 2)) {
-					// Parameter name
-					static assert(is(typeof(p) == string), "Internal error ("~__FUNCTION__~"): Parameter name is not string.");
-					query.put('&');
-					filterURLEncode(query, p);
-				} else {
-					// Parameter value
-					query.put('=');
-					static if (is(typeof(p) == Json))
-						filterURLEncode(query, p.toString());
-					else // Note: CTFE triggers compiler bug here (think we are returning Json, not string).
-						filterURLEncode(query, toRestString(serializeToJson(p)));
+				// gets value corresponding to the current parameter
+				auto value = params[(idx / 2) * 2 + 1];
+				// generates condition for inclusion of this parameter
+				enum condition = isNullable!(typeof(value)) ? q{ !value.isNull } : q{ true };
+				if (mixin(condition))
+				{
+					static if (!(idx % 2)) {
+						// Parameter name
+						static assert(is(typeof(p) == string), "Internal error ("~__FUNCTION__~"): Parameter name is not string.");
+						query.put('&');
+						filterURLEncode(query, p);
+					} else {
+						// Parameter value
+						query.put('=');
+						filterURLEncode(query, toRestString(p));
+					}
 				}
 			}
 			return query.data();
@@ -469,7 +473,11 @@ class RestInterfaceClient(I) : I
 					nextName = p;
 				} else {
 					// Parameter value
-					jsonBody[nextName] = serializeToJson(p);
+					static if (isNullable!(typeof(p))) {
+						if (!p.isNull) jsonBody[nextName] = serializeToJson(p);
+					} else {
+						jsonBody[nextName] = serializeToJson(p);
+					}
 				}
 			}
 			debug return jsonBody.toPrettyString();
@@ -523,7 +531,7 @@ unittest
 }
 
 
-/**
+/**dictionary list
 	Encapsulates settings used to customize the generated REST interface.
 */
 class RestInterfaceSettings {
@@ -599,14 +607,19 @@ private HTTPServerRequestDelegate jsonMethodHandler(T, string method, alias Func
 					// @headerParam.
 					static if (paramsArgList[0].origin == WebParamAttribute.Origin.Header) {
 						// If it has no default value
-						static if (is (ParamDefaults[i] == void)) {
+						static if (is (ParamDefaults[i] == void) && !isNullable!P) {
 							auto fld = enforceBadRequest(paramsArgList[0].field in req.headers,
 							format("Expected field '%s' in header", paramsArgList[0].field));
 						} else {
 							auto fld = paramsArgList[0].field in req.headers;
-								if (fld is null) {
-								params[i] = ParamDefaults[i];
-								logDebug("No header param %s, using default value", paramsArgList[0].identifier);
+							if (fld is null) {
+								static if (is (ParamDefaults[i] == void)) {
+									params[i] = P();
+									logDebug("No header param %s, using empty nullable", paramsArgList[0].identifier);
+								} else {
+									params[i] = ParamDefaults[i];
+									logDebug("No header param %s, using default value", paramsArgList[0].identifier);
+								}
 								continue;
 							}
 						}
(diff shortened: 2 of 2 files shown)


Geod24:
I'll review in depth on Monday.
In the meantime, it'll need an addition to example/rest.

fast4shoot:
I'll get on the example right away.

I also noticed that I forgot to add support for Nullable to body params on the REST server side. I'll fix that and most likely create a new pull request, together with the example.

source/vibe/utils/dictionarylist.d
```diff
@@ -59,6 +59,12 @@ struct DictionaryList(VALUE, bool case_sensitive = true, size_t NUM_STATIC_FIELD
 		foreach (k, ref v; this) ret ~= FieldTuple(k, v);
 		return ret;
 	}
+	/// ditto
+	const(FieldTuple)[] toRepresentation() const {
+		const(FieldTuple)[] ret;
+		foreach (k, ref v; this) ret ~= const(FieldTuple)(k, v);
+		return ret;
+	}
```
schuetzm: Try merging these two functions with [`inout`](http://dlang.org/function.html#inout-functions):
```d
inout(FieldTuple)[] toRepresentation() inout {
    inout(FieldTuple)[] ret;
    foreach (k, ref v; this) ret ~= inout(FieldTuple)(k, v);
    return ret;
}
```

Geod24: :+1:

fast4shoot: I tried doing that, but the compiler complains:

    ../../source/vibe/utils/dictionarylist.d(59): Error: cannot implicitly convert expression (v) of type const(FilePart) to inout(FilePart)
    ../../source/vibe/inet/webform.d(142): Error: template instance vibe.utils.dictionarylist.DictionaryList!(FilePart, true, 1LU) error instantiating

Something is wrong with FilePart's const-correctness, I guess.

source/vibe/web/rest.d
```diff
@@ -433,18 +433,22 @@ class RestInterfaceClient(I) : I
...
+					static if (!(idx % 2)) {
+						// Parameter name
+						static assert(is(typeof(p) == string), "Internal error ("~__FUNCTION__~"): Parameter name is not string.");
+						query.put('&');
+						filterURLEncode(query, p);
+					} else {
+						// Parameter value
+						query.put('=');
+						filterURLEncode(query, toRestString(p));
+					}
```
Geod24: I'm not sure the indexing logic is what you expected: `idx / 2 * 2 + 1` => [0, 1, 2, 3, 4, 5] => [1, 3, 3, 5, 5, 5]
So you never reference an even parameter (which would be a bug), but you reference the fitfh parameter thrice in a row. That means the bug would exhibit with 3 or more query parameters.

The logic in itself is quite hard to grasp (but the code was already obscure). I wouldn't mind an explanation in the code which explains why the `if` wraps the `static if` and not the other way around, and/or a little example of the code generated.

source/vibe/web/rest.d
```diff
@@ -969,12 +987,13 @@ private string genClientBody(alias Func)() {
 				static if (is(PT == Json))
 					url_prefix = q{urlEncode(id.toString())~"/"};
 				else
-					url_prefix = q{urlEncode(toRestString(serializeToJson(id)))~"/"};
+					url_prefix = q{urlEncode(toRestString(id))~"/"};
```
Geod24: Why ?
I'm not a big fan of this line, but that'll break existing code. Unless it's required by this P.R. I suggest to create a separate one that tackles the Json problem (and gives motivation).

source/vibe/web/rest.d
```diff
@@ -1008,7 +1027,7 @@ private string genClientBody(alias Func)() {
 				if (p.startsWith(":")) {
 					foreach (pn; ParamNames) {
 						if (pn.startsWith("_") && p[1 .. $] == pn[1 .. $]) {
-							request_str ~= q{ ~ urlEncode(toRestString(serializeToJson(%s)))}.format(pn);
+							request_str ~= q{ ~ urlEncode(toRestString(%s))}.format(pn);
```
Geod24: Same comment as on line 990

source/vibe/web/rest.d
fast4shoot: [0, 1, 2, 3, 4, 5] shouldn't result in [1, 3, 3, 5, 5, 5], but in [1, 1, 3, 3, 5, 5], therefore indexing every odd element, which would be the value of the parameter that's currently being processed. 

This is needed so that the code can decide whether or not it should emit the parameter if it's an nullable. And this decision must be done for both the parameter name and the parameter value.

Although I do agree that it's a bit unclear. I'm thinking of rewriting it so that the foreach processes only even (or odd, doesn't really matter) elements and then handle both the name and value at the same time.

source/vibe/web/rest.d
```diff
@@ -969,12 +987,13 @@ private string genClientBody(alias Func)() {
...
-					url_prefix = q{urlEncode(toRestString(serializeToJson(id)))~"/"};
+					url_prefix = q{urlEncode(toRestString(id))~"/"};
 			} else static if (anySatisfy!(mixin(GenCmp!("ClientFilter", i, ParamNames[i]).Name), paramAttr)) {
 				alias paramsArgList = Filter!(mixin(GenCmp!("ClientFilter", i, ParamNames[i]).Name), UDATuple!(WebParamAttribute, Func));
-				static if (paramsArgList[0].origin == WebParamAttribute.Origin.Header)
-					param_handling_str ~= format(q{headers__["%s"] = to!string(%s);}, paramsArgList[0].field, paramsArgList[0].identifier);
-				else static if (paramsArgList[0].origin == WebParamAttribute.Origin.Query)
+				static if (paramsArgList[0].origin == WebParamAttribute.Origin.Header) {
+					static if (isNullable!PT) param_handling_str ~= format(q{ if (!%s.isNull) }, paramsArgList[0].identifier);
+					param_handling_str ~= format(q{headers__["%s"] = toRestString(%s);}, paramsArgList[0].field, paramsArgList[0].identifier);
```
Geod24: On the other hand, this introduce Json for the header params.

source/vibe/web/rest.d
Geod24: > [0, 1, 2, 3, 4, 5] shouldn't result in [1, 3, 3, 5, 5, 5], but in [1, 1, 3, 3, 5, 5], therefore indexing every odd element, which would be the value of the parameter that's currently being processed.

Exact, my bad.

> Although I do agree that it's a bit unclear. I'm thinking of rewriting it so that the foreach processes only even (or odd, doesn't really matter) elements and then handle both the name and value at the same time.

You could do it with `continue`, although you'll need to add some unittests - This area is bug prone (compiler bug, sadly).

source/vibe/web/rest.d
fast4shoot: You're right, there's a possibility of breaking applications that rely on headers being serialized via a simple `to!string` call, and that's a problem.

On the other hand, should anyone really be sending non-scalar values over HTTP headers? I don't think so. But if they seriously need to do it, shouldn't it be done the same way it's done everywhere else?

Or maybe we need to introduce a general mechanism that allows users to change how different parameters are serialized.

source/vibe/web/rest.d
fast4shoot: >You could do it with `continue`
Either that or just putting all the logic inside the `static if (!(idx % 2))` branch, without the `else`.

source/vibe/web/rest.d
Geod24: Sorry for the late reply, was quite busy lately.

> On the other hand, should anyone really be sending non-scalar values over HTTP headers? I don't think so. But if they seriously need to do it, shouldn't it be done the same way it's done everywhere else?

My personal opinion is that you should only serialize an aggregate as an aggregate if the representation allows you to. For example, it's perfectly fine to serialize an aggregate as body parameter (if we're using Json) because you're free to put whatever you want in the body, and `Content-Type` allows you to type it.
OTOH, query parameters and header parameters don't have a structure. Each parameter is essentially a single field. There are very little use cases where you want to pass an aggregate and see it serialized as Json, and in such situation, I think we should consider using `Json` parameters instead of aggregates.

I'd like to free aggregates for query and header parameters for grouping of parameters. For example, Github API has a [rate limit](https://developer.github.com/v3/#rate-limiting) which defines 3 parameters which are related to each others. Those would be good candidate for such usage. Example in code:

```D
public struct GhRate {
  @headerParam("X-RateLimit-Limit")
  uint limit;
  @headerParam("X-RateLimit-Remaining")
  uint remaining;
  @headerParam("X-RateLimit-Reset")
  time_t reset;
}

interface GhRoot {
  // Note: This uses 'out' parameters as implemented in: #1091
  Json get(out GhRate rate);
}
```

Does that sound right ?

Geod24:
@fast4shoot : This needs a rebase. Also, I see that Github list the branch as being in `unknown repository`, did you delete and re-forked Vibe.d by any chance ?

source/vibe/web/rest.d
fast4shoot: That actually sounds like a good way to do it, although it would be a brekaing change as well, because query parameters are passed as serialized Json right now.

fast4shoot:
Yes, unfortunately I did. I am currently in the process of fixing some of the issues, although I haven't had much time to do it lately.

I'll then probably take just the Nullable parts of this PR and submit that separately.

Geod24:
Okay, then I guess you can close this (you have no way to edit this P.R.) and reopen it once you're ready, with a link to this.

PR #1128, opened 2015-06-08
Add websockets and CONNECT tunnel support to the HTTP proxy

georgi-d:
- added accessors  to the raw streams in the http request and response
- added handling of CONNECT in the proxy request handler
- added handling of websockets upgrade in the proxy request handler

dub test passes
ran tests that websockets proxying works
ran tests that CONNECT proxying works

### DIFF
--- a/source/vibe/http/proxy.d
+++ b/source/vibe/http/proxy.d
@@ -67,12 +67,43 @@ HTTPServerRequestDelegateS reverseProxyRequest(HTTPReverseProxySettings settings
 		auto rurl = url;
 		rurl.localURI = req.requestURL;
 
+		//handle connect tunnels
+		if (req.method == HTTPMethod.CONNECT) {
+			if (!settings.handleConnectRequests)
+			{
+				throw new HTTPStatusException(HTTPStatus.methodNotAllowed);
+			}
+
+			auto ccon = connectTCP(settings.destinationHost, settings.destinationPort);
+			auto scon = res.connectProxy();
+			assert (scon);
+
+			import vibe.core.core : runTask;
+			runTask({ ccon.write(scon); });
+
+			scon.write(ccon);
+			return;
+		}
+
+		//handle protocol upgrades
+		auto pUpgrade = "Upgrade" in req.headers;
+		auto pConnection = "Connection" in req.headers;
+
+
+		import std.algorithm : splitter, canFind;
+		import vibe.utils.string : icmp2;
+		bool isUpgrade = pConnection && (*pConnection).splitter(',').canFind!(a => a.icmp2("upgrade"));
+
 		void setupClientRequest(scope HTTPClientRequest creq)
 		{
 			creq.method = req.method;
 			creq.headers = req.headers.dup;
-			creq.headers["Connection"] = "keep-alive";
 			creq.headers["Host"] = settings.destinationHost;
+
+			//handle protocol upgrades
+			if (!isUpgrade) {
+				creq.headers["Connection"] = "keep-alive";
+			}
 			if (settings.avoidCompressedRequests && "Accept-Encoding" in creq.headers)
 				creq.headers.remove("Accept-Encoding");
 			if (auto pfh = "X-Forwarded-Host" !in creq.headers) creq.headers["X-Forwarded-Host"] = req.headers["Host"];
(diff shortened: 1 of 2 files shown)


source/vibe/http/proxy.d
```diff
@@ -67,12 +67,50 @@ HTTPServerRequestDelegateS reverseProxyRequest(HTTPReverseProxySettings settings
...
+		auto pConnection = "Connection" in req.headers;
+
+		auto isUpgrade = false;
+
+		import std.array : split;
+		import std.string : strip, toLower;
+		if( pConnection ) {
+			auto connectionTypes = split(*pConnection, ",");
+			foreach( t ; connectionTypes ) {
+				if( t.strip().toLower() == "upgrade" ) {
```
9il: icmp?

georgi-d: yes, I will change it

georgi-d:
I have addressed 9il's comment and rebased to the TOT master

s-ludwig:
Hm, looks like I never submitted the reply that I had written. If I remember right, I had two questions:
- The WebSocket tunnel addition looks good. I never realized that this doesn't work, yet. However, I think it should be safe to drop the "websocket" comparison and make it work for all upgrade requests.
- What is the motivation for CONNECT support in a *reverse* proxy? In a forward proxy it's clear, but here it seems like a security risk, it should at least be an opt-in feature.

Lastly, it would be ideal, instead of adding those new and unsafe `tunnelStream` properties, to use the existing `upgradeProtocol` methods in `HTTPServerResponse` and `HTTPClientResponse`. I'm not 100% sure if the API is sufficient for this use case, but it should be in principle.

georgi-d:
I am fine with dropping the "websocket" comparison, I put it there just as additional check since I am not sure that all "upgrade" requests should be handled as transparent tunnels. If I remember correctly upgrades to http2 should also be handled this way. 

As for the CONNECT tunnel support we do use it in the company I am working for being able to terminate a SSL handshake between the client and the actual service rather then the proxy. This is used for some types of authentication.

The way it is implemented in this change it is not a security risk IMHO because it always connects to the predefined host and port and not the one specified in the CONNECT header. 

My idea is that there would be a router like handler in-front of the proxy handler which would look up the path in the request and proxy to specific internal host:port based on it. The router handler can have logic to deny requests (either CONNECT) or any other ones based on the requested path or CONNECT host:port.

I have not added it to the current pull request but in my project I have a LongestPathMatchRouter HTTP Handler implementation which based on the path chooses which back-end service to proxy to by calling the proxy handler with the correct destination host:port.
 
If you prefer I do not mind moving the CONNECT handling in another handler but it felt more natural to handle all the HTTP verbs in one location.

I will look into the `upgradeProtocol` methods again and see if they can be used instead.

georgi-d:
I checked and did not find a method `upgradeProtocol`, I did find a method called `switchProtocol` which is very close to what I need and tunnelStream does but cannot be used because it does some extra modification to the state. The server `switchProtocol` adds headers to the response while in the proxy we need to just check and forward the headers. The client `switchProtocol` checks for the presence  of the "Upgrade" headers and throws if they are not there while in the proxy we need to only check for them and if they are not there continue as if Upgrade was not requested at all.
For the CONNECT case `switchProtocol` cannot be used as well because in it there are no extra headers involved.
Actually `switchProtocol` can be implemented on top of `tunnelStream` but cannot substitute it.

s-ludwig:
Sorry, I meant `switchProtocol` indeed. I've changed it now slightly to accept an empty string (or `null`) as an argument to skip looking at the "Upgrade" header. The rest of the checks can stay there for the proxy, because they just enforce correct (required) HTTP protocol use. See cfe41544374f2a6152839d30ee37688f4ef21405.

The point of `switchProtocol` is to make it as hard as possible (without a linear type system or similar means) to misuse the API. `tunnelStream` requires a specific sequence of preceding commands, as well as great care when managing the response lifetime. `switchProtocol` on the other hand (at least in case of the callback based version, but the other one can be made safe, too) ensures a correct sequence of events without requiring and specific additional calls. Equally important, it also makes sure that the HTTP protocol can't get corrupted.

georgi-d:
I checked the change in cfe4154 and it will work for the websocket case but not for the CONNECT case. For the CONNECT case we need a version of the HTTPServerResponse.switchProtocol that does not write a response back to the client. In this case the standard states that the server should directly switch to a tunnel transferring raw bytes between the client and the server. 

Can you add an overload that does this? Then I would be able to update the change to use switchProtocol.

georgi-d:
I have updated the change to use switchProtocol and handle all protocol upgrades not just websockets.

I still needed to add accessors to the raw stream for the connect proxy case.
Rebased to the TOT

georgi-d:
Hi,

Do you have any additional concerns about the current state of the change?

s-ludwig:
I just had a look and I think it should be good to go like this. The only thing that I'd still want to add is to make the `CONNECT` handling opt-in using a `HTTPReverseProxySettings.handleConnectRequests` option. I obviously forgot to send the reply to that earlier. So my main concern is that default CONNECT support would make it possible to exploit security wholes and gather information about the target server that would otherwise be blocked by the proxy. Another concern in this direction is that it might open up an easily overlooked way to circumvent request filters or safety measures, e.g. when filtering by request URI before passing a server request to the reverse proxy, or when implementing client request rate limits.

georgi-d:
Hi,

I have added a handleConnectRequests member to the proxy settings class and a check for it. If the handling of CONNECT requests is not allowed a 405 Method Not Allowed error is returned.

s-ludwig:
Okay, thanks a lot! I'll merge the PR once 0.7.27 is out in hopefully a few days.

PR #1131, opened 2015-06-10
Add travis testing for GDC 4.9.2 (FE 2.066, yay)

Geod24:
I tried to include libasync, but: https://travis-ci.org/Geod24/vibe.d/jobs/66281101 (FYI @ibuclaw ).
Also, while searching if GDC 4.9.2 was supported, I stumbled accross https://github.com/travis-ci/apt-package-whitelist/ , so I took the opportunity to remove the old hack for libev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ d:
   - ldc-0.14.0
   - ldc-0.15.1
   - gdc-4.9.0
+  - gdc-4.9.2
 
 before_install:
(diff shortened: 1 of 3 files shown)


ibuclaw:
Reduce please.  Should be simple because it's got the symbol name and type.  Only need to work out who owns the problem.

s-ludwig:
Always good to see hacks go :+1:

PR #1132, opened 2015-06-10
Add callback for SMTPConnectionType.startTLS to vibe.mail.smtp.sendEmail

nathanchristenson:
Callback is missing for startTLS - just a one line change


s-ludwig:
Thanks! I remember looking at that line, intending to do the same change.

PR #1135, opened 2015-06-13
[REST] Allow empty path to be treated as '/'

Geod24:
The first steps taken are to assert on an empty / null path in the router.

I then considered a special handling for empty string as path, which would basically make it work like '/'.
But that would be quite an annoying change, but it could be potentially dangerous on update. So there's now an assert in `vibe.web.common.extractHTTPMethodAndName`.
In addition, I realized that the web interface was using it, so I made the error message clearer. The change involved is quite trivial.

Note about unrelated changes: I (finally) set up editor config on my emacs. Which means everytime I save a file I edited, it is normalized to be conformant to the settings in `.editorconfig`, which is why there is so much noise.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -94,7 +94,9 @@ final class URLRouter : HTTPServerRequestHandler {
 	@property string prefix() const { return m_prefix; }
 
 	/// Returns a single route handle to conveniently register multiple methods.
-	URLRoute route(string path) { return URLRoute(this, path); }
+	URLRoute route(string path)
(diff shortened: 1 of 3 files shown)


s-ludwig:
Disallowing `@path("")` sounds fine, but why do you also want to remove `get()` as `GET /`? This has been in use almost from the beginning on and would potentially break a lot of code for no gain. It should be allowed for both, the web and the REST interface generators.

Geod24:
Updated to transform `""` into "/"`.

Issue #1140, opened 2015-06-14
REST breaks queries?


IgorStepanov:
I wrote a simple rest interface to the OrientDB server.

```
interface IOrientDBRoot
{
    @property IOrientDBQuery query();
}



interface IOrientDBQuery
{
    @method(HTTPMethod.GET)
    @path(":db_name/sql/:query/:result_set_size")
    Json sql(string _db_name, string _query, int _result_set_size);
}
```
I called sql method with `_query="select DownloadedDateTime from Message"` and I got an error.
In logs I see the following line:
`GET /query/twitter_data/sql/select%2520DownloadedDateTime%2520from%2520Message/1 HTTP/1.1`
However, correct query should looks `/query/twitter_data/sql/select%20DownloadedDateTime%20from%20Message/1`

This bug looks like rest inteface handler calls filterURLEncode twice: `filterURLEncode(filterURLEncode("select DownloadedDateTime from Message"))`
Is it really bug, I do something wrong?

IgorStepanov:
Yes, the path is encoded in the `vibe.web.rest` module [first time](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/web/rest.d#L1011)
and after that, it done in [URL.localURI](https://github.com/rejectedsoftware/vibe.d/blob/4070d0645b94c68471109a76b18ba18ececb6399/source/vibe/inet/url.d#L164).
What we may do with it?

Geod24:
I can reproduce. We can probably ditch this URLEncode. I'll write a complete test case.

Geod24:
#1141

IgorStepanov:
@Geod24 Thanks! I've written one possible regression test in PR disscussion

Geod24:
Since #1143 gets merged, this can be closed

IgorStepanov:
Yep.

PR #1141, opened 2015-06-14
Fix issue #1140 - Path entries are url encoded twice

Geod24:


tests/vibe.web.rest.1140/source/app.d
```diff
@@ -0,0 +1,52 @@
...
+	override Json sql(string _db_name, string _query, int _result_set_size) {
+		assert(_db_name == Param1, _db_name);
+		assert(_query == Param2, _query);
+		assert(_result_set_size == Param3, to!string(_result_set_size));
+		return Json.emptyObject;
+	}
+}
+
+enum Param1 = "twitter_data";
+enum Param2 = "select DownloadedDateTime from Message";
```
IgorStepanov: please also try to use another query: "foo/bar"
It should be finally encoded as "foo%2Fbar", not as two parts of url path (`foo` and `bar`), I think.

Geod24: Right. That might explain the initial design. Will update.

IgorStepanov: I may suggest another one solution (looks like ugly hack=)
You may encode argument at `source/vibe/web/rest.d:1011` and after `Path` will be constructed [here](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/web/rest.d#L373), decode parts separately:
```
auto name_path = Path(name);
immutable(PathEntry)[] name_path_entries;
name_path_entries.reserve(name_path.nodes.length);
foreach(cur; name_path.nodes)
{
    name_path_entries ~= immutable(PathEntry)(filterURLDecode(cur.toString()));
}
url ~= Path(name_path_entries);
``` 
In this case, you also should remove assert from PathEntry [constructor](https://github.com/rejectedsoftware/vibe.d/blob/4070d0645b94c68471109a76b18ba18ececb6399/source/vibe/inet/path.d#L365): now path entry may contain `/` characters. Anyway they will be encoded.

Geod24: I just went for an `urlDecode` in the server, this way it should work as expected.

IgorStepanov: It works only if we use both client/server with vibe.
I use rest API to connect to orientdb server. It takes sql query and returns Json answer.
Thus, server decoding doesn't help in this case.

Geod24: *Shoots bullet*
Gah, I see. I don't like the idea of undoing work because we've done it too early. In addition, removing this assert is a Bad Idea, since it wouldn't catch other bugs.
I probably need to sleep over this, thanks for reviewing :)

IgorStepanov: Ok. Will return to this issue tomorrow:)

>In addition, removing this assert is a Bad Idea, since it wouldn't catch other bugs.

We should answer to the simple question: Can Path part contain '/' character? (Not relevant with REST). 
For example, I got create a file with name '\' on linux:
```
[wazar@localhost Proj]$ echo "hello" | cat > '\'
[wazar@localhost Proj]$ cat '\'
hello
```
This example already breaks your assert.
Of course, URL  path part may contain '/' character and it should be encoded as '%2f' (similar to ' ', '?', ':' which also may be in URL part and which should be encoded):
https://specs.webplatform.org/url/webspecs/develop/#path-segment
>A path segment must be zero or more URL units, __excluding "/" and "?"__.
>The URL units are __URL code points__ and __percent-encoded bytes__.
>The URL code points are ASCII alphanumeric, "!", "$", "&", "'", "(", ")", "*", "+", ",", "-", ".", _"/"_, ":", ";", "=", _"?"_
>A percent-encoded byte is "%", followed by two ASCII hex digits. Sequences of percent-encoded bytes, after conversion to bytes, should not cause a utf-8 decoder to run into any errors.

Ergo: URL path segment can't contain '/'directly, however it may contain '%2f' part.
As long as `PathEntry` contains non-encoded path segmants, it may contain a '/' or '?' characters, given that they will be encoded inside `URL` struct.

IgorStepanov: @Geod24 I thought on the issue and I find the following:
Lets for the time forget about REST and solve a simple task.
I want to create an `/sql/{urlEncode(select date from files where path = '/')}/1` query.
It should looks like `/sql/select%20date%20from%20files%20where%20path%20%3D%20%27%2F%27/1`.
Is this task legit? I think so.
Now lets try to solve it.
The first try:
```
auto url = URL("http", db_host, db_port, Path("/sql/select date from files where path = '/'/1"));
url.username = db_username;
url.password = db_password;
auto req = requestHTTP(url);
```
Obviously, fail: `Path` splits the url path by '/'.
Ok. Lets make try two:
```
auto url = URL("http", db_host, db_port, Path("/sql/select%20date%20from%20files%20where%20path%20%3D%20%27%2F%27/1"));
url.username = db_username;
url.password = db_password;
auto req = requestHTTP(url);
``` 
Strange result: `url.localURI` returns `/sql/select%20date%20from%20files%20where%20path%20%3D%20%27/%27/1`: All ok, but `%2f` has been decoded back to `/`
```
url.localURI = "/sql/select%20date%20from%20files%20where%20path%20%3D%20%27%2F%27/1";
```
same result.

IgorStepanov: And strange moment:
`URL` stores path string in the `string m_pathString;` field and it contains the decoded path. When we call `localURI`, `m_pathString` converted to `Path` with decoding: `Path(urlDecode(m_pathString))`, and after that `localURI` takes `path.toString`, and calls `filterURLEncode` for it.
Note: when we constructs `URL` from Path, `urlEncode` [is not called](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/inet/url.d#L42) for it, but property `path(Path p)` [does it](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/inet/url.d#L128).
Thus, `m_pathString` may contain encoded or non-encoded string.
I suggest the next idea:
1. `m_pathString` should _always_ contain encoded string.
2. When it constructed from `Path`, string _should_ be forcibly encoded.
3. When it constructed from `string` (via `this(string url_string)` or `@property void localURI(string str)`), it _shouldn't_ be forcibly encoded, but correct encoding should be checked (`checkUrlEncoding(m_pathString)`).
4. `@property string localURI()` should use `m_pathString` as is.
5. REST handler should encode path string independently and use `@property void localURI(string str)` for setting it.

Geod24: @IgorStepanov: Thanks a lot for investigating this ! I was recently investigating the possibility of freshing those modules up, they could definitely use some love.
However, it is potentially dangerous to change the behavior of `URL` as some external applications could be using it. How do you feel about adding a `pathEncoded` method that'll do what you described ?

IgorStepanov: Yes, it would be wise.
However,  we need to change internal URL behaviour anyway (withoit changing external behaviour).
Now it provides encoding in output method (`localURI`) and sometimes in input methods (sometimes not).
We should provide encoding in _all_ input method (except new `pathEncoded`, which should only check the correct encoding of input string), and we shouldn't provide any encoding in output methods (like `localURI`). Also would be reasonable to add `invariant` to URI, which checks the correct encoding of `m_pathString` everytime.

IgorStepanov: @Geod24 I've created my version of this fix: #1143

Geod24:
@IgorStepanov : Thanks ! I sadly have more urgent matter this week, I'll just review it then :)
Letting this open until your change is merged.

PR #1142, opened 2015-06-15
Fix route registration

etcimon:
Copy-paste error I guess.

On another subject: I was getting an infinite loop in the tree match on my fork using a set of routes I can't really show here. I'll try to investigate further I only had to disable it for now, but I think the deprecation of "classical" matching strategy should be avoided right now.

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -136,6 +136,8 @@ final class URLRouter : HTTPServerRequestHandler {
 		else m_routes ~= Route(method, path, cb);
 		return this;
 	}
+	/// Adds a new route for request that match the path and method
+	HTTPRouter match(HTTPMethod method, string path, HTTPServerRequestDelegate cb);
 	/// ditto
 	URLRouter match(HTTPMethod method, string path, HTTPServerRequestHandler cb) { return match(method, path, &cb.handleRequest); }
 	/// ditto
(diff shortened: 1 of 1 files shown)


etcimon:
I think it was because of a route that was simply `*`

Geod24:
Looks like this bug has been lingering here for a long time: https://github.com/etcimon/vibe.d/commit/74b8b4b2910a2111f18ec7d79cb8a5d86a8370e1#diff-3aa06f046d3fbf944da8bbed853ce7c8L61

source/vibe/http/router.d
```diff
@@ -136,6 +136,8 @@ final class URLRouter : HTTPServerRequestHandler {
 		else m_routes ~= Route(method, path, cb);
 		return this;
 	}
+	/// Adds a new route for request that match the path and method
+	HTTPRouter match(HTTPMethod method, string path, HTTPServerRequestDelegate cb);
```
Geod24: `/// ditto` ?

Geod24:
Any chance you add some unittests ? Otherwise LGTM.

source/vibe/http/router.d
s-ludwig: Wait, I think those two lines shouldn't be in this diff at all, right? It's the same signature as above, but returns the deprecated `HTTPRouter` interfaces and has no implementation.

s-ludwig:
D'oh! I guess I introduced that when I added the `scope` type family of request handlers. Fortunately this hasn't been part of a release, yet.

s-ludwig:
> I was getting an infinite loop in the tree match on my fork using a set of routes I can't really show here. I'll try to investigate further I only had to disable it for now, but I think the deprecation of "classical" matching strategy should be avoided right now.

Can't you just replace any common character sequences with random letters in the routes and post those (assuming that those still trigger the bug, of course)? Did it hang during the graph assembly or while matching a route? It would be great to have that fixed and still deprecate/remove the old router implementation.

etcimon:
> Can't you just replace any common character sequences with random letters in the routes and post those (assuming that those still trigger the bug, of course)?

I can't trigger it at all anymore. I don't know what I did, must've been a mistake.

Btw, I'm currently testing on a vmware virtual machine and file access is very slow (for static file server). I've been motivated into writing a FileCache, in vibe.core.file you can trigger `cacheFolder(Path)` and a directory watcher will keep the file cache updated with an LRU and max cache size. I'm wondering if I should finish it up today or something else, do you think this would prove useful?

etcimon:
Tested this, the disk didn't end up being the bottleneck for full file send latency. A simple 304 is fine though. Some 50kb file produces 120ms latency vs 12ms for a 304. So it might be an issue in Windows Vmware, I'll put this on ice for now.

source/vibe/http/router.d
s-ludwig: Ping @etcimon

s-ludwig:
Having a directory watcher backed file meta data cache could probably still be useful, but good to know that it isn't as badly needed. A full file cache triggers the question of using memory mapped files. I wonder if there is any way to get that to work nicely and reliably in an AIO environment, there are certainly situations where that can be extremely useful.

etcimon:
> Having a directory watcher backed file meta data cache could probably still be useful, but good to know that it isn't as badly needed.

I haven't found it to cause a 15ms roundtrip on localhost, but I would still have to check if it's the bottleneck in other situations.

PR #1145, opened 2015-06-17
[DRAFT][REST] Document vibe.web.rest better

Geod24:
This aims to extend the documentation of Vibe.web.rest into something that's exhaustive.
In it's present state, it is **not** ready to be pulled, because it hasn't been thoroughly proof-read, the server part is not complete, and I hadn't started on the client part.

However, I'm quite happy with the coverage of the module documentation. I intend it for general (as in both client and server) behaviour, while the individual client / server are for their specific behaviour (e.g. exception behaviour for the server, optimization via `final` which is not there yet, etc...).

So if you would like to review the module part / discuss the features, feel free (guess you can put a 'discussion' tag directly ;) ).

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -1,11 +1,226 @@
 /**
-	Automatic REST interface and client code generation facilities.
-
-	Copyright: © 2012-2013 RejectedSoftware e.K.
-	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
-	Authors: Sönke Ludwig, Михаил Страшун
-*/
+ * Module to work with RESTful API, from both server and client side.
+ *
+ * This modules aims to provide a typesafe way to deal with REST API.
+ * As they are quite a lot of similarities between them, this module use
+ * D interfaces as the way to represent REST API.
+ * While REST is not HTTP specific, the vast majority of use cases rely on it,
+ * and so does this module.
+ *
+ * This module, while being usable and more convenient from both client and
+ * server side, is also usable solely for its client or server implementation.
+ * The following documentation will explain in details how this module work,
+ * without any specifics about the client or server side.
+ * Since each implementation only require the use of one symbol from here,
+ * specifics are explained on those. You can thus find server only
+ * documentation on $(D registerRestInterface) and client only documentation
+ * on $(D RestInterfaceClient).
+ *
+ * There are few advantages in using this module over manual handling:
+ * - Automatic client generation: once the interface is defined, it will be used
+ *   both by the client side and the server side, which means there is no way
+ *   to have a mismatch between your client and server code.
+ * - Automatic route generation for the server: one of the job of the
+ *   REST module is to generate the route for your API.
+ * - Automatic serialization / deserialization: Instead of doing your own
+ *   serialization and deserialization, you just design normal member functions
+ *   and let us take care of the heavy lifting. You still have the possibility
+ *   to send direct JSON input by defining a parameter as Json, or pure string
+ *   by returning a string.
+ * - Higher level representation integrated into D: Some concepts of the
+ *   interfaces, such as optional parameter or header return values feat
+ *   nicely in D by the use of `std.typecons.Nullable` and `out` parameters.
+ *
+ *
+ * The most basic interface that can be defined is as follow:
+ * ----
+ * @path("/api/")
+ * interface APIRoot {
+ *     string get();
+ * }
+ * ----
+ *
+ * It defines an API that will be constitued of a single route (see below),
+ * 'GET /api/', so if your API sits on 'http://api.example.com', performing
+ * a ' GET http://api.example.com/foo' will call your foo method, and the
+ * returned string will be passed verbatim in the response body.
+ *
+ *
+ * Route_generation:
+ * A route is a combination of an HTTP method, GET, POST, PUT, DELETE,
+ * PATCH, HEAD, OPTIONS being the most common one, but we supports any method
+ * defined in `vibe.http.common.HTTPMethod`.
+ * For each public method of the interface, a route, and only one route,
+ * has to be set. It can be done explicitly through the use of
+ * two user defined attributes, `path` and `method`.
+ *
+ * Example of an explicit route:
+ * ----
+ * @path("/api/")
+ * interface APIRoot {
+ *     // Here we use a POST method
+ *     @method(HTTPMethod.POST)
+ *	   // Our method will located at '/api/foo'
+ *	   @path("/foo")
+ * }
+ * ----
+ *
+ * When you specify an explicit path, you have full control over the route.
+ * For trivial cases such as the one shown above, it might however not be
+ * necessary, as will be shown below, but there's one case where it is:
+ * path entry.
+ *
+ * Path entries are path of the form: `@path("/users/:name")`. They are useful
+ * when one of the component of the path is a parameter to the API.
+ *
+ * ----
+ * @path("/users/")
+ * interface UsersAPI {
+ *     @path(":name")
+ *     Json getUserByName(string _name);
+ * }
+ * ----
+ *
+ * Note that the parameter is called '_name'. It is a requirement that the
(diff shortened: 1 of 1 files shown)


source/vibe/web/rest.d
```diff
@@ -1,10 +1,226 @@
-/**
-	Automatic REST interface and client code generation facilities.
+// TODO: sub-interfaces
```
Geod24: Note: This comment will obviously go. It's just a personal reminder I didn't document behaviour for sub-interfaces (e.g. it's nowhere documented why getters for sub-interface needs to be parameter-less... @Dicebot ? :P )

source/vibe/web/rest.d
```diff
@@ -1,10 +1,226 @@
...
+ *   interfaces, such as optional parameter or header return values feat
+ *   nicely in D by the use of `std.typecons.Nullable` and `out` parameters.
+ *
+ *
+ * The most basic interface that can be defined is as follow:
+ * ----
+ * @path("/api/")
+ * interface APIRoot {
+ *     string get();
+ * }
```
Geod24: I didn't document `@rootPath` for obvious reasons.
Likely, I didn't document `@rootPathFromName` yet. I don't expect this to be pulled before 0.7.24, and by then, as you might remember, I'd like to revisit the default behaviour.

source/vibe/web/rest.d
```diff
@@ -1,10 +1,226 @@
...
+ * The rules are the following:
+ * - If the parameter is passed via the headers or the query, either implicitly
+ *   or explicitly, the aggregate is viewed as a collection of single fields.
+ *   Each member is serialized following the usual rules: the name of a field
+ *   is the (possibly empty) field name of `@headerParam` or `@queryParam`,
+ *   to which the field name, or the value given via the `@name` UDA, if present,
+ *   is used.
+ * - If the parameter is passed via the body, the datastructure is serialized
+ *   as a Json sub-object of the root one. Its field name is either the one
+ *   given via `@bodyParam` or the parameter name.
```
Geod24: This whole section describe a planned feature.

source/vibe/web/rest.d
```diff
@@ -24,47 +240,77 @@ import std.typetuple : anySatisfy, Filter;
...
+ * value is being returned for the body.
+ *
+ * Non-success:
+ * In the case where you want to return an error code, you can throw an
+ * `HTTPStatusException` from within your business code. It will be turned into
+ * a Json object that'll have a `statusMessage` with the exception message.
+ * In the case of other exception being thrown, a Json object containing a
+ * `statusMessage` set to the exception message and a `statusDebugMessage`
+ * set to the complete string representation of the `Exception` (== `toString`)
+ * will be returned, with an error code of 500.
```
Geod24: I really think we should also come up with a solution to avoid throwing exceptions for the return value.

source/vibe/web/rest.d
```diff
@@ -1,12 +1,228 @@
 /**
-	Automatic REST interface and client code generation facilities.
-
-	Copyright: © 2012-2013 RejectedSoftware e.K.
-	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
-	Authors: Sönke Ludwig, Михаил Страшун
-*/
+ * Module to work with REST API, from both server and client side.
```
Dicebot: AFAIK "official" term is `RESTful API`

source/vibe/web/rest.d
```diff
@@ -1,12 +1,228 @@
...
+ *
+ * This modules aims to provide a typesafe way to deal with REST API.
```
Dicebot: I'd go straight to the point immediately "This module implements automatic mapping between D interfaces and set of HTTP handlers that define desired RESTful API. It figures out necessary HTTP routes and generates handler delegates that do any necessary serialization and URL matching."

source/vibe/web/rest.d
```diff
@@ -1,12 +1,228 @@
...
+ * As they are quite a lot of similarities between them, this module use
+ * D interfaces as the way to represent REST API.
+ * While REST is not HTTP specific, the vast majority of use cases rely on it,
+ * and so does this module.
+ *
+ * This module, while being usable and more convenient from both client and
+ * server side, is also usable solely for its client or server implementation.
```
Dicebot: I have no idea what this sentence means

source/vibe/web/rest.d
```diff
@@ -1,12 +1,228 @@
...
+ * The following documentation will explain in details how this module work,
+ * without any specifics about the client or server side.
+ * Since each implementation only require the use of one symbol from here,
+ * specifics are explained on those. You can thus find server only
+ * documentation on $(D registerRestInterface) and client only documentation
+ * on $(D RestInterfaceClient).
+ *
+ * There are few advantages in using this module over the more basic `vibe.http`
+ * package provided:
```
Dicebot: Probably worth mentioning that one in `vibe.http` is the legacy version that is going to be eventually deprecated (I presume so?)

Dicebot: Or do you mean overall utilities of that package and not `vibe.http.rest`?

s-ludwig:
Just a quick general note, please stick to the doc comment style in the [style guide](http://vibed.org/style-guide#doc-comments) to keep things consistent. Judging by the size of the documentation, it might indeed make sense to start a dedicated documentation page on vibed.org. However, I think that small parts of it could instead be moved to `vibe.web.common` or similar to distribute the weight a little bit (e.g. the part about `MethodStyle`). I'll read through this properly in the coming days.

source/vibe/web/rest.d
Geod24: Yeah, I wanted to give the pros of using this module over doing it manually.

Geod24:
> I think this is simply wrong place to put such bigger chunk of docs. Inline DDOC is supposed to provide reference manual, not learning articles - and REST module needs the latter in some structure manner. 

I agree that module doc is where you put the reference manual - exhaustive documentation -.
This is right now a bit of a mixup, but what I'm aiming for is a complete reference. Introduction articles / how-to / tutorials should come at a later time, in a separate place.

The main reason I'm doing so is that right now I'm going in one direction with this module, and I have some idea of what I want to do. But I'd hate to start piling up features inside with no clear direction available to reviewers (and even more, I'd hate to introduce a feature that would be a design mistake but makes sense on its own).

I'll review / edit this again and will host it on my server ASAP (I'll try to do it tomorrow).

Geod24:
http://vibed.mimiks.net/api/vibe.web.rest/

Now I have to review it in depth...

Geod24:
@s-ludwig : DDOX bug (the `example:` behavior) ?

image

s-ludwig:
Does Ddoc behave differently? It seems to be according to the spec (single word with a following colon starts a section).

Geod24:
My bad, I just screwed up the line break :)

source/vibe/web/rest.d
Geod24: IMO that's a bit too low-level.

source/vibe/web/rest.d
Geod24: The point is that the server and client part should be able to be used independently, so you could write an interface that allows you to query a server which runs Java. That was the motivation behind the various `@XXXXparams`. Does it makes it clearer ?

source/vibe/web/rest.d
Geod24: I went for:

```diff
- * There are few advantages in using this module over the more basic `vibe.http`
- * package provided:
+ * There are few advantages in using this module over manual handling:
```

Not mentioning http at all feels like the best course of action after all.

Geod24:
Rebased on master and updated my server: http://vibed.mimiks.net/api/vibe.web.rest/

s-ludwig:
I'm starting to slowly work through the old PRs that have been lying here far too long. I've gone through the new texts, applied any changes that I thought were necessary and have pushed it to master: 3564729e636e057255bb374935df19544622d79a

Geod24:
Thanks a lot!

PR #1146, opened 2015-06-20
Avoid involuntary DDoS when computer goes to sleep

etcimon:
If a periodic timer is set to check on a remote resource, assuming this is in a software on millions of computers, a DDoS would occur when the computer wakes up from sleep/hibernate.


s-ludwig:
I think we should do sometime along the lines of:
```
auto nskipped = (Clock.currStdTime - pt.timeout) / pt.repeatDuration;
pt.timeout += (nskipped + 1) * pt.repeatDuration;
```
I.e. round up to the next multiple of the repeat duration, so that we stay within the original time pattern.

PR #1148, opened 2015-06-22
URL encoding rework

IgorStepanov:


s-ludwig:
Thanks! Merging.

IgorStepanov:
@s-ludwig thanks! #1143 has been rebased. Please, comment it.

source/vibe/inet/url.d
```diff
@@ -187,7 +199,7 @@ struct URL {
 			str = str[0 .. qi];
 		} else m_queryString = null;
 
-		m_pathString = str;
+		m_pathString = urlEncode(str, "/");
```
s-ludwig: I didn't realize this during the review, but this causes  `m_pathString` to be encoded twice when `URL` is constructed from a string. Fixed by 9ae8806047319ca31101471e9c34a2b9f9fd3dd6.

PR #1150, opened 2015-06-24
MongoCollection: added new findAndModify method

sigod:
I really need support for all `findAndModify`'s options. So, here's the try.

Note: haven't had time to test, yet. Just wanted to hear your thoughts on method signature.

Related to #911.

### DIFF
--- a/source/vibe/db/mongo/collection.d
+++ b/source/vibe/db/mongo/collection.d
@@ -222,6 +222,41 @@ struct MongoCollection {
 		return findAndModify(query, update, null);
 	}
 
+	/// ditto
+	Bson findAndModify(T)(T arguments)
+	{
+		auto bargs = serializeToBson(arguments);
+		assert(bargs.type == Bson.Type.object,
+			"The arguments parameter to findAndModify must be a BSON object.");
+
+		Bson cmd = Bson.emptyObject;
+		cmd["findAndModify"] = m_name;
+
+		foreach (string key, value; bargs) {
+			import std.algorithm : canFind;
+
+			immutable supported_fields = ["query", "sort", "remove", "new", "fields", "upsert",
+				"bypassDocumentValidation", "writeConcern"];
+			assert(supported_fields.canFind(key), "Field '" ~ key ~ "' is not supported.");
+
+			cmd[key] = value;
+		}
+
+		auto ret = database.runCommand(cmd);
+		enforce(ret["ok"].get!double != 0, "findAndModify failed.");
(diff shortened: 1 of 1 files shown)


sigod:
Okay, tested a bit. Works flawlessly so far.

sigod:
@s-ludwig, ping

s-ludwig:
What I don't like about the API is the reliance on `bool` parameters. It makes a function call like `.findAndModify(true, {"name": "foo"}, null, false)` very opaque and it requires adding more parameters for each newly supported option. I think what we need is something like an optional `FindAndModifyOptions` argument that can configure all possible options. Alternatively, we could try to go a route closer to the dynamic JS API and allow things like:
```
struct FAM {
    string[string] query;
    bool remove = true;
    int[string] sort = ["name": 1];
}
auto ret = coll.findAndModify(FAM(["name": "foo"]));

// or

auto ret = coll.findAndModify(tuple!("query", "remove", "sort")(["name": "foo"], true, ["name": 1]));
```

sigod:
> it requires adding more parameters for each newly supported option.

I don't think there's ever will be new options.

> I think what we need is something like an optional `FindAndModifyOptions` argument that can configure all possible options.

This isn't very practical. As this optional argument almost always will be used.

> we could try to go a route closer to the dynamic JS API

This might do. (Well, there's not much options...) I'll try to implement it in upcoming days. Only question: what do you think about providing struct, such as `FAM`, to end user, for whose who like to call it like `coll.findAndModify(FAM(["foo": "bar"]))` without writing his own types?

s-ludwig:
For now, I've added a generic `findAndModifyExt` function in 9201794b28ee3b980ecc8426093d35f679b150e5, so that it's possibly to do this at all until we possibly have found a better API alternative.

sigod:
I finally took my time and wrote discussed method. What do you think of it?

sigod:
@s-ludwig ping

s-ludwig:
What would be the advantage over the current [`findAndModifyExt`](https://github.com/rejectedsoftware/vibe.d/blob/b419f50f5f3ae2fcdef289e592cab11d2842c7b4/source/vibe/db/mongo/collection.d#L232)? I think that the latter makes sense, because query and update fields will always be present and it makes the API more clear.

WebFreak001:
I think having this function is good for compatibility because then you can write something which looks like a normal mongodb call. Though this PR should probably use `among` instead of `canFind`

PR #1152, opened 2015-06-25
Test with DMD 2.068.0-b1 too

Geod24:
Reopening since I was using the wrong Github account.

This is more of a tracking issue, so that you are aware it is being worked on :)

This will go red because the following snippet doesn't compile anymore:

```D
void main()
{
  import std.string;
  enum Test = "Hello\0World";
  char[64] foo;
  foo[0 .. Test.length] = Test[];
  auto greetings = foo[0 .. foo[].indexOf('\0')];
}
```

Will report the regression ASAP.
Ping @etcimon for the segfault.


Geod24:
https://issues.dlang.org/show_bug.cgi?id=14735
https://github.com/D-Programming-Language/dmd/pull/4779
https://github.com/D-Programming-Language/phobos/pull/3446

Geod24:
Fixed in master. I'll see if I have time to build DMD and look for other regressions this week.

s-ludwig:
Beta 2 seems to work fine now. The only remaining error looks like a fault on our side.

PR #1153, opened 2015-06-25
Libasync fix 2.068.0 b1

etcimon:
I had already seen this coming, in my great master plan

ping @Geod24  ;)

### DIFF
--- a/source/vibe/appmain.d
+++ b/source/vibe/appmain.d
@@ -65,6 +65,7 @@ int main()
 		} else {
 			status = runEventLoop();
 		}
+
 		logDiagnostic("Event loop exited with status %d.", status);
 		return status;
 	}
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -557,6 +557,10 @@ final class SyslogLogger : Logger {
 
 	unittest
 	{
+		version(VibeLibasyncDriver) {
+			import libasync.threads;
+			spawnAsyncThreads();
+		}
 		import vibe.core.file;
 		auto fstream = createTempFile();
 		auto logger = new SyslogLogger(fstream, Facility.local1, "appname", null);
--- a/dub.json
+++ b/dub.json
@@ -39,7 +39,7 @@
 		},
 		{
 			"name": "libasync",
-			"dependencies": {"libasync": "0.7.3"},
+			"dependencies": {"libasync": "0.7.5"},
 			"versions": ["VibeUseNativeDriverType", "VibeLibasyncDriver"]
 		},
 		{
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -1731,10 +1731,3 @@ unittest {
 	static assert(needsMove!V);
 	static assert(!needsMove!W);
 }
-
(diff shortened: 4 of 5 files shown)


source/vibe/appmain.d
```diff
@@ -65,6 +65,13 @@ int main()
 		} else {
 			status = runEventLoop();
 		}
+
+		
+		version(VibeLibasyncDriver) {
+			import libasync.threads : destroyAsyncThreads;
+			destroyAsyncThreads(); // destroy threads
+		}
```
Geod24: So IIUC, they can be safely destroyed multiple time ?

source/vibe/core/log.d
```diff
@@ -594,6 +598,9 @@ final class SyslogLogger : Logger {
 		assert(lines[5] == "<138>1 0000-01-01T00:00:00.000001 - " ~ BOM ~ "appname - - - " ~ BOM ~ "αβγ\n");
 		assert(lines[6] == "<137>1 0000-01-01T00:00:00.000001 - " ~ BOM ~ "appname - - - " ~ BOM ~ "αβγ\n");
 		removeFile(fstream.path().toNativeString());
+		
+		version(VibeLibasyncDriver)
+			destroyAsyncThreads();
```
Geod24: If we need this, it probably means this should be an integration test :D

source/vibe/appmain.d
etcimon: It can only be destroyed once and for all actually. I should probably look into that more

source/vibe/core/log.d
etcimon: Ah yes, that's because the logger uses a file stream which requires the thread pool on this driver :-p

Thinking about the logger, I had to do some adjustments to remove GC allocations so that I could use it in destructors: https://github.com/etcimon/vibe.d/blob/master/source/vibe/core/log.d#L123

s-ludwig:
Can't the whole tear-up and tear-down process be put into `LibasyncEventDriver`? Maybe using a reference count?

etcimon:
> Can't the whole tear-up and tear-down process be put into LibasyncEventDriver? Maybe using a reference count?

Would be nice, I'm not sure which object and at which point I would trigger the ref count incr/decr though

s-ludwig:
Shouldn't it work to increment the count in `LibasyncEventDriver.this()` and decrement it in `.dispose()`? A valid driver should also have been created whenever a unit test block gets executed.

etcimon:
> Shouldn't it work to increment the count in LibasyncEventDriver.this() and decrement it in .dispose()? A valid driver should also have been created whenever a unit test block gets executed.

Yes, that seems to work !

source/vibe/core/log.d
```diff
@@ -557,6 +557,10 @@ final class SyslogLogger : Logger {
 
 	unittest
 	{
+		version(VibeLibasyncDriver) {
+			import libasync.threads;
+			spawnAsyncThreads();
+		}
```
s-ludwig: Seems to be the same artifact as in the http2-botan-cleanup branch.

PR #1155, opened 2015-06-26
Allow paths with '/' in translationModule

schuetzm:
Example: `mixin translationModule!"path/to/po/file";`

### DIFF
--- a/source/vibe/web/i18n.d
+++ b/source/vibe/web/i18n.d
@@ -91,11 +91,13 @@ html
 			This is a complete paragraph of translated text.
 */
 
-mixin template translationModule(string NAME)
+mixin template translationModule(string FILENAME)
 {
+	import std.string : tr;
+	enum NAME = FILENAME.tr(`/.-\`, "____");
 	mixin template file_mixin(size_t i) {
 		static if (i < languages.length) {
-			enum decl_strings = extractDeclStrings(import(NAME~"."~languages[i]~".po"));
+			enum decl_strings = extractDeclStrings(import(FILENAME~"."~languages[i]~".po"));
 			mixin("enum "~languages[i]~"_"~NAME~" = decl_strings;");
 			//mixin decls_mixin!(languages[i], 0);
 			mixin file_mixin!(i+1);


s-ludwig:
That still doesn't work on DMD/Windows AFAIK (string imports cannot be in sub directories due to security considerations).

schuetzm:
It works on Linux at least... I see there's an open bug report for Windows:
https://issues.dlang.org/show_bug.cgi?id=14349

I added `\` translation, so it will start working there too once that bug is fixed.

s-ludwig:
Okay, I'll merge and add a doc comment with a warning to avoid using `/` because of the Windows behavior.

schuetzm:
Thanks!

etcimon:
> That still doesn't work on DMD/Windows AFAIK (string imports cannot be in sub directories due to security considerations).

Do you think it's going to stay this way? I think larger projects might have to consider using a preprocessor to automatically generate the views/ directory if things don't change.

s-ludwig:
I think the general stance on this was that it can't be changed, because it would be impossible to predict if a certain path really still is within a certain sub directory, thus creating a security threat. However, assuming that both the source code, as well as the contents of the sub directory are controlled by the developer, I think that checking for absolute paths and `..` entries should be sufficient.

etcimon:
> I think the general stance on this was that it can't be changed, because it would be impossible to predict if a certain path really still is within a certain sub directory, thus creating a security threat.

Yes, for those online tools allowing DMD to compile live. I think I remember. I fail to understand why executing DMD under a limited user account isn't an option.

schuetzm:
`/` isn't a valid character for paths on Windows, so it should be sufficient to normalize those to `\` (for both the paths of `import` statements and `-J` options).

s-ludwig:
> Yes, for those online tools allowing DMD to compile live. I think I remember. I fail to understand why executing DMD under a limited user account isn't an option.

I'd even never run something like that on a server without at least a chroot environment, or a virtualization solution (docker).

> `/` isn't a valid character for paths on Windows

Hm, `/` is valid since quite a while in all places/APIs that I know of.

schuetzm:
> Hm, / is valid since quite a while in all places/APIs that I know of.

On Windows? Judging from [this table](https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations), `/` is a reserved character on all Windows filesystems. Do the APIs accept it formally anyway?

s-ludwig:
It's not allowed to occur in file names, but as a path separator it is certainly accepted. At least the main Windows API function, [`CreateFile`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx) explicitly mentions `/` as valid.

etcimon:
> I'd even never run something like that on a server without at least a chroot environment, or a virtualization solution (docker).

So, in practice the use of `-J` as a forced path is unjustified I guess

PR #1156, opened 2015-06-26
Fix range violation when writing to access log

marcioapm:
When the requested URL is > ~2000 bytes the server crashes and the process hangs forever using 100% CPU on Linux. Easily exploitable by malicious users.

### DIFF
--- a/source/vibe/http/log.d
+++ b/source/vibe/http/log.d
@@ -24,7 +24,7 @@ class HTTPLogger {
 		string m_format;
 		const(HTTPServerSettings) m_settings;
 		InterruptibleTaskMutex m_mutex;
-		FixedAppender!(const(char)[], 2048) m_lineAppender;
+		Appender!(char[]) m_lineAppender;
 	}
 
 	this(in HTTPServerSettings settings, string format)
@@ -32,6 +32,7 @@ class HTTPLogger {
 		m_format = format;
 		m_settings = settings;
 		m_mutex = new InterruptibleTaskMutex;
+		m_lineAppender.reserve(2048);
 	}
 
 	void close() {}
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks. Does this happen for debug builds or just for release builds? I would have really hoped for at least a `RangeError` being thrown and the application exiting gracefully. Of course that still is a DoS vector.

etcimon:
AllocAppender?

etcimon:
> Of course that still is a DoS vector.

Biggest DoS vector is the unlimited json upload size. Unless I'm mistaken

marcioapm:
It does crash with a range error. I believe in release as well. The problem is that it then hangs there forever using 100% CPU. I think last time I checked it was stuck in _d_throw or something... It's been like this forever for a long time.

@etcimon Unlimited? Isn't maxRequestSize honored before trying to do anything with the request contents?

etcimon:
> Unlimited? Isn't maxRequestSize honored before trying to do anything with the request contents?

Sure, but if you're a little creative you can send a 2mb json multiplied by unlimited connections, because while max request size is implemented, the request timeout isn't.

etcimon:
> I think last time I checked it was stuck in _d_throw or something... It's been like this for a long time.

invalid memory operation. I get infinite loops there too, only solution is to send a signal in druntime's `onInvalidMemoryOperation`

s-ludwig:
> AllocAppender?

I think `Appender` is actually OK in this case, since it usually just allocates once at startup.

s-ludwig:
BTW, has anyone seen an open bug report on issues.dlang.org for the exception issues? What I often get is an access violation when an exception is supposed to be caught, and of course the infinite loop here.

PR #1157, opened 2015-06-26
Fix wrong request time in access log when the connection is dropped

marcioapm:


s-ludwig:
I wonder if we don't have a larger issue here. Not calling `finalize()` not only means that the request time isn't properly recorded, but also that we potentially leak resources. Do you have an exact reproduction recipe for this?

marcioapm:
I believe it's happening mostly when refreshing while loading.
If you have a route that takes slightly longer to serve, press refresh a few times, not letting it finish and it should reproduce.

s-ludwig:
Okay, I reproduced it and now `finalize()` is simply always called. I've added some checks to avoid additional log noise/exceptions.

marcioapm:
Good stuff! Thanks! :)

PR #1158, opened 2015-06-27
fix redis config command order

hpohl:
I've stumbled across this bug and just wrote a pull request real quick.


s-ludwig:
Looking good, thanks!

Issue #1144, opened 2015-06-16
REST client doesn't support (typesafe) variadic arguments


s-ludwig:
The current implementation uses `ParameterTuple` to define the parameters of the overridden function. However, `ParameterTuple` doesn't contain the ellipses part of the last parameter (e.g. `void foo(string[] bar...)`) and thus the generated method doesn't override the corresponding one in the interface. This used to work in earlier versions where the parameter declaration was done using a string mixin.

Geod24:
Out of interest, what's the use case for this ? Is it relatd to http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/25596/ ?

Geod24:
Can be closed.

Geod24:
Ping @s-ludwig

PR #1159, opened 2015-06-28
[REST] Fix issue #1144 : Support variadic typesafe arguments

Geod24:
Turns out it was way simpler than anticipated. I actually didn't expect this snippet of code to compile...


s-ludwig:
Oh wow, that's indeed a surprising language feature there. Thanks for finding a (non-disrupting) fix!

PR #1163, opened 2015-06-30
http/2 botan - Try going green

etcimon:
This is for 2.068.0 but also fixes a GC error that prevents it from going green

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -44,7 +44,7 @@ version (Windows) import core.sys.windows.winsock2;
 private __gshared EventLoop gs_evLoop;
 private EventLoop s_evLoop;
 private DriverCore s_driverCore;
-
+private shared int s_refCount; // will destroy async threads when 0
 EventLoop getEventLoop() nothrow
 {
 	if (s_evLoop is null)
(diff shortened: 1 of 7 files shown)


source/vibe/core/drivers/libasync.d
```diff
@@ -117,6 +119,11 @@ final class LibasyncDriver : EventDriver {
 		logTrace("Deleting event driver");
 		m_break = true;
 		getEventLoop().exit();
+		s_refCount.atomicOp!"-="(1);
+		if (s_refCount.atomicLoad() == 0) {
```
s-ludwig: Not important at all in this case, but you could use the return value of `atomicOp` directly to avoid another load.

source/vibe/core/log.d
```diff
@@ -604,6 +604,10 @@ final class SyslogLogger : Logger {
 
 	unittest
 	{
+		version(VibeLibasyncDriver) {
+			import libasync.threads;
+			spawnAsyncThreads();
+		}
```
s-ludwig: Is this still necessary? Shouldn't the driver be initialized here anyway?

source/vibe/core/drivers/libasync.d
etcimon: Hm I'm probably not thinking things through enough at this point. I just can't wait to mark it done :)

source/vibe/core/log.d
etcimon: Artifact probably

etcimon:
Tested the rest client too now, the stream was being freed during readAll() due to EndCallback thing so druntime generated an access violation when trying to read m_endCallback in the object invariant.

etcimon:
Also update memutils 0.4.0 tag again

PR #1170, opened 2015-07-05
[REST] Fix URL concatenation in the client to avoid double slashes

Geod24:
This:
```D
@path("/")
interface RootInterface {
    Data get();
}

        auto r = new GithubAPI(agent, "https://api.github.com");
```

Would produce the following (Note the 2 slashes):
`REST call: GET https://api.github.com// -> 404, {"documentation_url":"https://developer.github.com/v3","message":"Not Found"}`

The URL concatenation part was a little too sharp for it's own good, so I simplified it in 3 branches.

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -372,10 +372,14 @@ class RestInterfaceClient(I) : I
 
 			if (name.length)
 			{
-				string sep = "";
-				if (name[0] != '/' && (!url.pathString.length || url.pathString[0] != '/'))
-					sep = "/";
-				url.pathString = url.pathString ~ sep ~ name;
+				if (url.pathString.length && url.pathString[$ - 1] == '/'
+					&& name[0] == '/')
+					url.pathString = url.pathString ~ name[1 .. $];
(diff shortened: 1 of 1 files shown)


Geod24:
Somehow this seem to have broken master (at least, DMD 2.065):
https://travis-ci.org/rejectedsoftware/vibe.d/builds/69776342
Though the P.R. was passing: https://travis-ci.org/rejectedsoftware/vibe.d/builds/69662167

Since it was based on master (and only happen with 2.065 DMD), I believe it's a random failure. @s-ludwig any chance you could retrigger the build from Travis' UI to confirm ?

s-ludwig:
I tried to reproduce the failure yesterday, but failed at that and also dismissed it as random. No idea if/where you can trigger a manual rebuild, but I can just push make a small commit instead.

s-ludwig:
Green again, was indeed random.

Geod24:
Thanks !

FYI: You can retrigger a build by using the "refresh" button in the UI (the circle on the right, under "Settings"):
image

s-ludwig:
Ah, thanks. My fault was that I wasn't logged in.

PR #1176, opened 2015-07-09
Libasync FileStream must createTrunc earlier

etcimon:
When no data is to be written/read, createTrunc must still create the file to avoid further file operations to fail.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -426,9 +426,20 @@ final class LibasyncFileStream : FileStream {
 
 	this(Path path, FileMode mode)
 	{
-		import std.file : getSize;
+		import std.file : getSize,exists;
 		if (mode != FileMode.createTrunc)
 			m_size = getSize(path.toNativeString());
+		else {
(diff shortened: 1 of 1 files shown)


source/vibe/core/drivers/libasync.d
```diff
@@ -426,9 +426,20 @@ final class LibasyncFileStream : FileStream {
...
+		import std.file : getSize,exists;
 		if (mode != FileMode.createTrunc)
 			m_size = getSize(path.toNativeString());
+		else {
+			auto path_str = path.toNativeString();
+			if (!exists(path_str))
+			{ // touch
+				import std.c.stdio;
+				import std.string : toStringz;
+				FILE * f = fopen(path_str.toStringz, "w\0".ptr);
```
s-ludwig: You can actually just directly use `"w"` here. String constants always have an implicit zero appended in memory and implicitly cast to `const(char)*`.

etcimon: Like this?

s-ludwig:
Exactly!

PR #1177, opened 2015-07-10
implement moveFile's copy_fallback parameter

MartinNowak:


s-ludwig:
What the heck happened there?? Looks like I either partially reverted the change, or partially re-reverted it.

Thanks for catching and fixing this!

PR #1178, opened 2015-07-10
Added basic support for multi-part form uploads in HTTPClient

etcimon:
This still needs a test, but the logic should be right. These two objects (file, memory) will cover 99% of what's usually needed in multi-part form uploads. The alternative to using these is a lot of boilerplate and some specialized low-level protocol knowledge.

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -598,6 +598,14 @@ final class HTTPClientRequest : HTTPRequest {
 		bodyWriter.write(data);
 		finalize();
 	}
+	/// ditto
+	void writeBody(MultiPartPart linked_parts) 
+	{
+		headers["Content-Length"] = linked_parts.size.to!string;
+		linked_parts.read(bodyWriter);
+		finalize();
+	}
+
 
 	/**
 		Writes the response body as JSON data.
--- a/source/vibe/http/common.d
+++ b/source/vibe/http/common.d
@@ -330,6 +330,135 @@ final class MultiPart {
 	string[string] form;
 }
 
+
+/// The client multipart requires the full size to be known beforehand in the Content-Length header.
+/// For this reason, we require the underlying data to be of type RandomAccessStream
+class MultiPartPart {
+	import vibe.stream.memory : MemoryStream;
+	private MultiPartPart m_sibling;
+	private	string m_boundary;
+	// todo: MultiPartPart child;
+	protected {
+		MemoryStream m_headers;
+		RandomAccessStream m_data;
+	}
+	
+	@property void sibling(MultiPartPart part) { m_sibling = part; }
+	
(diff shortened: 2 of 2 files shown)


etcimon:
Any comments on the interface for this in general? Should I make the headers editable after the object is created?

source/vibe/http/client.d
```diff
@@ -599,6 +599,18 @@ final class HTTPClientRequest : HTTPRequest {
 		finalize();
 	}
 
+	/// Writes a multipart upload as a body to the request. This must be written in its entirety because the Content-Length
+	/// must be known beforehand.
+	/// Usage: req.writeBody(new FileMultiPart(req.headers, "Photo", "images/picture.jpg"));
```
s-ludwig: Needs to be reformatted to be rendered properly in the documentation:
```
/** Writes a multipart form as a body to the request.

	This must be written in its entirety because the Content-Length
	must be known beforehand.

	Usage: `req.writeBody(new FileMultiPart(req.headers, "Photo", "images/picture.jpg"));`
*/
```

Ideal would be to put the example into a documented unit test to verify that it doesn't get out of date.

s-ludwig:
Instead of putting the headers (and string data) into `MemoryStream`s, it would be good to instead directly write to the connection. I guess a similar approach as used for `writeJsonBody` can be used here (i.e. once generate the output on a special counting range to determine the length and then generate again and write to a `StreamOutputRange`). `formattedWrite` could also be used to make the header generation code a bit shorter.

etcimon:
> Instead of putting the headers (and string data) into MemoryStreams, it would be good to instead directly write to the connection.

I think the headers need to be kept aside until being written because the multipart forms were designed like linked lists and can be nested, and the content-length needs to be known beforehand in many browsers/parsers.

s-ludwig:
> I think the headers need to be kept aside until being written because the multipart forms were designed like linked lists and can be nested, and the content-length needs to be known beforehand in many browsers/parsers.

What about my suggestion to use a counting range to determine the length? I guess that should be pretty fast and should be an overall win over using dynamic memory allocations.

etcimon:
> What about my suggestion to use a counting range to determine the length? I guess that should be pretty fast and should be an overall win over using dynamic memory allocations.

I have a hard time visualizing it, but as long as you have the length before you start writing the body it should be fine. After all, `writePart` doesn't support chunked transfer so the name "Part" is a little misleading, it should be `writeMultiformBody` or something like that.

WebFreak001:
@s-ludwig merge either this or #1876, but you can't just leave both PRs unmerged

dariusc93:
Any update in regards to adding multipart for HTTPClient? Other than the conflicts github is reporting, is there anything else that would be required for this to get pulled into vibe-d or is there already a multipart implementation in works that isnt pushed into the master branch?

PR #1182, opened 2015-07-14
Add missing Policy param when deserializing member with @asArray

ColdenCullen:
`deserializeImp` requires a `Policy` param, and it wasn't being passed. Seems like a simple oversight.


s-ludwig:
Thanks, I'll also add a unittest after merge.

PR #1183, opened 2015-07-14
Correctly parse quoted boundary field for multi part form

### LINKED ISSUE #1167: Multi-part form parser does not handle quoted boundary strings correctly.
See http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/25900/

Marenz:
Fixes #1167

Also adds a unittest for #1167


s-ludwig:
Looks good, thanks! At some point this will need a proper parser for multiple attributes, but this is a step in the right direction.

PR #1184, opened 2015-07-15
Libasync timers not processing on idle

etcimon:
This fixes numerous issues related to timers that I had only started noticing while doing more tests on OS X.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -360,8 +360,11 @@ final class LibasyncDriver : EventDriver {
 			logTrace("Timer %s fired (%s/%s)", timer, owner != Task.init, callback !is null);
 
 			if (!periodic) releaseTimer(timer);
-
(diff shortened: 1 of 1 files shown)


etcimon:
The additional issues were related to HTTP/2 requests asking to sleep, somehow OS X hadn't triggered pending timers when more were being scheduled, and this prevented them from being scheduled. I don't know if this fixes the underlying issue though, it helped my test cases work.

etcimon:
This was reworked today to avoid processing timers inside tasks. I think ill add it along with Duration.max default timers to ready it for release

s-ludwig:
The `Duration.max` change will have to be carefully reviewed. Obviously the existing code was unfortunately quite messy w.r.t. semantics, but we'll have to evaluate the behavior changes in detail. Ideally I'd like to postpone that to the next release.

etcimon:
Ok ill at least make the change for 0 seconds to not be immediate in the libasync driver because it causes bugs right now. Libevent is implemented as 0=forever currently

PR #1185, opened 2015-07-15
Fix ConnectionPool and LocalTaskSemaphore

etcimon:
These were now tested in a [postgresql driver](https://github.com/etcimon/ddb/blob/master/source/ddb/postgres.d#L2243).


etcimon:
I think the priority ordering needs some work, a task with the highest priority should be allowed to be in front

PR #1187, opened 2015-07-16
added docs for DELETE mappings

chalucha:
new pull from rebased branch


s-ludwig:
Thanks! And sorry for the additional burden, git can be a "little" annoying during the learning phase (I like to call it the C++ of VCS).

PR #1188, opened 2015-07-17
changed iff -> if in some docs

chalucha:


s-ludwig:
This is actually intended (see <https://en.wikipedia.org/wiki/If_and_only_if>), maybe it would make sense to use `$(I iff)`, to make that more clear (or just to write it out, but that's a little verbose usually).

chalucha:
ok changed it to $(I iff) then

s-ludwig:
Okay, thanks!

PR #1190, opened 2015-07-20
Immutable data deserialization support

japplegame:
Adds the ability to deserialize immutable data. See unittest.

### DIFF
--- a/source/vibe/data/serialization.d
+++ b/source/vibe/data/serialization.d
@@ -446,8 +446,12 @@ private void serializeImpl(Serializer, alias Policy, T, ATTRIBUTES...)(ref Seria
 	} else static assert(false, "Unsupported serialization type: " ~ T.stringof);
 }
 
+private T deserializeImpl(T, alias Policy, Serializer, ATTRIBUTES...)(ref Serializer deserializer) if(!isMutable!T)
+{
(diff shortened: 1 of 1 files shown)


source/vibe/data/serialization.d
```diff
@@ -446,8 +446,12 @@ private void serializeImpl(Serializer, alias Policy, T, ATTRIBUTES...)(ref Seria
 	} else static assert(false, "Unsupported serialization type: " ~ T.stringof);
 }
 
+private T deserializeImpl(T, alias Policy, Serializer, ATTRIBUTES...)(ref Serializer deserializer) if(!isMutable!T)
+{
+	return cast(T) deserializeImpl!(Unqual!T, Policy, Serializer, ATTRIBUTES)(deserializer);
```
Geod24: I'd use `assumeUnique` here.

japplegame:
In this case assumeUnique will not help, because it works only with arrays.

source/vibe/data/serialization.d
japplegame: In this case assumeUnique will not help, because it works only with arrays.

Geod24: Wut. I always assumed it worked with everything and was just documenting.
I'm not familiar with this part, but could you also add an unittest for a `const`/`immutable` object if there's not one already ?

japplegame: I slightly modified the unittest.

Issue #1098, opened 2015-05-09
synchronized / nothrow


schuetzm:
What is the current situation with `synchronized` being `nothrow`? The comment here https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/core/drivers/libevent2.d#L624 seems to indicate that the workaround is only necessary up to DMD 2.067, but I'm getting compile errors on DMD Git master (which has recently bumped its version to 2.068):

```
source/vibe/core/drivers/libevent2.d(631): Error: '_d_monitorenter' is not nothrow
source/vibe/core/drivers/libevent2.d(631): Error: '_d_monitorexit' is not nothrow
source/vibe/core/drivers/libevent2.d(622): Error: function 'vibe.core.drivers.libevent2.Libevent2ManualEvent.emit' is nothrow yet may throw
source/vibe/utils/memory.d(131): Error: '_d_monitorenter' is not nothrow
source/vibe/utils/memory.d(131): Error: '_d_monitorexit' is not nothrow
source/vibe/utils/memory.d(124): Error: function 'vibe.utils.memory.LockAllocator.alloc' is nothrow yet may throw
source/vibe/utils/memory.d(146): Error: '_d_monitorenter' is not nothrow
source/vibe/utils/memory.d(146): Error: '_d_monitorexit' is not nothrow
source/vibe/utils/memory.d(134): Error: function 'vibe.utils.memory.LockAllocator.realloc' is nothrow yet may throw
source/vibe/utils/memory.d(160): Error: '_d_monitorenter' is not nothrow
source/vibe/utils/memory.d(160): Error: '_d_monitorexit' is not nothrow
source/vibe/utils/memory.d(149): Error: function 'vibe.utils.memory.LockAllocator.free' is nothrow yet may throw
```

etcimon:
The workaround seems to be with the `scope(failure) assert(false)` which is restricted to DMD <= 2.067 right now. On my build I changed this to DMD <= 2.068

schuetzm:
Yes, but judging from the comment, this shouldn't be necessary anymore?

Geod24:
The change was reverted in 2.067, thanks to @MartinNowak . But it was just
a delay, and I remember @s-ludwig mentionning that 'Vibe.d mutexes are
never nothrow'. Don't know what the current status is though, but it seems
safe to edit the `static if` to be `<= 2068`.

PR #1191, opened 2015-07-22
Fix win32 with 2.068

chalucha:
Fix for same problem as #1098 but on win32 driver.
And additionally resolve method name conflict with 2.068


s-ludwig:
Thanks!

PR #1193, opened 2015-07-23
Proof of concept for simple route aggregation helper

Dicebot:
Recently I was playing with simple vibe.d application for personal pleasure (https://github.com/Dicebot/mood) and found myself very unsatisfied with how route definitions for non-REST methods are separated from actual handlers in code, making it harder to maintain them in sync.

This has resulted in writing a very simple helper to register bunch of request handlers defined in a single class object in a given router. Not sure where such utility should be placed thus module name is a stub :)

### DIFF
--- a/source/vibe/web/something.d
+++ b/source/vibe/web/something.d
@@ -1,0 +1,119 @@
+module vibe.web.something;
+
+import vibe.http.router;
+import vibe.web.common;
+
+void register (T) (URLRouter router, T app)
+	if (is (T == class))
+{
+	static bool isUserMethod(string name)
+	{
+		class Empty { }
+		import std.algorithm : canFind;
+		return !canFind([__traits(allMembers, Empty)], name);
+	}
+
+	import vibe.internal.meta.uda : findFirstUDA;
+
+	foreach (method_name; __traits(allMembers, T)) {
+		static if (isUserMethod(method_name)) {
+			mixin ("alias method    = app." ~ method_name ~ ";");
+			mixin ("auto  method_dg = &app." ~ method_name ~ ";");
+
+			static assert (is(
+				typeof(method_dg) ==
+					void delegate (HTTPServerRequest, HTTPServerResponse)
+			));
+
+			enum http = findFirstUDA!(MethodAttribute, method);
+			enum path = findFirstUDA!(PathAttribute, method);
+
+			static if (path.found) {
+				static if (http.found) {
+					switch (http.value.data) {
+						case HTTPMethod.GET:
+							router.get(path.value, method_dg);
+							break;
+						case HTTPMethod.POST:
+							router.post(path.value, method_dg);
+							break;
+						case HTTPMethod.DELETE:
+							router.delete_(path.value, method_dg);
+							break;
+						case HTTPMethod.PATCH:
+							router.patch(path.value, method_dg);
+							break;
+						case HTTPMethod.PUT:
+							router.put(path.value, method_dg);
+							break;
+						default:
+							assert (false);
+					}
+				}
+				else {
+					router.any(path.value, method_dg);
+				}
+			}
+		}
+	}
+}
+
+unittest
+{
+	// default method (any)
+
+	class WebApp
+	{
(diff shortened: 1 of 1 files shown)


s-ludwig:
What about `vibe.web.web.registerWebInterface`?

s-ludwig:
Yes, it's more or less a superset. There is documentation [here](http://vibed.org/docs#web) and [here](http://vibed.mimiks.net/api/vibe.web.web/registerWebInterface). Is there anything in particular that is still missing? I'll have a look at the missing comments in the example(s).

s-ludwig:
Okay, I've added comments to the "web" and "web-i18n" examples now.

Geod24:
> Also related question - any seems to register a dozen of routes, one for each HTTP method. That feels unnecessary wasteful in terms of route lookup speed. Wouldn't it be more efficient to have a dedicated wildcard enum value for any in `HTTPMethod`?

IIUC you want something like `anyHTTP` which registers only HTTP, not WebDAV, right ?
That's a change I wasn't found of ( #984 )

s-ludwig:
> I have tried it and looks like registerWebInterface is much more restrictive in terms of class definition - it attempted to register private object fields of my application class as web interfaces of their own :)

It should exclude all non-public members (in fact it shouldn't be able to access those in the first place). Can you give a short example?

> I think trying to do anything with non-attributed symbols is impractical for such utility because, contrary to REST module, you are actually totally expected to mix route handlers with additional utilities in application class.

Such utilities can be - and IMO should be - `private` or `protected` and thus not registered as routes. But I've often thought that we should support an opt-out attribute for both, the REST and the web interfaces. I really want to keep the original name-inferred approach as a first-class citizen, even though I find myself using explicit attributes more often than I'd like (mainly due to shortcomings w.r.t. object collections).

> Also related question - any seems to register a dozen of routes, one for each HTTP method. That feels unnecessary wasteful in terms of route lookup speed. Wouldn't it be more efficient to have a dedicated wildcard enum value for any in HTTPMethod?

The router performs a simple index lookup based on the incoming HTTP method as the first step, so this is a very cheap O(1) operation. Then, for each method, there is a DFA through which the incoming URI is piped. In terms of lookup speed this is quite optimal.

s-ludwig:
> IIUC you want something like anyHTTP which registers only HTTP, not WebDAV, right ?
That's a change I wasn't found of ( #984 )

In which situation would that be? "any" meaning *any* is useful for defining catch-all routes. If "any" would just mean some methods, then that could mean bugs and security holes, because the catch-all route misses certain requests. So if an alternative sub set of methods makes sense for some other typical use case, we can add that with a different name, but it should be obvious from the name (WebDAV is "HTTP", too, tough, so something else would be good to have).

s-ludwig:
> So this is probably by design but not really convenient. I definitely don't want all public methods to be registered.

What about adding a UDA à la `@onlyAttributedRoutes`?

s-ludwig:
I thought about that (a flag in `WebInterfaceSettings`), but logically that information really belongs to the class instead of the code that registers it...

s-ludwig:
Or just a `@noRoute` attribute in `vibe.web.common` that is supported for both kinds of interface generators.

s-ludwig:
Using `@ignore` from `vibe.data.serialization` (maybe moving it to a lower level) was my first idea, but it feels a little strange to apply multiple semantics to a single attribute. I think the goal should be to document clear semantics for the attribute itself and not based on the context in which it is used. So for that reason I think an attribute defined exclusively for `vibe.web` is the cleaner approach (even though I don't see a practical, compelling use case where the serialization and the web interface meanings would clash).

What I'm using a lot for `@auth` is the pattern `enum auth = before!performAuth("_authInfo");`, as you've probably already seen in the "web" example. I found the side effect of having an additional parameter with authentication information to be quite useful. A more integrated approach has its merit, though, as the `@before` based method isn't really obvious from the docs. But I'd like to evaluate some more possibilities apart from the optional delegate approach before merging a concrete implementation - especially to make sure all use cases and authentication methods are properly addressed.

PR #1194, opened 2015-07-25
Describe FileLogger.Format

chalucha:


source/vibe/core/log.d
```diff
@@ -37,6 +37,10 @@ nothrow @safe {
 	Sets the log format used for the default console logger.
 
 	This level applies to the default stdout/stderr logger only.
+
+	Params:
+		fmt = The log format for the stdout (default is FileLogger.Format.thread)
+		infoFmt = The log format for the stderr (default is FileLogger.Format.plain)
```
s-ludwig: These two are reversed (stdout/stderr). The `FileLogger.Format.*` should be in backticks, as that will be mandatory to get cross referencing at a later point.

PR #1199, opened 2015-07-27
enforce JSON string to contain non-whitespace characters

SekiT:
When `parseJsonString` receives a string which consists of only whitespace characters, it tries to read the front of empty string at [line 1040 in original code](https://github.com/rejectedsoftware/vibe.d/blob/38784138ad8cc8f442bd0e76a1b740040ff33fe5/source/vibe/data/json.d#L1040).
Then it throws `AssertError` which is not an `Exception`.
So I fixed it by enforcing string to be not empty before taking front of it.

### DIFF
--- a/source/vibe/data/json.d
+++ b/source/vibe/data/json.d
@@ -1155,6 +1155,8 @@ unittest {
 }
 
 unittest {
+	try parseJsonString(" \t\n ");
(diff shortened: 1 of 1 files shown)


s-ludwig:
Looks good, thanks!

PR #1201, opened 2015-07-28
Fixes #1196 - Wait for redis listener to stop on Subscriber dtor

### LINKED ISSUE #1196: RedisSubscriber stops listening if the FreeListRef goes out of scope
Instead it should stay alive until the listening task has exited.

etcimon:
The task that owns RedisSubscriber will now block if it must go out of scope, waiting for ..something.. to stop the listener on its object.

This also fixes a compiler issue I had on 2.068 with `std.string.indexOf`

`fromStringz` takes a slice and uses `strlen`: https://github.com/D-Programming-Language/phobos/blob/b34ff324f0bfe539a12797f0b29359681c8c888b/std/string.d#L207

### DIFF
--- a/source/vibe/db/redis/redis.d
+++ b/source/vibe/db/redis/redis.d
@@ -603,6 +603,7 @@ final class RedisSubscriberImpl {
 		Task m_listener;
 		Task m_listenerHelper;
 		Task m_waiter;
+		Task m_stopWaiter;
 		InterruptibleRecursiveTaskMutex m_mutex;
 		InterruptibleTaskMutex m_connMutex;
 	}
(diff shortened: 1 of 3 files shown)


source/vibe/db/redis/redis.d
```diff
@@ -637,8 +638,33 @@ final class RedisSubscriberImpl {
...
 
 	~this() {
-		logTrace("~this");
-		bstop();
+		//logTrace("~this");
+		waitForStop();
+	}
+
+	// Task will block until the listener is finished
+	void waitForStop() {
```
s-ludwig: Since this is not allowed to be called concurrently, it should be `private`.

s-ludwig:
I'd ideally turn around the logic and instead of waiting to stop in the destructor, keep a reference within the listener task, so that the destructor isn't run at all before everything is stopped. But for now this fix should work.

WebFreak001:
@s-ludwig @etcimon status? merge or close?

PR #1202, opened 2015-07-31
Add struct type injection to vibe.web.web

etcimon:
I haven't tested this but I think it's more intuitive than UDAs so I'm suggesting it here.

It would also be nice to scan the class for TaskLocal!(struct) with a __ctor(req, res), and do the type injection at that level as well.

This would allow me to write more objective deserialization types, enforcemed types or authentication types in general and avoid writing @before UDAs and contextual handlers for them.

### DIFF
--- a/source/vibe/web/web.d
+++ b/source/vibe/web/web.d
@@ -611,6 +611,8 @@ private void handleRequest(string M, alias overload, C, ERROR...)(HTTPServerRequ
 				else static if (!isNullable!PT) enforceHTTP(false, HTTPStatus.badRequest, "Missing request parameter for "~param_names[i]);
 			} else static if (is(PT == bool)) {
 				params[i] = param_names[i] in req.form || param_names[i] in req.query;
+			} else static if (is(PT == struct) && __traits(compiles, { PT(req, res); }())) {
+				params[i] = PT(req, res);
 			} else {
 				static if (!is(default_values[i] == void)) {
 					if (!readFormParamRec(req, params[i], param_names[i], false))


Geod24:
Pretty neat Although from a metaprog POV, `__traits(compiles)` is a very blunt instrument, and might match in a variety of situation you wouldn't want it too.
I'd say be conservative and check that the first 2 arguments are of the right type.

@s-ludwig : W.r.t. https://github.com/rejectedsoftware/vibe.d/commit/ac618784f26b4d87cb29e29be15a4d9d2e633035#commitcomment-12469252 sadly this solution isn't applicable to the REST module (since the type has to be on the interface, so the constructor has to be defined, and then it's `@before` again).

etcimon:
Ok so if we're happy with the idea, allow me to push it further. We'd allow use of an @inject UDA on the struct ctors instead of strictly checking for (req,res) parameters, and we can resolve THOSE parameters as well to inject all the parameter's ctor's parameters recursively.

Or maybe we wouldn't even need the @inject UDA if it's the only constructor in the struct.

Geod24:
Sometimes I feel for the poor people that have name that matches UDA ;)

Why would you want to match it recursively ? I don't see a scenario where it couldn't be done from within the constructor itself, and if you want a correct implementation to match recursively, you'll have to deal with attribute hell.

I'd favor a simple, more straightforward but error-proof approach over something slightly more convenient if it doesn't bring more functionality.

etcimon:
> Why would you want to match it recursively ?

For code reuse, for convenience, for modularity, cleanliness, etc.

For example, I'd have a `struct User` that has `this(string username, string password)` for a login form or `this(Cookie session_id)` for resuming sessions. I can inject it in my handler easily and it seems fairly straightforward to re-use it internally as well.

Geod24:
What would the code look like then ?

etcimon:
For struct user?

Geod24:
Yep

etcimon:
```D
struct User
{
	long user_id;
	@vibeInjectable
	this(string username, string password)
	{
		
		auto password_query = scoped!PGCommand(conn, "SELECT password FROM users WHERE username=$1").bind(1, PGType.TEXT, username);
		{
			auto dbres = password_query.executeQuery!string().unique();
			string password_salted = dbres.front;
			bool match = true;
			foreach(immutable(char) c; password)
			{
				foreach(immutable(char) c2; password_salted)
				{
					if (c != c2) match = false;
				}
			}
			if (!match) throw new Exception("Invalid username/password");
		}
		auto user_id_query = scoped!PGCommand(conn, "SELECT id FROM users WHERE username=$1").bind(1, PGType.TEXT, username);
		auto dbres = user_id_query.executeQuery!long().unique();
		user_id = dbres.front		
	}
	@vibeInjectable
	this(Cookie session_id)
	{
		if (session_id == "") {
			startSession();
			return;
		}
		user_id = redisConn.get!long("session_" ~ session_id ~ "_user_id");
	}
	....
}

class MyWebInterface {
   void getPrivatePage(User user) {
        enforce(user.user_id != 0, "You must be logged in");
        /// do some query with user.user_id
   }
}
```

etcimon:
You'd probably need a priority to choose a constructor, so make it @vibeInjectable(int priority) where 1 is top priority

s-ludwig:
> W.r.t. ac61878#commitcomment-12469252 sadly this solution isn't applicable to the REST module (since the type has to be on the interface, so the constructor has to be defined, and then it's @before again).

I didn't get what you mean here. What we need is basically the convertion `HTTPServerRequest->T` (server) and `T->HTTPClientRequest` (client), to make this whole thing cleaner and more consistent with other parts of the library, I'd use a static method instead of the constructor: `static T fromRequest(HTTPServerRequest);`, `void toRequest(HTTPClientRequest)`. For `out` or `ref` parameters, there could also be optional access to the response object.

BTW, I think I'd also prefer something like this as the basis for a header-parameter feature, so you'd simply do `void doSomething(Header!("X-Something", string) my_header);`. This would be similar to what is already supported in the `vibe.web.web` module for validation and `Nullable!T`, and it avoids the unfortunate distance that UDAs and function parameters have in the source code.

Also kind of neat is that it enables a very concise notation for repeated use if you alias it to something shorter.

s-ludwig:
I've though about this some more in a broader context. The basic set of requirements for any solution is to support the following scenarios (in the REST interface at least, the web interface may drop some points):
- be user extensible
- automatically generated D server
- automatically generated D client
- direct D calls to the implementation interface
- automatically generated foreign language client (at least JavaScript)
- (supporting protocols other than HTTP)

The last point is something I'm currently working on and which will be vital to any progress on the status quo of the dynamic web UI front. Basically I'm seeing the following options:

1. Special handling of known types (or annotations) with fixed semantics: this succeeds on all points but the first. It wouldn't allow user defined semantics (except in a very limited setting such as `@headerParam`). It also blows up the size of any client or server implementation instead of decentralizing the functionality.

2. Use a DSL to specify semantics. This would allow translation to any non-D language. However, the implementation complexity and the learning curve is the highest of all. It wouldn't pass the non-required last point of supporting other protocols.

3. Require user implemented methods to support each use case (a.k.a. this pull request). This would require `toRequest`, `fromRequest` and, for the last case something like `toJSRequestSetup`. The implementation complexity is low, but it requires the user to implement each direction manually (error potential), and the scope of supported implementation languages (or protocols) is strictly limited to the set of implemented methods.

All three solutions support the "direct D calls" if they leave the constructors alone and let those be used for normal D code interactions. Obviously, for any special type/attribute with clear semantics, defined within vibe.d itself, we can always resort to option 1. For this reason, we should definitely provide a strong set of standard primitives, so that the need for externally defined attributes becomes lower.

The recursion idea is intersting, because it allows to build complex primitives upon simpler ones, while keeping the general semantics visible:

    struct CustomType {
        static CustomType fromRequest(HeaderParam!"Auth" auth, QueryParam!string session) { ... }
        void toRequest(ref HeaderParam!"Auth" auth, ref QueryParam!string session) { ... }
    }

However, it still requires a special third method for a JS client, because the internal semantics aren't visible.

All in all, I'm in favor of method 3 combined with a broad set of built-in primitives, considering how (in)frequent it would be to implement custom types. I'd say it would make sense to start with only simple support for `fromRequest(HTTPServerRequest)` and `toRequest(HTTPClientRequest)`, with optional `HTTPServerResponse`/`HTTPClientResponse` parameters, and possibly also an optional `error` parameter to signal processing errors without throwing exceptions. We can then evaluate the concept a bit before blowing up the implementation with full recursion support.

etcimon:
> All in all, I'm in favor of method 3 combined with a broad set of built-in primitives, considering how (in)frequent it would be to implement custom types. I'd say it would make sense to start with only simple support for fromRequest(HTTPServerRequest) and toRequest(HTTPClientRequest), with optional HTTPServerResponse/HTTPClientResponse parameters, and possibly also an optional error parameter to signal processing errors without throwing exceptions. We can then evaluate the concept a bit before blowing up the implementation with full recursion support.

I think I know where this is going. In this sense, we can talk about communication a bit like we'd talk about serialization in vibe.d, it definitely makes the API more intuitive.

I want to insist early on that we don't need to introduce any polymorphic type injection here, because the struct can implement that under the hood using any library available. However, along the years, we might start seeing some very generic types like the `User` type above, which looks for a user-defined DBO in some library like poodinis, and uses that to abstract away which database is used. We'd see other generic types use another library for injection or even a ddbc-type generic database object, and it will be too late to standardize and make them work together. So it's important to figure out what will be used for polymorphic user-defined type injection in the first place to avoid self-inflicting a death sentence from incompatibilities.

etcimon:
Now that I think of it, the javascript framework for code-generation needs to be decided upfront as well. This is a typical case where making no decision is worse than making any random decision.

s-ludwig:
> Now that I think of it, the javascript framework for code-generation needs to be decided upfront as well. This is a typical case where making no decision is worse than making any random decision, because users are left dealing with dependency hell.

I've uploaded a first proof-of-concept: #1209

I don't think we need a JS framework for this part. In fact I'd set that as a goal to avoid something like that. Connecting this would be a layer on top.

s-ludwig:
Sorry, didn't realize that the PR was in a dysfunctional state due to some uncommitted changes. Should be good now.

etcimon:
I like that JavaScript rest client, now if only I could encapsulate it in angular. The $resource API seems to be able to do all the work and there doesn't seem to be a more convenient way.

Btw do you believe vibe should offer some default SQL abstractions to get major incompatibility risks out of the way?

s-ludwig:
I'm not very familiar with Angular. Does it need some kind of special parameters, return values, or class fields/methods to use the generated code? We could add those without losing interoperability with other frameworks (or plain JS).

We could of course also generate something based on that `$resource` API, but it looks like that may not be able to express everything in our interface generator, and I think we should generally have a good reason to tie things to a particular JS framework. I like how Angular looks on the surface, but the way it operates makes me nervous in terms of performance expectations (at least for complex UIs).

But of course, with that `RestInterface` abstraction struct, the implementation of any similar generators becomes more or less trivial (see [jsclient.d](https://github.com/rejectedsoftware/vibe.d/blob/restjsclient/source/vibe/web/internal/rest/jsclient.d)), so we could in fact just offer a battery of options if we wanted.

PR #1204, opened 2015-07-31
Fix behavior of a 0 second timer, avoid processing them in a task

etcimon:
This fixes the immediate return issue in waitForData for 0 second timers, amonst other problems.

I think a negative duration might be better suited for immediate timers, since the 0 second already has defined behavior in some cases, mostly because libevent seems to never set a timer (infinite duration)

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -504,13 +504,13 @@ final class LibasyncFileStream : FileStream {
 
 	void close()
 	{
-		if (m_task != Task())
-			getDriverCore().resumeTask(m_task, new Exception("The file was closed during an operation"));
 		if (m_impl) {
 			m_impl.kill();
 			m_impl = null;
 		}
 		m_started = false;
+		if (m_task != Task() && Task.getThis() == Task())
+        		getDriverCore().yieldAndResumeTask(m_task, new Exception("The file was closed during an operation"));
 	}
 
 	@property bool empty() const { assert(this.readable); return m_offset >= m_size; }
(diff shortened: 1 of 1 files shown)


etcimon:
Don't forget this patch for release

etcimon:
Bump

source/vibe/core/drivers/libasync.d
```diff
@@ -373,26 +375,21 @@ final class LibasyncDriver : EventDriver {
...
-			processTimers();
-			next = m_timers.getFirstTimeout();
-			dur = next - now;
-		}
-		if (m_nextSched == next) {
-			logDebug("No upcoming timeouts beyond in: %s ms", (next-now).total!"msecs".to!string);
-			return;
-		}
-		else
+		if (m_nextSched != next)
```
s-ludwig: This `if` looks like it's redundant.

source/vibe/core/drivers/libasync.d
```diff
@@ -481,6 +478,8 @@ final class LibasyncFileStream : FileStream {
 
 	void close()
 	{
+		if (m_task != Task())
+			getDriverCore().resumeTask(m_task, new Exception("The file was closed during an operation"));
```
s-ludwig: This should probably be `yieldAndResultTask` instead, as it will be quite likely that `close` is called from within a task. It might also make sense to move this further down, so that when that task continues and queries `isOpen`, it will get `false` instead of `true`.

etcimon:
Bump this too: https://github.com/rejectedsoftware/vibe.d/pull/1247

s-ludwig:
Okay, looks good. Just that I unfortunately missed this before release anyway...

It needs a rebase, though (merge conflicts).

etcimon:
> It needs a rebase, though (merge conflicts).

Done

s-ludwig:
Thanks!

PR #1206, opened 2015-08-01
[post-0.7.24] Fix runWorkerTaskH when called outside of a task.

s-ludwig:
This fix has revealed some issues in the libev and libasync back ends and needs to be postponed to the next release.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -257,6 +257,14 @@ Task runWorkerTaskH(FT, ARGS...)(FT func, auto ref ARGS args)
 
 	alias PrivateTask = Typedef!(Task, Task.init, __PRETTY_FUNCTION__);
 	Task caller = Task.getThis();
+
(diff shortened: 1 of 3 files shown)


etcimon:
I tried to reproduce this but I'm getting linker errors for libasync on windows.. might need one last patch

etcimon:
Hm I tried fixing this and it's going to take a little time. I'll check back on it later.

s-ludwig:
Do you have a rough idea of what goes wrong/needs to be changed? I'd like to include this on master and either fix the failing drivers, or add assertions/skip tests for them.

etcimon:
Last I checked I stopped at the thought that it's a `runWorkerTaskH` ambuiguity in the recursion (after fixing the driver error). I think it's that template parameter inference causing issues the way they're nested.

s-ludwig:
Tests finally pass now after fixing the libev driver. The libasync issue seems to have gone, too.

(the failure is just due to a Travis internal issue)

Issue #1070, opened 2015-04-15
Generate Swagger API


Zalastax:
[Swagger](http://swagger.io/) is the best way I know to visualize your REST interface. It works by generating a JSON-document describing the API. The Swagger UI reads this document and generates a documentation page where you can read about the API functions, required schemas, and try out valid requests.
You have a REST interface generator so this should be quite straightforward to implement. I don't think it could work for freestanding routing without changes since the information about required types is never available to the router. The REST generator seem to be given an overhaul https://github.com/rejectedsoftware/vibe.d/issues/1040 so should I wait with looking into this?

s-ludwig:
Since it should only depend on the interface of the API, there is no reason to make this dependent on any other PR. I'd probably implement it in a separate module anyway, so that there is no direct overlap.

It surely looks like an interesting addition. I think someone recently proposed to support WADL (WSDL?) recently, too. What is a little sad is that, although I was sure it had been implemented long ago, there doesn't seem to be `__traits(comment)` support in the language to get the Ddoc comment, so that we'd have to resort to UDAs for the descriptions. That or creating an external tool instead of a library module...

rikkimax:
I've investigated __traits(comment) in the past. It's doable but honestly? There is no way it would get accepted. It would slow down the compiler. Memory increased wayy too much.

s-ludwig:
I seemed to remember that most people were quite exited to get it at some point. But well, memory use would be an argument (although, why would memory use ever increase more than the size of the input source code, which should not really be worth to mention?). For me, the main worry at the time was that people may start to abuse this as a poor men's replacement for UDAs - that risk isn't really there anymore today obviously.

rikkimax:
The problem is, to be able to get it via traits you need to enable keeping and parsing DDOC for every symbol. While parsing probably won't be necessary, it still would be a major issue.
So not only adding the trait is required, it's also necessary to enable storing the comments in case of needing it always.
Right now adding the trait is simple. Making it usable without -D flag. Not so much.

Realistically what is being suggested is well and truely out of scope of vibe.d. It should be a separate tool. Mostly likely as a plugin for dub or dmd itself. This is definitely web service framework territory and vibe.d is not that.

s-ludwig:
[...]

Issue #662, opened 2014-05-14
Google Discovery Services


etcimon:
I didn't know if this really belongs in dub or vibe.d, but a good hit-them-all solution for library shortage would be to port this to D and Vibe.d:
https://github.com/dart-gde/discovery_api_dart_client_generator

It seems to me that a bunch of APIs could be made available very easily with this. See the list here:
http://developers.google.com/apis-explorer/#p/

PR #1209, opened 2015-08-03
JavaScript based REST interface client

s-ludwig:


Geod24:
Any ETA for when you want this ? I'll review by the end of this week.
Though at first sight it looks like a there's a fair amount of duplication ?

s-ludwig:
I have actually started to work on the duplication issue (which is already present even without this addition), but the current `RestInterface` struct intended for this is still not ready for CT generation, because it currently requires the `RestInterfaceSettings` for a lot of things. I'll work on that later.

I'll probably create a 0.7.24 branch during the next couple of days and start merging your remaining changes in `vibe.web.rest`. Then I'd do a larger refactoring, moving things out of the `vibe.web.rest` module and switching over to use `RestInterface` there, too.

s-ludwig:
Cleaned up the commit history for review.

source/vibe/web/internal/rest/common.d
```diff
@@ -0,0 +1,407 @@
...
+	bool isIn, isOut;
+}
+
+enum ParameterKind {
+	query,       // req.query[]
+	body_,       // JSON body
+	header,      // req.header[]
+	attributed,  // @before
+	internal     // req.params[]
+}
```
Geod24: IIUC, internal is for path attribute, isn't it ? Maybe that'll be a more descriptive name. Also body parameters are not meant to stay JSON forever. On that topic: any way to check the server setting ? It's not a big deal, but if  someone disable `parseQueryString` for example, there's no way to detect it.

source/vibe/web/internal/rest/jsclient.d
```diff
@@ -0,0 +1,106 @@
+module vibe.web.internal.rest.jsclient;
```
Geod24: It could really use some documentation / documented unittest :)

examples/rest-js/dub.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "rest-js-example",
+	"description": "Simple dynamic web UI that accesses a REST interface",
+	"dependencies": {
+		"vibe-d": { "path": "../../" }
+	},
+	"versions": ["VibeDefaultMain"]
+}
```
Geod24: Nitpick: Missing newline.

As per the example that follow, there's no comment so it's a bit hard on newcomers to see how useful it is.

source/vibe/web/rest.d
```diff
@@ -248,6 +179,47 @@ unittest
 
 
 /**
+	Returns a HTTP handler delegate that serves a JavaScript REST client.
+*/
+HTTPServerRequestDelegate serveRestJSClient(TImpl)(RestInterfaceSettings settings = null)
```
Geod24: I would only allow interfaces for `Timpl` here.

source/vibe/web/rest.d
```diff
@@ -248,6 +179,47 @@ unittest
...
+	}
+
+	return &serve;
+}
+
+
+/**
+	Generates JavaScript code to access a REST interface from the browser.
+*/
+void generateRestJSClient(TImpl, R)(ref R output, RestInterfaceSettings settings = null)
```
Geod24: Interface contraint + `isOutputRange!R`

source/vibe/web/rest.d
```diff
@@ -516,7 +488,7 @@ unittest
 class RestInterfaceSettings {
 	/** The public URL below which the REST interface is registered.
 	*/
-	URL baseURL;
+	URL baseURL = URL("http://api.example.com/");
```
Geod24: Why ?

source/vibe/web/internal/rest/common.d
```diff
@@ -0,0 +1,407 @@
...
+				// determine parameter source/destination
+				if (IsAttributedParameter!(func, pname)) {
+					pi.kind = ParameterKind.attributed;
+				} else static if (anySatisfy!(mixin(CompareParamName.Name), WPAT)) {
+					alias PWPAT = Filter!(mixin(CompareParamName.Name), WPAT);
+					final switch (PWPAT[0].origin) with (WebParamAttribute.Origin) {
+						case Header: pi.kind = ParameterKind.header; break;
+						case Query: pi.kind = ParameterKind.query; break;
+						case Body: pi.kind = ParameterKind.body_; break;
+					}
```
Geod24: We could also kill `WebParamAttribute.Origin` and use `ParameterKind` instead.

Geod24:
Hum, another P.R. blocked by 2.065 :(

Also, it looks like there's a discrepancy between the client and server with the refactoring:
```console
REST route: GET /root/get_check/:param/:param2 []
add route GET /root/get_check/:param/:param2
[...]
no route match: GET /root/getCheck/one/two
No response written for /root/getCheck/one/two
REST call: GET http://127.0.0.1:8000/root/getCheck/one/two -> 404, 404
```

For the rest, the diff is large, so I'll probably have to try it out on my machine to be sure, but I like where it's heading at. I'm not that familiar with JS, but the new functionality looks cool :+1:

source/vibe/web/internal/rest/common.d
s-ludwig: It's for path attributes or any other attributes that are passed using `req.params`. Since there are valid uses outside of path placeholders, I decided against using that as the name. With the "JSON" comment I just wanted to describe the current state, since I'm not sure yet how to best handle multiple formats in terms of the internal architecture.

Currently there is no way to check the server settings. Really the only possibility that I see would be to throw an exception from inside `HTTPServerRequest` if the `.json` field is accessed without the `parseJson` setting. Another, probably better, path would be to switch the whole JSON body parsing to happen lazily and optionally strongly typed.

source/vibe/web/internal/rest/jsclient.d
s-ludwig: I'll add an example in `vibe.web.rest` and the missing module header, as well as a doc string for `generateInterface` here.

examples/rest-js/dub.json
s-ludwig: Will fix.

source/vibe/web/rest.d
s-ludwig: Hm, you are right, I copied this from the server.

source/vibe/web/rest.d
s-ludwig: Will fix.

s-ludwig:
Fixed the inconsistency (c5d3208) and 2.065 compilation error (ed739a9).

s-ludwig:
BTW, for xref purposes, `RestInterface` also makes things like #662 and #1070 an easy target.

s-ludwig:
Rebased on master. I've also fixed the handling of optional parameters. Commit ca3def1 needs some re-review, because during rebase it was just a single bit conflict with the write-to-headers change. Apart from that, everything except the default service URL comment and the possibly merge of `ParameterKind` and `WebParamAttribute.Origin` should be addressed.

I'll have a look at the latter now. Regarding the URL, I don't remember exactly anymore, but I think I got some internal assertions due to an empty service URL, which was quite unobvious. We could alternatively use something like `"http://localhost/"` or just add high-level assertions with a proper error message. The latter sounds like the better way to go...

s-ludwig:
@Geod24, did you see any remaining issues? Otherwise I'd rebase/squash and merge now.

Geod24:
I'll do a last sweep this evening ;)

Geod24:
Can't see anything.
All your project build out of the box with the new version ?

s-ludwig:
I haven't tested all of them yet, but I'll do that now.

s-ludwig:
Okay, when testing my projects I found one bug in the REST client (09ee6a27f9a57a4a9757dcc8aff073148be94d39), but none in the new code. Will merge once Travis goes green.

s-ludwig:
And thanks for reviewing this lengthy change set!

etcimon:
I would have loved to review this but got caught up in the busiest time of the past few years. Sorry!

source/vibe/web/internal/rest/jsclient.d
```diff
@@ -0,0 +1,116 @@
...
+			output.put("    var postbody = {\n");
+			foreach (p; route.bodyParameters)
+				output.formattedWrite("      %s: toRestString(%s),\n", Json(p.fieldName), p.name);
+			output.put("    };\n");
+		}
+
+		// XHR setup
+		output.put("    var xhr = new XMLHttpRequest();\n");
+		output.formattedWrite("    xhr.open('%s', url, true);\n", route.method.to!string.toUpper);
+		static if (!is(RT == void)) {
```
ColdenCullen: Sorry for not catching this sooner, but is `RT` actually defined here? I think it's causing this block to execute regardless of return type.

s-ludwig: Thanks, looks like that... `is()` and `__traits(compiles)` are really suboptimal constructs w.r.t. correctness. I'll add a test and fix this.

s-ludwig: Err.. or rather just merge your PR and add a test ;)

PR #1211, opened 2015-08-04
TimerQueue assertion failure in GC Destructors

etcimon:
Make sure the timers don't fail assertion if the GC destroys the queue before the timers are descheduled by a destructor on the GC.

### DIFF
--- a/source/vibe/core/drivers/timerqueue.d
+++ b/source/vibe/core/drivers/timerqueue.d
@@ -69,9 +69,9 @@ struct TimerQueue(DATA, long TIMER_RESOLUTION = 10_000) {
 
 	ref inout(DATA) getUserData(size_t timer_id) inout { return m_timers[timer_id].data; }
 
-	bool isPending(size_t timer_id) const { return m_timers[timer_id].pending; }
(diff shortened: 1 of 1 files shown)


s-ludwig:
Mixing GC and deterministic destruction really sucks! I wish there'd be a way to cope with such situations in a more robust way. But for now that looks like the only practical workaround.

etcimon:
Tell me about it! When chaining FreeListObjectRef destructors for objects on the GC it often segfaulted when casting to the original type. This fixed 95% of my crashes at one point:

https://github.com/etcimon/druntime/commit/0c5e1a67af4c9f658371fb37413b411442537806

s-ludwig:
...and, even though I have to admit that I'm not entirely sure how much overhead that would require for the book-keeping, I think the GC should make a best effort to finalize objects in reference dependency order. In 99% of the cases, there will be a clear ordering, which would make referencing other GC objects well-defined in the first place.

etcimon:
I think that would take a RB Tree with an auto incrementing ID and the object pointer, which would fetch individually each item during destruction to build another RB Tree and then scan through it during destruction.

PR #1213, opened 2015-08-05
http/2 : Avoid returning null streams at all costs

etcimon:
This is a fix for the client. Returning null prevents reading if the stream is empty during a stream operation

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -1257,7 +1257,13 @@ final class HTTPClientResponse : HTTPResponse {
 	*/
 	@property InputStream bodyReader()
 	{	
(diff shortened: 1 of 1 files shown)


s-ludwig:
This is a good idea, I'll add a `nullSource` function similar to the existing `nullSink` in `vibe.core.stream`. Hasn't there been a similar fix recently?

etcimon:
> Hasn't there been a similar fix recently?

Yes, I removed some assert(stream !is null) in vibe.stream.operations and had them replaced by enforcements instead, I had been trying to get null streams to work but it gets a little too complicated, maybe a `nullSource` would be more of a long-term solution.

PR #1208, opened 2015-08-10
registerRestInterface and registerWebInterface should return the router

MartinNowak:
So that the functions become chainable just like get/post/any and such.


s-ludwig:
Still not sure if method chaining is a good paradigm in D, considering the existence of `with`, but since we already have it in other places, I'd agree that it improves consistency. I'd just not want to advertise it as the standard way to access the API in the examples, maybe as a commented-out alternative right after the usual stand-alone UFCS call.

MartinNowak:
updated

s-ludwig:
Thanks!

PR #1221, opened 2015-08-13
multipart request: always skip boundary, fix #1220

### LINKED ISSUE #1220: 0.7.24: exception "Header is missing ':'"
This exception happens in 0.7.24 when we make multipart request:
```
    Bad Request
    Internal error information:
    object.Exception@../.dub/packages/vibe-d-0.7.24/source/vibe/inet/message.d(56): Header is missing ':'.
    ----------------
    ./bin/server(pure @safe bool std.exception.enforce!(Exception, bool).enforce(bool, lazy const(char)[], immutable(char)[], ulong)+0x6b) [0x8b314b]
    ./bin/server(void vibe.inet.message.parseRFC5322Header(vibe.core.stream.InputStream, ref vibe.utils.dictionarylist.DictionaryList!(immutable(char)[], false, 32uL).DictionaryList, ulong, vibe.utils.memory.Allocator, bool)+0x130) [0xa45518]
    ./bin/server(bool vibe.inet.webform.parseMultipartFormPart(vibe.core.stream.InputStream, ref vibe.utils.dictionarylist.DictionaryList!(immutable(char)[], true, 16uL).DictionaryList, ref vibe.utils.dictionarylist.__T14DictionaryListTS4vibe4inet7webform8FilePartVbi1Vmi1Z.DictionaryList, immutable(char)[], ulong)+0x8c) [0xaac22c]
    ./bin/server(void vibe.inet.webform.parseMultiPartForm(ref vibe.utils.dictionarylist.DictionaryList!(immutable(char)[], true, 16uL).DictionaryList, ref vibe.utils.dictionarylist.__T14DictionaryListTS4vibe4inet7webform8FilePartVbi1Vmi1Z.DictionaryList, immutable(char)[], vibe.core.stream.InputStream, ulong)+0x24a) [0xaac156]
    ./bin/server(bool vibe.inet.webform.parseFormData(ref vibe.utils.dictionarylist.DictionaryList!(immutable(char)[], true, 16uL).DictionaryList, ref vibe.utils.dictionarylist.__T14DictionaryListTS4vibe4inet7webform8FilePartVbi1Vmi1Z.DictionaryList, immutable(char)[], vibe.core.stream.InputStream, ulong)+0xf8) [0xaabf08]
    ./bin/server(bool vibe.http.server.handleRequest(vibe.core.stream.Stream, vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo, ref vibe.http.server.HTTPServerSettings, ref bool)+0x146d) [0xaa9b45]
    ./bin/server(void vibe.http.server.handleHTTPConnection(vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo)+0x1b3) [0xaa85d3]
    ./bin/server(void vibe.http.server.listenHTTPPlain(vibe.http.server.HTTPServerSettings).doListen(vibe.http.server.HTTPListenInfo, bool).__lambda3(vibe.core.net.TCPConnection)+0x36) [0xaa7fa2]
    ./bin/server(void vibe.core.drivers.libevent2_tcp.ClientTask.execute()+0x4b8) [0xa82c14]
    ./bin/server(void vibe.core.core.makeTaskFuncInfo!(void delegate()).makeTaskFuncInfo(ref void delegate()).callDelegate(vibe.core.core.TaskFuncInfo*)+0x72) [0x9f794a]
    ./bin/server(void vibe.core.core.CoreTask.run()+0x12e) [0x9f5a46]
    ./bin/server(void core.thread.Fiber.run()+0x2a) [0xb65042]
    ./bin/server(fiber_entryPoint+0x61) [0xb64f31]
    [(nil)]
```

0.7.23 works flawlessly.

sigod:
#1220

### DIFF
--- a/source/vibe/inet/webform.d
+++ b/source/vibe/inet/webform.d
@@ -216,7 +216,7 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 			import std.conv : to;
 			file.write(stream, (*plen).to!long);
 		}
-		else stream.readUntil(file, cast(ubyte[])boundary);
+		stream.readUntil(file, cast(ubyte[])boundary);
 		logDebug("file: %s", fp.tempPath.toString());
 		file.close();
 


s-ludwig:
I'd like to eventually add a validation that the `readUntil` actually reads 0 bytes until it hits the boundary. But that requires either adding a byte counter to `readUntil`, or using something else to consume the marker. I'll pull this in the meantime to see if it fixes the test mentioned above.

sigod:
I think it's a good idea.

Something like `stream.consumeExpected(cast(ubyte[])boundary);`?

s-ludwig:
I think I'd go with something like `enforceHTTP(stream.trySkip(cast(ubyte)[])boundary), HTTPStatus.badRequest);`, so that a proper HTTP status is generated. I'll add that and create an alpha tag for 0.7.25.

sigod:
> I think I'd go with something like enforceHTTP(stream.trySkip(cast(ubyte)[])boundary), HTTPStatus.badRequest);, so that a proper HTTP status is generated.

I thought about validation as in `readUntil`. But, yes, your version is much better. And more reusable, which is very important.

> I'll add that and create an alpha tag for 0.7.25.

Why doesn't you use SemVer normally? Just increase patch number and release new version. Because I can't really get this bug fix. I'm stuck with 3 options: use old 0.7.23 or use alpha tag or edit local source of 0.7.24. I like neither of this options.

s-ludwig:
Hm, I do intend to tag a new release shortly, but right now it doesn't make sense yet to switch to full SemVer (>=1.0.0), because more or less every release still contains some breaking changes or new features. But the plan is to split up the library into multiple pieces and tag some of them with a stable version shortly afterwards.

But you can safely use the alpha tag for now, as there are no critical changes and I will not introduce any of those until 0.7.25.

etcimon:
I'm receiving a EOF before end stream marker on this using the MultiPartPart script with FileMultiPart:

https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/inet/webform.d#L262

I don't add CRLF to the end boundary, so I'm wondering if the readLine is erroneous or if I should append CRLF.

s-ludwig:
According to [RFC 1341](http://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) there is no CRLF required after the "--", so it looks like the correct thing to do would be to always read two bytes and require them to be either `"--"` or `"\r\n"` in case of "--", it should then `nullSink.write(bodyReader);` to drop the epilogue.

etcimon:
That's what I was assuming, I'll submit a pull request sometime today or tomorrow for this

PR #1232, opened 2015-08-24
Fix Multi-part Upload

etcimon:
The end of stream is not always `--\r\n` in forms, but the boundary is always followed by `--`

### DIFF
--- a/source/vibe/inet/webform.d
+++ b/source/vibe/inet/webform.d
@@ -259,7 +259,16 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 		auto data = cast(string)stream.readUntil(boundary);
 		form.addField(name, data);
 	}
-	return stream.readLine(max_line_length) != "--";
+	
+	ubyte[2] ub;
+	stream.read(ub);
+	if (ub == "--")
+	{
+		nullSink().write(stream);
+		return false;
+	}
+	enforce(ub == cast(ubyte[])"\r\n");
+	return true;
 }
 
 /**


source/vibe/inet/webform.d
```diff
@@ -259,7 +259,20 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 		auto data = cast(string)stream.readUntil(boundary);
 		form.addField(name, data);
 	}
-	return stream.readLine(max_line_length) != "--";
+	
+	enforce(stream.leastSize >= 2);
+	auto ub = stream.peek();
```
s-ludwig: This is not a good idea in general (leastSize could return just 1 even if the connection is still open and could yield more data, and `peek` is not required to return anything at all). But in this case it's possible to just use `ubyte[2] ub; stream.read(ub);` instead. It has to be either "--" or "\r\n". Both of them terminate the form part.

etcimon: For me the peek returns `--[boundary]--`, so maybe I should read the boundary + 2

s-ludwig: That doesn't look right, since the boundary is already consumed above (lines 250/251 and 259). Maybe a `peek` bug?

etcimon: No that's a bug in the content-length reader I implemented on my fork.

source/vibe/inet/webform.d
```diff
@@ -259,7 +259,20 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
...
 	}
-	return stream.readLine(max_line_length) != "--";
+	
+	ubyte[2] ub;
+	stream.read(ub.ptr[0 .. 2]);
+	if (ub == "--")
+	{
+		static NullOutputStream null_stream;
+		if (!null_stream)
+			null_stream = new NullOutputStream;
```
s-ludwig: `vibe.core.stream.nullSink`

source/vibe/inet/webform.d
```diff
@@ -259,7 +259,20 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
...
+
+		null_stream.write(stream);
+		return false;
+	}
+	stream.readLine(max_line_length);
```
s-ludwig: I think this should rather be `enforce(ub == cast(ubyte[])"\r\n");`. Any other text between the boundary and the newline is invalid AFAICS. Also, in this case it would skip the *next* line instead of the current, because the "\r\n" has already been read by `stream.read(ub)`.

source/vibe/inet/webform.d
```diff
@@ -259,7 +259,20 @@ private bool parseMultipartFormPart(InputStream stream, ref FormFields form, ref
 		auto data = cast(string)stream.readUntil(boundary);
 		form.addField(name, data);
 	}
-	return stream.readLine(max_line_length) != "--";
+	
+	ubyte[2] ub;
+	stream.read(ub.ptr[0 .. 2]);
```
s-ludwig: Why not simply `read(ub);`?

s-ludwig:
Looks good to me now! Merging in.

PR #1235, opened 2015-08-27
Implement support for TLS connections using client certificates and X509 authentication

machindertech:


s-ludwig:
Looks good, thanks! I'll just adjust the API to keep `connectMongoDB`  as the primary entry point.

PR #1247, opened 2015-09-04
Libasync (linux) - Make sure TCP write ready events always occur

etcimon:


etcimon:
This is a pretty bad bug actually, the protocol freezes when write blocks. Should be merged urgently.

etcimon:
Tested in production as well

s-ludwig:
I didn't merge this earlier because I didn't understand why a write operation would depend on a "read ready" event. Do you have any further insights on this? Anyway, this is safe to merge, so I'll include it in the release.

etcimon:
I also have an issue understanding why I was blocking on write and receiving only a read event

PR #1252, opened 2015-09-08
move test script to a separate file

MartinNowak:
- makes it easier to manually test vibe.d
- choose decent default values for env vars


s-ludwig:
Makes sense. I'll also move the script to the tests directory.

s-ludwig:
Hm, or rather not since it also builds examples and the unit test build...

PR #1258, opened 2015-09-11
fix BUILD_EXAMPLE and provide defaults

MartinNowak:


s-ludwig:
Okay, got the issue, it was not in the test script, but in the exclude section of .travis.yml, which doesn't match the env section anymore.

PR #1259, opened 2015-09-12
Allow user-supplied implementation of `readFormParamRec`

schuetzm:
vibe.d's built-in implementation of `readFormParamRec` uses a strange convention (underscores as separators for array and AA indices), which is incompatible with the `[]` convention used by almost everyone else, including Rails, PHP, and jQuery.param().

This PR allows the user to specify their own implementation of `readFormParamRec`, enabling not only any desired parameter naming conventions, but also used-defined deserialization.


WebFreak001:
@s-ludwig no comment?

PR #1263, opened 2015-09-13
Make logTrace @nogc

etcimon:
This avoids any allocation related to logging to avoid `InvalidMemoryOperation` due to the GC calling destructors that allocate.

I didn't add the `@nogc` attribute because it isn't tested, but this simple change eliminated 99% of `InvalidMemoryOperation`s in my applications.

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -120,13 +120,58 @@ void log(LogLevel level, /*string mod = __MODULE__, string func = __FUNCTION__,*
 	try {
 		foreach (l; getLoggers())
 			if (l.minLevel <= level) { // WARNING: TYPE SYSTEM HOLE: accessing field of shared class!
-				auto app = appender!string();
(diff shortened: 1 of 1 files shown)


s-ludwig:
I wanted to change the logger to work in an output range like fashion anyway, and I've just seen that fortunately `std.experimental.logger` already works the same way. Let me quickly implement that instead, that that's off the table, too.

etcimon:
Yeh this is not good: https://github.com/D-Programming-Language/phobos/blob/master/std/experimental/logger/core.d#L823

s-ludwig:
> Yeh this is not good: https://github.com/D-Programming-Language/phobos/blob/master/std/experimental/logger/core.d#L823

Interestingly this looks the same as what I got now, but I only added it for backwards compatibility reasons. Maybe it was added to std.experimental.logger after the first release, too?

PR #1265, opened 2015-09-16
don't use TLS when being redirected from http to https

MartinNowak:
- disconnect when changing host, schema, or port
- add defaultPort to URL to simplify code

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -253,6 +253,7 @@ final class HTTPClient {
 		Rebindable!(const(HTTPClientSettings)) m_settings;
 		string m_server;
 		ushort m_port;
+		bool m_useTLS;
 		TCPConnection m_conn;
 		Stream m_stream;
 		TLSContext m_tls;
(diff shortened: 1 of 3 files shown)


source/vibe/http/client.d
```diff
@@ -244,6 +244,7 @@ final class HTTPClient {
 		HTTPClientSettings m_settings;
 		string m_server;
 		ushort m_port;
+		bool use_tls;
```
s-ludwig: Should be named `m_useTLS`.

s-ludwig:
Apart from the naming and the merge conflict in client.d, LGTM!

s-ludwig:
I was wondering if it would make sense to make the `.port` property of `URL` smart, so that it automatically uses the default port when no explicit port is given (and sets the port to 0 if it equals `defaultPort`). Maybe with an additional `rawPort` property that accesses the actual port written in the URL string representation.

MartinNowak:
Updated

> I was wondering if it would make sense to make the .port property of URL smart, so that it automatically uses the default port when no explicit port is given

I think `url.port ? url.port : url.defaultPort` is short enough and less complex.

MartinNowak:
Ping @s-ludwig.

s-ludwig:
LGTM!

PR #1266, opened 2015-09-17
Refactor REST client

s-ludwig:
Switches the REST client over to use `RestInterface!T`. When this is merged, I'd like to finally move all private client/server implementation parts from `vibe.web.rest` into separate modules.

This reverts the mixin template based generation of methods, because the string mixin based version got a lot shorter due to `RestInterface!T`. At the same time, the method body is now a regular template, making it *much* easier to grasp.

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -338,7 +338,7 @@ class RestInterfaceClient(I) : I
 			}
 		}
 
-		m_baseURL = url;
+		m_settings.baseURL = m_baseURL = url;
 		m_methodStyle = settings.methodStyle;
 
 		mixin (generateRestInterfaceSubInterfaceInstances!I());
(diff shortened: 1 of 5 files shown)


Geod24:
Oh yeah !
Sadly, I won't be able to review this before next week.

s-ludwig:
I'm just going to keep it rebased until then. Splitting up the module is sure tempting, but that can wait.

source/vibe/web/rest.d
```diff
@@ -296,52 +296,33 @@ unittest {
 */
 class RestInterfaceClient(I) : I
 {
-	static assert(getInterfaceValidationError!(I) is null, getInterfaceValidationError!(I));
```
Geod24: Why was it removed ? (I can't see it being added anywhere else)

Geod24: Oh right it's in `RestInterfaceClient`...

source/vibe/web/rest.d
```diff
@@ -368,17 +349,15 @@ class RestInterfaceClient(I) : I
...
+			m_subInterfaces[i].requestFilter = v;
 	}
 
-	//pragma(msg, "subinterfaces:");
-	//pragma(msg, generateRestInterfaceSubInterfaces!(I)());
-	mixin (generateRestInterfaceSubInterfaces!I());
-
 	//pragma(msg, "restinterface:");
-	mixin RestClientMethods!I;
+	mixin(generateRestClientMethods!I());
```
Geod24: TBH I was hoping to get rid of string mixins in this module. What's the  rationale to bring them back here ?

source/vibe/web/rest.d
```diff
@@ -762,336 +613,253 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
 }
 
-/// private
-private string generateRestInterfaceSubInterfaces(I)()
-{
-	if (!__ctfe)
-		assert (false);
 
-	import std.algorithm : canFind;
+string generateRestClientMethods(I)()
```
Geod24: `private`?

source/vibe/web/rest.d
```diff
@@ -762,336 +613,253 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
-					tps ~= RT.stringof;
-					string implname = RT.stringof ~ "Impl";
-
-					ret ~= q{
-						m_%s.requestFilter = m_requestFilter;
-					}.format(implname);
-					ret ~= "\n";
+	InetHeaderMap headers;
+	InetHeaderMap reqhdrs;
+	InetHeaderMap opthdrs;
```
Geod24: I'm open to ideas to clean this up :)

Geod24:
Segfault with LDC 0.15.1.
Just tested with master, it works (at least the REST example).

source/vibe/web/rest.d
s-ludwig: See my original PR message. The data available in `RestInterface!T` simply lends itself very well for this and the code got a lot shorter that way. Remember that this is just a thin layer between the `RestInterfaceClient` class and the `executeClientMethod` template function, so the drawbacks of string mixins are unlikely to appear in practice.

source/vibe/web/rest.d
s-ludwig: Oh right, I think it ended up as public because I reverted my initial attempt to move the client code to an internal module.

s-ludwig:
I had a similar segfault on LDC a few days ago. It happened in the HTTP server code, where a `TCPConnection` is implicitly cast to a `Stream` (segfault in `__d_dynamicCast` or similar). The strange thing was that the changes that triggered this were totally unrelated (adding a `Json.toString` overload). Maybe this one can be dustmited to a small test case...

source/vibe/web/rest.d
Geod24: What I'm really looking to kill is actually `generateModuleImports`. And yeah, taking a second look at the code, it looks like it's not a problem.

s-ludwig:
Anything left in terms of review? I've rebased again and will look into the LDC failure after pulling (before the next release).

source/vibe/web/rest.d
```diff
@@ -762,336 +613,253 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
-private mixin template RestClientMethods_MemberImpl(Members...) {
-	import std.traits : MemberFunctionsTuple;
-	static assert (Members.length > 0);
-	// WORKAROUND #1045 / @@BUG14375@@
-	static if (Members[0].length != 0) {
-		private alias Ovrlds = MemberFunctionsTuple!(I, Members[0]);
-		// Members can be declaration / fields.
-		static if (Ovrlds.length > 0) {
-			mixin RestClientMethods_OverloadImpl!(Ovrlds);
+	scope (exit) {
```
Geod24: Note that this may potentially throw, which is illegal according to the specs

source/vibe/web/rest.d
```diff
@@ -762,336 +613,253 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
+	if (name.length)
+	{
+		if (url.pathString.length && url.pathString[$ - 1] == '/'
+			&& name[0] == '/')
+			url.pathString = url.pathString ~ name[1 .. $];
+		else if (url.pathString.length && url.pathString[$ - 1] == '/'
+				 || name[0] == '/')
+			url.pathString = url.pathString ~ name;
+		else
+			url.pathString = url.pathString ~ '/' ~ name;
```
Geod24: I'd really like to have a way to avoid allocations here (and in other parts of the module). At work we're using buffer + memory stomping (which was the default in D1 Tango's runtime), and can be done in D2 using [assumeSafeAppend](http://dlang.org/phobos/object.html#.assumeSafeAppend)

source/vibe/web/rest.d
```diff
@@ -762,336 +613,253 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
-					foreach (pn; ParamNames) {
-						if (pn.startsWith("_") && p[1 .. $] == pn[1 .. $]) {
-							request_str ~= q{ ~ urlEncode(toRestString(serializeToJson(%s)))}.format(pn);
-							match = true;
-							break;
-						}
-					}
-				}
+		if (body_ != "")
+			req.writeBody(cast(ubyte[])body_, hdrs.get("Content-Type", "application/json"));
```
Geod24: Can't remember why it was this way to begin with, but might as well use `.length` here instead.

Geod24:
> Anything left in terms of review? I've rebased again and will look into the LDC failure after pulling (before the next release).

Surely you mean *before* pulling ? Would be good to avoid breaking master :)
Besides that, changeset looks sane, though I didn't test it, and I'll try to take another look at `executeClientMethod` tomorrow.

source/vibe/web/rest.d
s-ludwig: Since this code existed before this PR, I'm going to open a separate ticket for this.

source/vibe/web/rest.d
s-ludwig: Getting the whole thing allocation-free is something that I've already worked towards "secretly". The code is for example very close now to being able to drop the explicit `Json` values and directly `dst.serializeToJson` instead. But let's also keep that for a separate change set, right now it's important to get the structural changes in.

source/vibe/web/rest.d
s-ludwig: I think it probably came from those compile fixes due to the temporary rejection of `if (!str)` for DMD master. Yeah, but I'm kinda used to `.length`, too.

s-ludwig:
> Surely you mean before pulling ? Would be good to avoid breaking master :)

No, I'd really want to postpone this if necessary, so that other things that piled up can be done first. I currently really feel held back by this. Not breaking master is not a religion for me at this phase, especially if just for one compiler and one use case (although it's an important one....), so getting it fixed/worked around within a week or so would be OK. But I'll see first if I can find a workaround quickly.

s-ludwig:
> Besides that, changeset looks sane, though I didn't test it, and I'll try to take another look at executeClientMethod tomorrow.

Thanks, I've run it through my own code, but it would definitely be good to get some broader test coverage. BTW, we should really copy the "rest" example to the "tests" folder, it caught some bugs that the existing tests didn't. I'll quickly do that here.

s-ludwig:
Unfortunately Dustmite wasn't able to reduce the code significantly, because HTTP client and server and a TCP connection are involved. However, it seems like this failure actually existed before this PR and is just triggered by the new "rest" test, which didn't exist before. I'll simply exclude this test from the LDC test build.

s-ludwig:
Hm, so the new test actually triggered the same segfault as commit f6dfae7a7176aefc7d44d2de8afa7a9292c50c6a did for the existing "restclient" test... not a nice commit to search for workarounds.

s-ludwig:
Okay, found a workaround. The crash is somehow caused by storing a `RestInterface!I` field in `RestInterfaceClient!I`. I'll merge when Travis goes green.

PR #1267, opened 2015-09-17
make logged url clickable in terminal by adding protocol

yannick:
at least os x's terminal supports clickable urls, so with this you can directly click.


s-ludwig:
Good idea. May make sense to also think about adding clickable links with the virtual host name...

PR #1268, opened 2015-09-17
REST collections

s-ludwig:
This PR adds syntactic support for item collections in REST interfaces. It does so in a backwards compatible manner w.r.t interface definition (i.e. existing interfaces stay the same except for defining an additional item key type/name).

Open issues:
 - ~~`@path` annotations currently have to be used manually to get nice URLs (see test project)~~
 - ... don't remember, but there was something else

~~The current `ItemID` `TypeTuple` is going to be replaced by a `struct` that at the same time defines the types and names of the key parameters. The interface generator will then validate all methods to have the right parameters and will automatically generate a proper method path (`":key/method"`).~~

This PR depends on #1266

### DIFF
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -511,8 +511,9 @@ struct Collection(I)
 	import std.typetuple;
 
 	alias Interface = I;
-	alias ItemID = TypeTuple!(I.ItemID)[$-1];
-	alias ParentIDs = TypeTuple!(I.ItemID)[0 .. $-1];
+	alias AllIDs = TypeTuple!(I.ItemID);
+	alias ItemID = AllIDs[$-1];
+	alias ParentIDs = AllIDs[0 .. $-1];
 
 	private {
 		I m_interface;
@@ -537,21 +538,40 @@ struct Collection(I)
 			m_id = id;
 		}
 
-		mixin(methodProxies!I());
-
-		static private string methodProxies(I)()
-		{
+		// forward all item methods
+		mixin(() {
 			string ret;
-			foreach (m; __traits(allMembers, I))
-				ret ~= "auto "~m~"(ARGS...)(ARGS args) { return m_interface."~m~"(m_id, args); }\n";
+			foreach (m; __traits(allMembers, I)) {
+				foreach (ovrld; MemberFunctionsTuple!(I, m)) {
+					alias PT = ParameterTypeTuple!ovrld;
+					static if (is(PT[0 .. AllIDs.length] == AllIDs))
+						ret ~= "auto "~m~"(ARGS...)(ARGS args) { return m_interface."~m~"(m_id, args); }\n";
+				}
+			}
 			return ret;
-		}
+		} ());
 	}
 
 	Item opIndex(ItemID id)
 	{
 		return Item(m_interface, m_parentIDs, id);
 	}
+
+	// forward all non-item methods
+	mixin(() {
+		string ret;
+		foreach (m; __traits(allMembers, I)) {
+			foreach (ovrld; MemberFunctionsTuple!(I, m)) {
+				alias PT = ParameterTypeTuple!ovrld;
+				static if (!is(PT[0 .. AllIDs.length] == AllIDs)) {
+					static assert(is(PT[0 .. ParentIDs.length] == ParentIDs),
+						"Collection methods must take all parent IDs as the first parameters.");
+					ret ~= "auto "~m~"(ARGS...)(ARGS args) { return m_interface."~m~"(m_parentIDs, args); }\n";
+				}
+			}
+		}
+		return ret;
+	} ());
 }
 
 ///
--- a/source/vibe/web/rest.d
+++ b/source/vibe/web/rest.d
@@ -515,12 +515,39 @@ class RestInterfaceSettings {
 
 
 /**
-	Makes item collection interfaces accessible using `opIndex`.
+	Models REST collection interfaces using natural D syntax.
+
+	Use this type as the return value of a REST interface getter method/property
+	to model a collection of objects. `opIndex` is used to make the individual
+	entries accessible using the `[index]` syntax. Nested collections are
+	supported.
+
+	The interface `I` needs to define a struct named `CollectionIndices`. The
+	members of this struct denote the types and names of the indexes that lead
+	to a particular resource. If a collection is nested within another
+	collection, the order of these members must match the nesting order
+	(outermost first).
+
+	The parameter list of all of `I`'s methods must begin with all but the last
+	entry in `CollectionIndices`. Methods that also match the last entry will be
+	considered methods of a collection item (`collection[index].method()`),
+	wheres all other methods will be considered methods of the collection
+	itself (`collection.method()`).
+
+	The name of the index parameters affects the default path of a method's
+	route. Normal parameter names will be subject to the same rules as usual
+	routes (see `registerRestInterface`) and will be mapped to query or form
+	parameters at the protocol level. Names starting with an underscore will
+	instead be mapped to path placeholders. For example,
+	`void getName(int __item_id)` will be mapped to a GET request to the
+	path `":item_id/name"`.
 */
 struct Collection(I)
 	if (is(I == interface))
 {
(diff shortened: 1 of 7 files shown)


source/vibe/web/rest.d
```diff
@@ -618,6 +489,144 @@ class RestInterfaceSettings {
...
 /**
+	Makes item collection interfaces accessible using `opIndex`.
+*/
+struct Collection(I)
+	if (is(I == interface))
+{
+	import std.typetuple;
+
+	alias Interface = I;
+	alias AllIDs = TypeTuple!(typeof(I.CollectionIndices.tupleof));
```
Geod24: Rather than convention, we could go with UDA

Geod24:
Sure looks interesting from a quick glance (and is definitely needed).
I'll do a full review once #1266 is in.

s-ludwig:
I don't know why I didn't have this idea earlier, but I got stuck with an approach where there would be an auto-generated class that sits between the original API interface and the implementation. This didn't really work out well due to performance and clarity issues. This approach with a `Collection` type on the other hand is really quite straight forward and is a much easier transition target for existing code.

source/vibe/web/rest.d
s-ludwig: I thought about that, too, but decided against it, because it would create more code. But it *is* clearer that a particular interface acts as a collection with an UDA, so I'm still on the edge.

tests/restcollections/source/app.d
```diff
@@ -0,0 +1,110 @@
...
+
+	auto api = new RestInterfaceClient!API("http://127.0.0.1:8000/");
+	assert(api.items["foo"].subItems.length == 2);
+	assert(api.items["foo"].subItems[0].name == "hello");
+	assert(api.items["foo"].subItems[1].name == "world");
+	assert(api.items.manager.databaseURL == "in-memory://");
+	exitEventLoop(true);
+}
+
+int main()
```
Geod24: Why the custom main ?

Geod24:
I'm concerned by the lack of example / documentation.
Collections are important, if not central, to REST apis, and this change leave it almost undocumented.

tests/restcollections/source/app.d
s-ludwig: Not really important. It was added to the "restclient" test long ago and I used that as a template.

s-ludwig:
I've added some more thorough documentation for `Collection!I`. There should probably also be an example in the examples/ folder, but the current "rest" example seems already a bit too crowded. Maybe I'll just take the one from "tests/rest-collections" and adjust it a little.

s-ludwig:
I think I remember the issue that I mentioned in the PR description. Property functions currently don't map well to this, because they have one or more parameters when they really should have none:

``` 
interface Item {
    struct CollectionIndices { int _id; }
    @property string name(int _id); // OOPS: this is now actually a setter property...
}
```

So this currently either means that you'd write `string getName(int _id);`, or that you'd do `@method(HTTPMethod.GET) string name(int _id)`.

A possible way to solve this would be to automatically convert `getXXX(...)` in the interface to `@property xxx()` in `Collection!I`. Or we could introduce another UDA à la `@collectionProperty string name(...);` to make this conversion explicit. Any other ideas?

s-ludwig:
Another point would be if we should have a shortcut syntax for collections that contain plain values:
``` 
interface Values {
    struct CollectionIndices { int index; }
    float getValue(int index);
}

auto val = values[0].get(); // current syntax
auto val = values[0]; // possible shortcut syntax

// same could be done for set() and opIndexAssign()
```

s-ludwig:
Example is there now + made the index parameter check more rigid + rebased.

Since both, the `@property` issue and possible shortcut syntaxes, can be implemented in a backwards-compatible manner, I'd exclude those from this PR, if there are no objections.

Geod24:
As we discussed, I tried it with a FS abstraction. It's still a WIP but I already have a bit of feedback.

I obviously forgot CollectionIndices on the first run, and got the following error:

```
../vibe.d/source/vibe/web/rest.d(532,35): Error: no property 'CollectionIndices' for type 'app.IFSItem'
../vibe.d/source/vibe/web/rest.d(534,22): Error: no property 'CollectionIndices' for type 'app.IFSItem'
../vibe.d/source/vibe/web/rest.d(540,2): Error: can only slice tuple types, not _error_
../vibe.d/source/vibe/web/rest.d(541,2): Error: can only slice tuple types, not _error_
../vibe.d/source/vibe/web/rest.d(624,6): Error: static assert  __error
source/app.d(41,12):        instantiated from here: Collection!(IFSItem)
```

Oups.

It seems that there is no support for listing objects in the collection ? Also, how do you plan to support collections that holds different types of object. z.b. an "animals" collection which hold a turtle and a dolphin ? In regular D you could just cast, but here you'll get null every time.

s-ludwig:
> It seems that there is no support for listing objects in the collection ?

You can define a `get` method that takes all but the last index as parameters. The result will be available as `GET /animals/` or similar. Maybe we could have an automated alternative, but that sounds like it could get hairy in terms of required flexibility/complexity vs. benefits.

> Also, how do you plan to support collections that holds different types of object. z.b. an "animals" collection which hold a turtle and a dolphin ? In regular D you could just cast, but here you'll get null every time.

That would have to be done using a "super"-interface that combines all methods and reacts with a 404 if a method is called on something that doesn't match the derived type. Do you have an idea how to improve this? We could define some kind of automated helper to at least make the interface definition part easier (`interface SuperInterface(Interface, DerivedInterfaces...)`, but I'm not sure if there is a reasonable way to go further, as that would require leaking the internal representation of the collection AFAICS.

Geod24:
> You can define a get method that takes all but the last index as parameters. The result will be available as GET /animals/ or similar. Maybe we could have an automated alternative, but that sounds like it could get hairy in terms of required flexibility/complexity vs. benefits.

Could you add it to the examples ? Being able to list ressources in a collection (and to customize the listing) is mandatory for collections (else it isn't delf-describing).

> That would have to be done using a "super"-interface that combines all methods and reacts with a 404 if a method is called on something that doesn't match the derived type.

That could be a good thing. We could also couple that with `opCast` maybe ?

s-ludwig:
Added an example project (including the definition of a `get()` member).

Geod24:
What seems *very* weird to me in it's usage is that `GET /threads/` returns a list of indices, and not a list of object.

Example here: https://api.github.com/users/s-ludwig/repos

In this example, it is even worst because you cannot do  `https://api.github.com/users/s-ludwig/repos/:id`. You're supposed to list repository this way and then use links from the object to access it's property (I even think this serves a "short representation" of the object, but I'm not 100% sure).

So to list issues in a repository you would access https://api.github.com/repos/s-ludwig/FrameworkBenchmarks/issues .

Would that be representable with the current P.R. ?

s-ludwig:
You can return anything you want from the `get()` method. Returning IDs is just the possibility that seemed to make most sense in the example scenario. Returning URLs would have been another alternative, but that would just have wasted a lot of bandwidth. Similarly, returning the full thread objects could mean a huge amount of data in a real-world scenario.

Another common approach is to return a list of partial resources (e.g. thread name, date, poster and contents of the first post).

s-ludwig:
Regarding recursive collections, I've thought about them and support can definitely be added on top later.  However, it requires a more elaborate implementation, because the router is of no help here (can't define routes with recursive patterns). But for the specific case of a file system structure it would suffice to simply make the "*"-part of a route accessible as a parameter (and adjust the client generator accordingly). I think we should definitely do that, regardless of collections.

Regarding this PR, I'd want to merge this now. We have a full release period ahead to make any required changes, or even to remove it if necessary. But having it in master gives it a lot more exposure for some actual real-world testing.

s-ludwig:
>> That would have to be done using a "super"-interface that combines all methods and reacts with a 404 if a method is called on something that doesn't match the derived type.

> That could be a good thing. We could also couple that with `opCast` maybe ?

The problem with `opCast` is that it would require sending type information over the wire, either directly when requesting a resource, or using a special HTTP request that happens within `opCast`. This sounds like something that requires more thought to avoid

PR #1269, opened 2015-09-17
Allow rest apis to use the raw HTTPServerRequest and HTTPServerResponse objects

mfrischknecht:
I was looking for the possibility to serve static files through a rest interface, but so far the only matching discussion I could find was located here: http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/2629/

I'd propose to allow api methods to access the underlying HTTPServerRequest and HTTPServerResponse objects used by the server (implemented by @rawRequest and @rawResponse). To avoid errors in the response, the special clause for handlers who return void should check whether the response has already been altered before sending empty JSON objects to the client.
The changes in examples/rest introduce a new example for this kind of method.

Note: On my machine, example 4 keeps failing for some reason; that's why I introduced the try-catch clauses to run all examples even if one breaks in the process.

Also, I'm just starting out with vibe (and D in general), so I would welcome any kind of feedback/coding advice :-)

### DIFF
--- a/source/vibe/core/stream.d
+++ b/source/vibe/core/stream.d
@@ -145,7 +145,7 @@ interface Stream : InputStream, OutputStream {
 	Connection streams are based on streaming socket connections, pipes and
 	similar end-to-end streams.
 
-	See_also: vibe.core.new.TCPConnection
+	See_also: vibe.core.net.TCPConnection
 */
 interface ConnectionStream : Stream {
 	/** Determines The current connection status.


s-ludwig:
If we should do this for the REST interface generator, I'd remove the need for `@rawRequest`/`@rawResponse` UDAs to make it the same as for `registerWebInterface`, which directly recognizes the parameter type (it wouldn't make sense/work to serialize such an object as a normal parameter anyway). However, this goes against the philosophy of this module, which is that `RestInterfaceClient` should generally be able to mirror all methods of the server and that the server `interface` should be usable locally without going through HTTP.

For this specific case, I could imagine that it would make sense to define a `BinaryResource` struct in `vibe.web.common` and let the REST interface generator handle the details of handling last changed times and etags:
```
struct BinaryResource {
    InputStream data;
    string fileName; // optional
    SysTime lastModified;
    string etag;
    Duration maxCache;
}
```

But generally, the UDA approach is also interesting and should at least be mentioned somewhere so that people now how to work around this limitation (maybe in a new FAQ section, @Laeeth has already prepared an initial list of questions: http://wiki.dlang.org/Vibed_faq_addition).

mfrischknecht:
Thank you for your input! If you want to always enforce the "mirrorability", I guess this PR can be declined.

However, I still see some drawbacks of the BinaryResource-approach in some cases - but I might also have an idea of how to generalize your approach.

Serving static files through the HttpServerResponse (using serveStaticFiles) provides some useful properties:
1. I don't have to read the whole file into memory (useful for serving bigger files).
2. The API's method won't exit before the reading is done, allowing custom synchronization in the API. Since the file is accessed through the REST API, it is very likely that it will also be changed this way.

As far as I understand the solution using BinaryResource, it would prevent either 1.) or 2.) (i.e. you don't get locking OR you have to read the whole file into memory). I guess this should be worth considering for any RestInterface dealing with files.

However, after having read the discussion on XML documents in [this thread](http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/26609/), I thought if maybe an attribute-based solution like this might be helpful:

```d
class MyCustomSerializer
{
    ubyte[] serialize(T value) { /* serialize the value */ }
    T deserialize(ubyte[] serialized) { /* de-serialize the data */ }
}

interface IMyRestInterface
{
    @serialized!(new MyCustomSerializer())
    T getSomethingBinary();
}
```

This would allow any well-defined format as a drop-in replacement for JSON: On the server side, the result returned by `getSomethingBinary()` is passed to `MyCustomSerializer.serialize()`, and the result is transmitted to the client. On the client, the data gets de-serialized again and passed to the application.
I think this idea might even be useful for the above features if the Serializer is extended accordingly:

```d
class MyCustomHTTPSerializer
{
    ubyte[] serialize(T value, HTTPServerRequest request, HTTPServerResponse response) 
    { /* write the serialized value directly to the response object */ }
    T deserialize(ref HTTPClientRequest request, HTTPClientResponse response) 
    { /* de-serialize the data from the response */ }
}

interface IMyRestInterface2
{
    @serialized!(new MyCustomHTTPSerializer())
    T getSomethingBinary();
}
```

This could even be combined with the BinaryResource approach (say you use it for convenience and slap a custom serializer onto the method if you need locking).

PR #1271, opened 2015-09-17
add log output for virtualhosts.

yannick:
not sure if thats what you referred to @s-ludwig .
currently they are printed for each address family which makes them seem duplicates.
so it should also output the address family, however i was unsure on how to retrieve that.


s-ludwig:
Yes, looks about like what I imagined. It would just have to happen at a different place for it to be always useful/correct. I'll have a second look later, as I don't remember the code details now.

yannick:
ok, i can fix it up once you give me a hint on what needs to be done

s-ludwig:
Finally got around to look into this. It was rather trivial to add, so I directly committed the change instead of wasting your time. Thanks anyway for pursuing this!

PR #1280, opened 2015-10-01
Disable SSL3 for TLSVersion.any and upgrade enum with TLS 1.1 and TLS 1.2

marcioapm:
What we had discussed in the forum thread. Hopefully it can still make 0.7.26


s-ludwig:
Will definitely make it!

PR #1281, opened 2015-10-01
Allow per request disabling of write to access log

marcioapm:
Not sure if this is the best place for this, but it's the most convenient at least :)

My use case is to disable some monitoring-only requests, which I don't want to pollute the access.log with. In my case, these are just some internal web-services.


s-ludwig:
Looks good to me.

PR #1282, opened 2015-10-03
infer nothrow synchronized support

MartinNowak:


MartinNowak:
So we avoid breaking vibe.d with every new version.

MartinNowak:
I'll eventually get to implement the nothrow synchronized, but it's not my top priority atm.

s-ludwig:
Indeed makes sense, thanks!

PR #1284, opened 2015-10-04
Split library up into sub packages.

s-ludwig:
This introduces separate sub modules to enable smaller compilation footprints for applications that use only a part of vibe.d's functionality.


s-ludwig:
I'm a bit puzzled by the "Aborting: Cycle detected between modules with ctors/dtors: vibe.http.server -> vbie.http.session -> vibe.http.server" error. Commit 1248003b9327e57eebfd98498906ac7c6dfe5b16 removes the static constructor from `vibe.http.session`, but it still fails.

Geod24:
Is this P.R. still necessary ?

s-ludwig:
It now just contains the (possible) second step of the split - separating the sources into different source directories. I'm not completely sure if this is a good idea at this point.

wilzbach:
this is just a short ping on this issue. What is blocking to move over to the "vibed" or "vibe-d" organization and split optional packages like mongo et. al into optional github repos?

s-ludwig:
I want to take the opportunity to make a clean cut while doing that. I'm currently (more or less) busy with the core module, and once that is done, I'll work on making the transition as seamless as possible. The modules in the vibe-d organization will then each start at 1.0.0 and will be versioned independently of each other.

PR #1287, opened 2015-10-08
Add a fiber-affine read write mutex implementation

mfrischknecht:
cf. [this forum thread](http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/26569/).

I figured it would be easier to discuss the last few changes to the code in a pull request itself rather than the forum.


source/vibe/core/sync.d
```diff
@@ -819,3 +819,490 @@ private struct TaskConditionImpl(bool INTERRUPTIBLE, LOCKABLE) {
...
+    /** The policy with which the lock has been created. */
+    @property Policy policy() const { return m_state.policy; }
+}
+
+/** A simple alias for $(D TaskReadWriteMutexImpl).
+ *
+ *  Use this alias to avoid providing an empty template parameter list when
+ *  instantiating a mutex.
+ */
+alias TaskReadWriteMutex = TaskReadWriteMutexImpl!();
```
mfrischknecht: As noted, adding empty template parameter lists does indeed allow the compiler to infer shared methods again. The downside is, that for this to work, each mutex would have to be initialized using an explicit empty parameter list, e.g. `TaskReadWriteMutex!()`.

To mitigate this, i decided to name the classes differently (`TaskReadWriteMutexImpl!()`) but leave them public to guarantee useful documentation. The mutex type itself can be referred to through an alias (i.e. `TaskReadWriteMutex x();` still works).

source/vibe/core/sync.d
```diff
@@ -819,3 +819,490 @@ private struct TaskConditionImpl(bool INTERRUPTIBLE, LOCKABLE) {
...
+    /** A common baseclass for both of the provided mutexes.
+     *
+     *  The intent for the according mutex is specified through the 
+     *  $(D INTENT) template argument, which determines if a mutex is 
+     *  used for read or write locking.
+     */
+    final class Mutex(LockingIntent INTENT): core.sync.mutex.Mutex, Lockable
+    {
+        /** Try to lock the mutex. cf. $(D core.sync.mutex.Mutex) */
+        override bool tryLock() { return m_state.tryLock!INTENT(); }
```
mfrischknecht: Using a `std.core.sync.mutex.Mutex` (as suggested) leads to the non-interruptible Mutex implementation to also not being `nothrow` anymore.
It doesn't seem to be a real downside though, as I can't really see any sense in using a `TaskReadWriteMutex` as an object monitor anyway...

s-ludwig: That plus that the initial reason for building the uninterruptible mutex implementations was that the one from `core.sync.Mutex` was supposed to be made `nothrow`, but it broke the ones in vibe.d that way. It was postponed for that reason, but the switch is planned to be made soon, now that the vibe.d classes have been made forward-compatible.

mfrischknecht: In that case, I'd propose to delay the `nothrow` until the switch is made, and re-add the attribute in `TaskReadWriteMutex` afterwards (thanks to inference, only the definitions of `TaskReadWriteMutex.Mutex.tryLock()`, as well as `lock()` and `unlock()` should be affected).

source/vibe/core/sync.d
s-ludwig: Ah okay, I didn't expect that. Indeed it sounds like the best first step then to keep it as a simple non-`shared` class for now. That way the functionality can get in today, and the decision on how to add support for `shared` can be made in a broader context when the whole picture becomes clearer. But we could of course also just add `shared` overloads of the member functions for example.

examples/read_write_mutex/source/app.d
```diff
@@ -0,0 +1,88 @@
...
+					auto api = new RestInterfaceClient!MyBlockingRestInterface("http://127.0.0.1:8080");
+
+					for (int i = 0; i < 100000; ++i)
+						api.getIndex();
+
+					//Since we can't join threads in a setTimer call,
+					//exit the event loop after the first thread has complete all
+					//of it's requests. In any other code, this is propably a bad idea.
+					scope(exit)
+						exitEventLoop(true);
```
mfrischknecht: I've added a modified version of my test application as an example. It is rather simplistic and uses the `vibe.core.concurrency.async` so far, but somehow I'm not able to join the threads correctly, leading to this hack:
Really, the `scope(exit)` part should be put after the outer loop, but I'm running into an error message when calling `Future!T.getResult()`:
```
Task terminated with unhandled exception:
Joining tasks in foreign threads is currently not supported.
```
Do you by any chance have an idea for a better implementation?
On a related note: Is it safe to call `RestInterfaceClient` methods from multiple ordinary D threads concurrently?

source/vibe/core/sync.d
```diff
@@ -819,3 +819,469 @@ private struct TaskConditionImpl(bool INTERRUPTIBLE, LOCKABLE) {
...
+ *      using $(D vibe.core.task.Task.interrupt()). The corresponding
+ *      InterruptException will be deferred until the next blocking
+ *      operation yields the event loop.
+ * 
+ *      Use $(D InterruptibleTaskReadWriteMutex) as an alternative that can be
+ *      interrupted.
+ * 
+ *  cf. $(D core.sync.mutex.ReadWriteMutex)
+ */
+class TaskReadWriteMutex
```
mfrischknecht: As I already noted, I had similar issues creating a global shared `TaskMutex` instance in the same fashion as my previous tests for the read-write version; so I decided to just revert to the basic non-template version for the two mutex classes.

If the missing shared overloads are an issue, I'd propose to solve that in a separate PR for all mutex classes in one go.

examples/read_write_mutex/source/app.d
s-ludwig: This is still a really unfortunate limitation currently, I really need to bite the bullet and implement that. A clean workaround would be to use a `TaskCondition` that gets notified at the end of the `async` callback. Probably `Future` should in fact also just use that internally until `Task.join()` is supported across threads.

`RestInterfaceClient`, if the `requestFilter` property isn't modified after initialization, should be thread-safe in practice, because all fields are *de-facto* immutable. I wouldn't do that in an example, though, before this use case is actually supported by the type system. It would be an interesting addition to mirror all methods as `shared` and to add a `shared` constructor. On the other hand, constructing a new `RestInterfaceClient` is cheap and doing it in a module constructor once per thread would eliminate the overhead anyway.

source/vibe/core/sync.d
s-ludwig: Sounds good!

examples/read_write_mutex/source/app.d
mfrischknecht: > On the other hand, constructing a new RestInterfaceClient is cheap and doing it in a module constructor once per thread would eliminate the overhead anyway.

So I assume I won't have to worry over some implicit dependencies on a shared event-loop then? In that case, it looks like D's `core.thread` should already suffice here.

s-ludwig: Yes, the actual HTTP/network operations are done using a connection pool that is thread-local, so that it will always use the event loop of the current thread. You could use `core.thread`, or you could also simply use `runWorkerTaskDist`, which will create one identical worker task in each of the thread pool worker threads.

examples/read_write_mutex/source/app.d
```diff
@@ -0,0 +1,88 @@
...
+		TaskReadWriteMutex m_mutex;
+	}
+
+	this()
+	{
+		m_mutex = new TaskReadWriteMutex(TaskReadWriteMutex.Policy.PREFER_WRITERS);
+	}
+
+	//This method naively simulates access to a shared resource protected
+	//by a TaskReadWriteMutex. Read operations are performed recently (95%)
```
Geod24: You mean performed `frequently` ?

source/vibe/core/sync.d
```diff
@@ -819,3 +819,469 @@ private struct TaskConditionImpl(bool INTERRUPTIBLE, LOCKABLE) {
...
+        }
+    }
+    
+    /** Unlock the mutex after a successful acquisition. */
+    @trusted void unlock(LockingIntent INTENT)()
+    {
+        version(RWMutexPrint)
+            printInfo!("unlock",INTENT)();
+        
+        debug assert(activeCounter!INTENT > 0);
```
Geod24: `debug assert` is redundant, just use `assert`.

s-ludwig: Not completely true, `assert` will be there when neither `-debug`, nor `-release` is used, but `debug` statements are only included if `-debug` is given. I usually use `debug assert` in performance critical places, so that it's possible to build a fast "release" build without resorting to the unsafe `-release` switch.

Geod24: Hum right, I forgot that defaults builds are not actually `debug` builds ;)
It looks like here the extra check won't hurt.

examples/read_write_mutex/source/app.d
mfrischknecht: Of course. My Bad.
Thanks for the correction :)

examples/read_write_mutex/source/app.d
```diff
@@ -0,0 +1,89 @@
...
+                auto api = new RestInterfaceClient!MyBlockingRestInterface("http://127.0.0.1:8080");
+                for (int i = 0; i < 1000; ++i)
+                    api.getIndex();
+
+            }, runningTasks);
+         
+        //Join all worker tasks. Currently, it's not possible to join tasks from non-vibe threads.
+        do 
+            sleep(3.seconds);
+        while(atomicLoad(*runningTasks) > 0);
```
mfrischknecht: I tried the threading methods we discussed but really had no luck so far, so I resorted to a very "poor man's join" using a `shared(int)*` and `atomicOp` for the time being.

If I understand the `TaskCondition` class right, using it would just introduce another race, because I would have to wait for multiple conditions which can be triggered in any order (and I have no possibility of extracting the `emitCount` of the underlying `ManualEvent` object)<sup>1</sup>.
Also, I had problems passing any non-shared objects to tasks by argument. This means that using e.g. `ManualEvent` objects would result in some really ugly casts to and from `shared` (since the according construction function and methods have no shared overloads).
Using D's `core.thread` on the other hand seems to work, but results in the example not being able to shut down  properly after receiving SIGINT (Ctrl+C), which is also rather annoying.

<sup>1</sup> I could imagine that this would be an interesting use-case, since waiting for the first task in a list to complete really isn't a very obscure idea. I would even assume that this would be necessary to use `TaskCondition` to properly wait for a `Future`.
As a sidenote: If `Future` would be changed as to use a `TaskCondition`, perhaps it would also be nice to be able to use `async()` with callbacks that return `void`, simply for the joining functionality...

s-ludwig: > If I understand the TaskCondition class right, using it would just introduce another race, because I would have to wait for multiple conditions which can be triggered in any order (and I have no possibility of extracting the emitCount of the underlying ManualEvent object)1.

Usually in this case you'd just use a single `TaskCondition` and use something like the counter above. Using a global one should be OK for the sake of this example:
```
__gshared {
    Mutex s_taskMutex,
    TaskCondition s_taskCondition;
    size_t s_taskCounter = 0;
}

{
    synchronized (s_taskMutex) s_taskCounter = workerThreadCount();

    runWorkerTaskDist(function() {
        scope (exit) {
            synchronized (s_taskMuted) s_taskCounter--;
            s_taskCondition.notifyAll();
        }
        // ....
    });

    synchronized (s_taskMutex)
        while (s_taskCounter > 0)
            s_taskCondition.wait();
}
```

`ManualEvent` is of course an obvious candidate for `shared` and the fact that it isn't is just because I'm still reluctant to commit to a specific `shared` approach until it's clear that this will stay valid in the long term (The approach that I implemented with `vibe.core.concurrency.lock` for example is valid according to ideas that Walter Bright once mentioned, but others, such as Andrei Alexandrescu, objected). But I still think it's better to have the `runWorkerTask` interface safe w.r.t. `shared` and require casts in some situations than making it completely unsafe.

mfrischknecht: Forgive me if I'm overlooking something; I'll try to explain which specific case I'm afraid of:

```{D}
synchronized (s_taskMutex)
        while (s_taskCounter > 0)
            s_taskCondition.wait();
```
This code basically comes down to check -> wait -> check -> wait etc.
The way `TaskCondition.wait()` is currently implemented, there seems (to me at least) to be a possibility, that the state of the condition changes *between* **check** and **wait** in a concurrent setup. This is the problem I also see if `Future` would be implemented with a `TaskCondition`.
If the check leads to a waiting operation, but the notify() has already propagated before the method is executed, `TaskCondition.wait()` will internally fetch the **new** emitCount and produce a deadlock.

Whereas with `ManualEvent`, one could write:
```{D}
auto count = s_event.emitCount;
while (atomicLoad(s_taskCounter) > 0) //Atomic since we don't use locking here
    count = s_event.wait(count);
```
`ManualEvent.emitCount()` seems to perform an `atomicLoad()` operation and the resulting memory fences should guarantee that the count is always extracted before the check is performed. This should make deadlocks impossible.

Again, forgive me if I'm nitpicking or misunderstanding something, but I have personally already been burned by example code that produced deadlocks with some very small propability. Because of that, I think it is important to promote *correct* code even if it might not be absolutely necessary for a simple example, as those kinds of bugs are a pain to hunt down in more complex systems.


> I'm still reluctant to commit to a specific shared approach until it's clear that this will stay valid in the long term

I completely understand that, the whole matter seems rather perplexing to me.

s-ludwig: Hm, but `TaskCondition` internally uses `emitCount` to avoid that particular issue. It records the emit count before unlocking the mutex, then unlocks and then calls `ManualEvent.wait` with that old count, so that it will immediately exit and not wait when a notification has happened in-between.

mfrischknecht:
Actually, you're right. Seems I haven't grasped the whole idea yet - I always assumed a case where the condition would be triggered without holding the mutex (i.e. from outside of a `synchronized()` block). I'll implement your pattern then.

examples/read_write_mutex/source/app.d
mfrischknecht: Actually, you're right. Seems I haven't grasped the whole idea yet - I always assumed a case where the condition would be triggered without holding the mutex (i.e. from outside of a synchronized() block). I'll implement your pattern then.

s-ludwig: Or, more specific, the *condition* has to be updated within the mutex lock. Then it's also safe to trigger the condition afterwards outside of the lock.

Yeah, definitely makes sense to add this as an example. I'll do that.

examples/read_write_mutex/source/app.d
```diff
@@ -0,0 +1,108 @@
...
+                for (int i = 0; i < 1000; ++i)
+                    api.getIndex();
+
+            });
+         
+        //Wait for all tasks to complete
+        synchronized(s_taskMutex) {
+            do s_taskCondition.wait();
+            while(s_runningTasks > 0);
+        }
```
mfrischknecht: The example now uses a `TaskMutex` for joining. Thanks for your advice and patience!

s-ludwig: Please no worries about patience, after all this has brought a couple of important issues to light and it's never wrong to take a detailed look at these places! I'll do a final pass over the changes.

s-ludwig: Theoretically, in the current version, the worker tasks could have finished before the wait loop at the bottom is reached, thus causing an infinite wait on `s_taskCondition`. This will of course never happen in this example, ~~but maybe it's better to just initialize `s_runningTasks` with `workerThreadCount` before starting up the tasks.~~ Damnit, lazy worker thread construction looks like it gets in the way of this and `workerThreadCount` would just return `0`.

I'm leaning towards letting `workerThreadCount()` simply call `setupWorkerThreads()`, as `0` is certainly not the expected answer is most practical situations.

mfrischknecht: In the meantime, to assure the correctness of this example, I could just extend the `synchronized` block so that it includes the `runWorkerTaskDist()` call. It's perhaps not the most readable, but would guarantee that the mutex is locked until `s_taskCondition.wait();` is performed. This means that the started tasks would block on the increment until this is the case.
The only assumption left would then be that `runWorkerTaskDist()` does start >0 threads (which seems sensible ;-) ).

s-ludwig: I've commited the change to `workerThreadCount` now and also updated the docs to mention that `runWorkerTaskDist` will start exactly `workerThreadCount` tasks, and `workerThreadCount` also has an out contract that it will always be `> 0`.

Extending the `synchronized` block is also an interesting solution, but I fear that most people looking at the example will have trouble to understand the implications. With the `workerThreadCount` change done, my vote would go for setting up the counter up-front.

examples/read_write_mutex/source/app.d
```diff
@@ -0,0 +1,103 @@
...
+
+	//Wait for a couple of seconds for the server to be initialized properly and then start
+	//multiple concurrent threads that simultaneously start queries on the rest interface defined above.
+	setTimer(5.seconds, {
+
+		scope(exit)
+			exitEventLoop(true);
+
+        //Start multiple tasks performing requests on "http://localhost:8080/" concurrently
+        synchronized(s_taskMutex) s_runningTasks = workerThreadCount();
```
mfrischknecht: I added the counter initialization with `workerThreadCount()` and did a rebase to the upstream master branch. Interestingly, `workerThreadCount()` even worked before the rebase (at least on my system).
I'm quite sure the `synchronized(s_taskMutex)` is not technically necessary, but since this line sets up the condition variable, I figured it doesn't hurt to have it there.

s-ludwig: Thanks! I'll merge now.

Yeah, I'd consider the `synchronized` display of good practices for the sake of a generic example.

PR #1290, opened 2015-10-10
Feature/i18n enhancement

s-ludwig:
This is an up-to-date version of #966.


s-ludwig:
I've reverted the variable renames and some other small things to make the diff smaller and better suited for a review. Also, if no translation is found, the incoming key is now selected based on the following logic instead of always returning the singular form: `!ptxt.length || cnt == 1 ? txt : ptxt;` (where `ptxt` is the plural form key).

The empty "msgid" entry is also still written as an entry, so that headers other than those necessary for plural form handling can also be processed by external code (and this makes an existing unit test pass).

ncoe:
Works for me, although I still don't know what the integration server was choking on. Thanks for taking a look.

PR #1293, opened 2015-10-14
Use ReturnType!FT in JS Client Generator instead of undefined RT

ColdenCullen:
This fixes the `on_result` code being inserted when it isn't required.


eco:
Thanks for this. Will this fix make it into 0.7.26?

s-ludwig:
Definitely!

PR #1295, opened 2015-10-18
Expand on driver's documentation

Geod24:
As I'm going through this part again, I figured it would be a good day to expand on the documentation.
I'm wondering, what should happen with the libev driver and the winrt driver ? The later looks more like a stub, is it actually usable on winRT ? Do you have anything in mind for this part of the code ?

### DIFF
--- a/source/vibe/core/driver.d
+++ b/source/vibe/core/driver.d
@@ -1,5 +1,11 @@
 /**
-	Contains interfaces and enums for evented I/O drivers.
+	Contains interfaces and enums for asynchronous drivers.
+
+	At the lowest level of Vibe.d sits a library which handle all the
+	asynchronous I/O operations.
+	There are currently 3 supported I/O backend: libasync, libevent and libev.
+	This module define the interface such a library must conform with
+	to work with Vibe.d
 
 	Copyright: © 2012-2015 RejectedSoftware e.K.
 	Authors: Sönke Ludwig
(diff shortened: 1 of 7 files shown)


s-ludwig:
I did some experiments with WinRT and a full automatically generated set of D bindings, but didn't have time to actually put it to use, so most of this is indeed just a stub. It also requires the bindings and some build tweaks, so it's not usable like it is. Nowadays it actually becomes interesting again, because we have `-m32mscoff` now, which is needed for Windows apps (because you need to use their linker with special flags). The long-term goal would surely be to put this into libasync or a possible adaption for Phobos, but for now it should probably just there with a note that it's just a stub.

The libev driver has just received some improvements (timers and manual events now fully work) which makes it actually usable for TCP server kind of applications. I don't actually use it for something, but I'd also like to keep it as a comparison with the other divers.

Geod24:
> The long-term goal would surely be to put this into libasync or a possible adaption for Phobos, but for now it should probably just there with a note that it's just a stub.

Or we could just remove it. Since it's probably not going to be implemented any time source, and with revision control you would be able to find the code anyway.

s-ludwig:
I don't know. I like that it gets CI tested at at least stays up to date. It doesn't really hurt as long as it's properly documented what to expect.

Geod24:
Ping ?

Geod24:
Ping ^2 ?

Geod24:
Pingitty-ping, this is quite trivial. If there is any wording you dislike let me know and I'll correct it.

s-ludwig:
Quite embarrassing that this slipped through the radar despite three "pings", sorry for that! I've cherry picked this now and resolved the conflicts.

PR #1299, opened 2015-10-20
Adding Cors to RestInterface

skoppe:
This pull request adds cors support for rest.

- `getRoutesGroupedByPattern()` to RestInterface struct
- add unittest for `getRoutesGroupedByPattern()`
- extends `jsonMethodHandler` signature to take `RestInterfaceSettings`
- add code to `jsonMethodHandler` to handle cors request
- add code to generate OPTIONS routes in `registerRestInterface` to handle preflight cors requests
- add test code in tests/restclient


skoppe:
need to replace the `asLowerCase` into dmd 2.066 equivalent

skoppe:
hmm, apparently chunkBy and the way I use tuple isn't valid on 2.066..

s-ludwig:
Seems like `groupBy` got renamed to `chunkBy`, but at the same time `SortedRange.groupBy` was added: https://issues.dlang.org/show_bug.cgi?id=14183

s-ludwig:
And `tuple` didn't yet have support for named fields at all... always fun to be forced to support a range of compiler versions.

skoppe:
Didn't know about the groupBy, that should make it easier. Will do after lunch.

source/vibe/web/rest.d
```diff
@@ -618,6 +660,88 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
+ *	A delegate suitable to use as an handler for an HTTP request.
+ */
+private HTTPServerRequestDelegate optionsMethodHandler(RouteRange)(RouteRange routes, RestInterfaceSettings settings = null)
+{
+	import vibe.http.server : HTTPServerRequest, HTTPServerResponse;
+	import std.algorithm : map, joiner, any;
+	import std.conv : text;
+	import std.array : array;
+	import vibe.http.common : httpMethodString, httpMethodFromString;
+	// NOTE: don't know what is better, to keep this in memory, or generate on each request
```
s-ludwig: Should be good as it is now. In case of such small allocations, I'd always opt to reduce the per-request GC pressure (even if there remains a lot to do for the REST module in that regard).

source/vibe/web/rest.d
```diff
@@ -618,6 +660,88 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
+		auto origin = "Origin" in req.headers;
+		if (origin is null)
+			return;
+
+		import std.algorithm : equal;
+		import std.uni : toLower;
+		auto originLowerCased = (*origin).toLower;
+		if (settings !is null &&
+			settings.allowedOrigins.length != 0 &&
+			!settings.allowedOrigins.any!(org => org.toLower.equal(originLowerCased)))
```
s-ludwig: To avoid the two allocations, instead of toLower, [sicmp](http://dlang.org/library/std/uni/sicmp.html) would be preferable here. BTW, shouldn't even `std.ascii` be sufficient in theory?

source/vibe/web/rest.d
```diff
@@ -618,6 +660,88 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
+			return;
+
+		res.headers["Access-Control-Allow-Origin"] = *origin;
+
+		// there is no way to know if the specific resource supports credentials
+		// (either cookies, HTTP authentication, or client-side SSL certificates), 
+		// so we always assume it does
+		res.headers["Access-Control-Allow-Credentials"] = "true";
+		res.headers["Access-Control-Max-Age"] = "1728000";
+		res.headers["Access-Control-Allow-Methods"] = *method;
```
s-ludwig: Looks like a good argument for starting to design a built-in generic authentication mechanism. Being able to remove `@before` at some point is another important argument, because that would also enable getting a full list of headers.

source/vibe/web/rest.d
skoppe: I knew there was some a case insensitive cmp somewhere.

Although the DNS does allow binary strings, in practice it is limited to ascii. So yeah, ascii is sufficient.

source/vibe/web/rest.d
skoppe: Its funny, elsewhere I was complaining that you don't get access to the raw request/response headers. Now it is obvious you don't want that. In fact, any dealings with low level http details should be know at compile time (at least with the rest interface)

Having said that, its going to take a long time before anybody complains about allow-credentials being non-configurable. Maybe max-age...

source/vibe/web/rest.d
skoppe: Yeah that makes sense. In principle I could get the array of methods at compile time, but that entailed more work.

About the GC, I suppose it would be possible to do some static analysis and see which lines of code would allocate. That would be cool. Although it depends entirely on which compiler is used.

s-ludwig: Yeah, that was actually my first thought, too, doing it at compile time. But for a first version, this is totally sufficient.

The long term goal would be to add some `@nogc` unit tests for cases where there should be no allocations necessary. But there is still a *lot* to do/annotate before that will work. I made some attempts, but it was mostly impossible due to so many Phobos functions that either allocated, or weren't annotated `@nogc`.

skoppe:
Hmm... groupBy isn't available on 2.066.1

s-ludwig:
Huh... I would have never expected this, I thought that that had been included ages ago, but obviously I mixed it up with `group()`. In that case it looks like some classical loops are needed, there seems to be no useful replacement in 2.066.

source/vibe/web/rest.d
```diff
@@ -79,16 +80,29 @@ URLRouter registerRestInterface(TImpl)(URLRouter router, TImpl instance, RestInt
 		router.registerRestInterface!T(__traits(getMember, instance, fname)(), intf.subInterfaces[i].settings);
 	}
 
+
 	foreach (i, func; intf.RouteFunctions) {
 		auto route = intf.routes[i];
 
 		// normal handler
-		auto handler = jsonMethodHandler!(func, i)(instance, intf);
+		auto handler = jsonMethodHandler!(func, i)(instance, intf, settings);
```
s-ludwig: There is `intf.settings`, so `settings` doesn't need to be passed here as an additional parameter. This also has the advantage of always being `!is null`, so that all `null` checks can be removed.

s-ludwig:
Apart from the `settings` parameter, looks good to merge now.

skoppe:
In the tests/restclient I depended on directly referencing/changing the settings, with intf.settings that is not possible since it is copied. Will have to change that.

skoppe:
Hmm, I might have done something wrong with rebasing...

I wanted to notch travis to do a recompile (I read somewhere that the libasync was working again), and also figured an update wouldn't hurt. Now all these other commits show up. Oops... Well, at least the CI passes.

What is the normal strategy here?

Should I cleanup this stuff by creating new branch?

s-ludwig:
Okay, thanks! Looks good to merge now. It'll not make it into the upcoming release, but I can tag an early alpha release later to make it available for testing.

PR #1301, opened 2015-10-21
Add HTTPServerRequest.routerRootDir property

Yoplitein:
This allows for emitting the correct path when using a URLRouter with a prefix.

Consider the index of a router with prefix "/a/deep/path" which renders:
```jade
a(href="#{req.rootDir}") rootDir
a(href="#{req.routerRootDir}") routerRootDir
```
The following is output:
```html
<a href="../../../">rootDir</a>
<a href="../../../a/deep/path/">routerRootDir</a>
```

### DIFF
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -251,6 +251,7 @@ final class URLRouter : HTTPServerRequestHandler {
 	/// Handles a HTTP request by dispatching it to the registered route handlers.
 	void handleRequest(HTTPServerRequest req, HTTPServerResponse res)
 	{
+		req.prefix = m_prefix;
 		auto method = req.method;
 
 		auto path = req.path;
--- a/source/vibe/http/router.d
+++ b/source/vibe/http/router.d
@@ -255,16 +255,14 @@ final class URLRouter : HTTPServerRequestHandler {
 
 		string calcBasePath()
 		{
-			import std.string: split;
-			
-			string combined = req.rootDir ~ m_prefix;
-			string result;
-			
-			foreach(part; combined.split("/"))
-				if(part != null)
-					result ~= part ~ "/";
-			
-			return result;
+			import vibe.inet.path;
+			auto prefix = m_prefix;
+			if(prefix == null) prefix = "/";
+			auto path = Path(req.path);
+			if(!path.endsWithSlash) path = path[0 .. $ - 1];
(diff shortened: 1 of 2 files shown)


s-ludwig:
This idea sounds useful, I worked around this by hand for some projects already. For new projects , something similar for `registerWebInterface` would be similarly useful.

I see a two issues with the concrete approach:
 - This obviously only works for a single router, but routers could be nested. Probably not the common case, but it exists. A solution would be to update the request when an actual route match was found instead of once at the beginning.
 - The router is a higher level concept that is layered on top. The low-level HTTP API shouldn't be affected by it. What I'd propose is to simply put the relative path into `HTTPServerRequest.params["routerRootDir"]` instead. For computing the relative path, it's probably a good idea to use `Path.relativeTo`, to get a shorter path.

Yoplitein:
I'm trying to use Path, but I can't figure out how to wrangle it into submission.
The following was the only way I could get it to actually do anything:
```d
Path path = Path([PathEntry("a"), PathEntry("deep"), PathEntry("path")], true);
Path root = Path([PathEntry(".."), PathEntry(".."), PathEntry("..")], true);
string routerRootDir = path.relativeTo(root).toString;
```
Everything else causes the assert at path.d:145 to fail.
And it still emits `../../../a/deep/path`.

How do I properly go about this?

s-ludwig:
It should be `Path(router.prefix).relativeTo(Path(req.path)).toString()` - assuming that both, `router.prefix` and `req.path`, start with a "/", which should always be true.

Yoplitein:
That doesn't seem to be quite right either.

For paths without a trailing slash, it leaves off the topmost path component. E.g. `/a/deep/path/param` => `/a/deep/`
And at the index of the router, it is equivalent to req.rootDir.

s-ludwig:
Hm right, forgot about that, this issue just came up somewhere else recently. Trailing slash handling doesn't quite match the rules needed here. This should fix it:
```
auto rp = Path(req.path);
if (!rp.endsWithSlash) rp = rp[0 .. $-1];
string result = Path(prefix).relativeTo(req.path).toString();
```

Yoplitein:
Ok, I think that's all the edge cases ironed out. Thanks for the help.

s-ludwig:
Thanks, looks good.

PR #1306, opened 2015-10-28
Added optional `new` and `upsert` parameters to findAndModify

IrenejMarc:
Functionality described here: https://docs.mongodb.org/manual/reference/method/db.collection.findAndModify/


s-ludwig:
Interestingly, there is an almost identical PR: #1150

My problem is with the `bool` parameters. They are hard to read at the call site and individual parameters are not very extensible. For the time being I've added a generic `options` parameter that does the job and at least is extensible (but unfortunately not statically checked).

PR #1313, opened 2015-11-01
Update WebSocket sample

WilliamDunne:
Previous code didn't work as intended, and was also a strange way of writing it. Easier to put directly in while conditions

### DIFF
--- a/examples/websocket/source/app.d
+++ b/examples/websocket/source/app.d
@@ -26,9 +26,8 @@ void handleWebSocketConnection(scope WebSocket socket)
 {
 	int counter = 0;
 	logInfo("Got new web socket connection.");
-	while (true) {
+	while (socket.connected) {
 		sleep(1.seconds);
-		if (!socket.connected) break;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Can you detail what didn't work in the old code? Looking at the code I believe that the reason for writing it that way was to handle the case where the WebSocket gets closed during the `sleep` call. With the condition in the `while` header, that would result in an exception being thrown now.

MrDunne:
That didn't work in its entirety - regardless of how I closed the websocket it wasn't actually closing itself. Similarly, I didn't see any errors when closing it with the changes - I tested a couple of times and I can't imagine much time is spent in the rest of the loop.

dariusc93:
During my test of the websocket example, it always close for me. Could you provide some code of your own to show that it isnt closing exactly as it is written in the code?

s-ludwig:
Closing for now. The existing code looks like it should be correct. Would be interesting to determine why it failed.

Issue #1315, opened 2015-11-02
Does not compile on recent Debian versions because of openssl brokenness


lultimouomo:
It appears that Debian has removed SSLv3_client_method, SSLv3_server_method and SSLv3_method from OpenSSL, so vibe.d fails to build on Debian; see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=768476
I'll be sending a PR shortly using SSLv23 methods with the appropriate options instead.

PR #1316, opened 2015-11-02
Dont use OpenSSL SSLv3* methods

lultimouomo:
They have been disabled on Debian, causing vibe.d to fail to compile.
SSLv3 only connections can still be achieved by passing the appropriate
SSL_OP_NO_* options to OpenSSL

See #1315


dariusc93:
I thought vibe.d was using botan by default? Could be wrong though

s-ludwig:
It's still using OpenSSL by default. Botan is still critical w.r.t. linking on Windows due to its size (although it received some improvements recently). You need to explicitly `dub fetch botan` to make it get used, but then it will currently still link against OpenSSL. D-Programming-Language/dub#361 is needed to be able to freely opt out/in of a certain implementation.

s-ludwig:
Changes look good. I'll merge and tag another RC.

PR #1319, opened 2015-11-04
Test with DMD 2.069.0

Geod24:
http://forum.dlang.org/post/rdymrxccnqxfpjokhdgr@forum.dlang.org


Geod24:
Oh I hadn't see you tagged 0.7.26
I'm inclined not to remove support for 2.066.1 (even if I'd really like to) until GCC is ready.

s-ludwig:
Yeah, we definitely still need to support 2.066.1. One can only hope that some day LDC/GDC will be part of the autotester...

PR #1323, opened 2015-11-06
json.d Browser compatibility: Don't output nan for floating point values.

rjmcguire:
Can't output nan in JSON documents. Browsers don't support it. {"float":nan} is not valid JSON for example.

### DIFF
--- a/source/vibe/data/json.d
+++ b/source/vibe/data/json.d
@@ -1673,7 +1673,7 @@ struct JsonStringSerializer(R, bool pretty = false)
 			else static if (is(T == bool)) m_range.put(value ? "true" : "false");
 			else static if (is(T : long)) m_range.formattedWrite("%s", value);
 			else static if (is(T == BigInt)) m_range.formattedWrite("%d", value);
-			else static if (is(T : real)) m_range.formattedWrite("%.16g", value);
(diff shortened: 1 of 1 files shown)


s-ludwig:
I think we should go with the same strategy as [std_data_json](https://github.com/s-ludwig/std_data_json/blob/master/source/stdx/data/json/generator.d#L450) and ouput `null` instead. This of course requires that the deserializer is also adjusted and converts `null` to `NaN`. Converting to 0 on the other hand looks like it would inadvertently hide potential bugs.

rjmcguire:
What would it be used for (other than internally as a novelty format)  if
it can't be used with web browsers?

s-ludwig:
> What would it be used for (other than internally as a novelty format)  if
it can't be used with web browsers?

I was suggesting to output `null` instead of `0`, which is also valid JSON. If you mean those special `NaN` and `Infinity` literals in the std_data_json code, please don't ask me, ask those that have requested it ;) (but it's a compile-time option, so no harm done by default)

rjmcguire:
Explains why the Chrome browsers  json parser's error is "unexpected 'a' "
when it finds nan.

Should I just delete my PR or should I change it? There is at least one
unittest that checks nan serialization but it doesn't seem to test the code
I changed.

s-ludwig:
Got fixed by #1548.

PR #1324, opened 2015-11-06
Fix leak when a connection is closed by the server

### LINKED ISSUE #1321: HTTP client leaks memory.
Configuration: CentOS 7 x86_64, vibe.d-0.7.26, libevent driver

There is something wrong in HTTP client error handling.
Client code:
```d
import vibe.d;

shared static this() {
	runTask({
		while(true) {
			try {
				requestHTTP("http://127.0.0.1:8080", (scope req) {
				}, (scope res) {
					res.dropBody();
				});
			} catch(Exception e) {

			}
			sleep(1.msecs);
		}
	});
}
```
If to run this code without a HTTP server, the client starts to leak memory very fast.

But if to run a simple HTTP server:
```d
import vibe.d;

shared static this()
{
	settings.port = 8080;

	listenHTTP(settings, (req, res) {
		res.writeBody("Hello, World!", "text/plain");
	});
}
```
The client immediately stops leaking. Stop the server and you will see leak again.
If to change server code to throwing error:
```d
listenHTTP(settings, (req, res) {
	throw new HTTPStatusException(HTTPStatus.badRequest);
});
```
the client leaks, but much more slowly then without a server. But still leaks.

etcimon:
The higher level doesn't implement the means to track passively closed connections, other than the actual object's destructor.

For this reason, remotely closed connections will leave objects on the heap unless we finish cleaning up in `~this`.

fixes #1321

### DIFF
--- a/source/vibe/core/drivers/libevent2_tcp.d
+++ b/source/vibe/core/drivers/libevent2_tcp.d
@@ -85,10 +85,12 @@ package final class Libevent2TCPConnection : TCPConnection {
 		bufferevent_setwatermark(m_ctx.event, EV_READ, 0, 65536);
 	}
 
-	/*~this()
+	~this()
 	{
+		if (m_ctx && m_ctx.state == ConnectionState.passiveClose)
+			TCPContextAlloc.free(m_ctx);
 		//assert(m_ctx is null, "Leaking TCPContext because it has not been cleaned up and we are not allowed to touch the GC in finalizers..");
-	}*/
+	}
 
 	@property void tcpNoDelay(bool enabled)
 	{
--- a/source/vibe/core/drivers/libevent2_tcp.d
+++ b/source/vibe/core/drivers/libevent2_tcp.d
@@ -376,6 +376,7 @@ package final class Libevent2TCPConnection : TCPConnection {
 		if (m_ctx && m_ctx.state == ConnectionState.passiveClose) {
 			if (m_ctx.event) bufferevent_free(m_ctx.event);
 			TCPContextAlloc.free(m_ctx);
+			m_ctx = null;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Not sure about the one in `finalize`, but it's good that the one in `~this` is now legal (missed that). But I'd remove the check for passive close in the latter, because we shouldn't leak for abandoned but alive connections either. Would also be good to output a warning to the log (now that logging doesn't allocate by default).

But regardless of this low-level fix, this also needs to be fixed in the HTTP client. I'll have a look at that later.

etcimon:
> Not sure about the one in finalize, but it's good that the one in ~this is now legal (missed that). But I'd remove the check for passive close in the latter, because we shouldn't leak for abandoned but alive connections either. Would also be good to output a warning to the log (now that logging doesn't allocate by default).

I think destruction logic belongs in finalize on every stream, because that's the last place that gets called before the GC comes around. It should be explicitely mentioned too in the definition of streams, because too often on my fork the destructor threw and triggered an infinite loop of invalid memory operations. I still get them every now and then, never sure of where it comes from.

vitalka200:
@etcimon / @s-ludwig ,
Is there any chance it can be merged?

s-ludwig:
@vitalka200 Do you still use the libevent driver? It looks like it wouldn't hurt to merge this, but I'm still a little hesitant, because this is ancient code, and vibe-core is a pretty mature replacement nowadays. There is a certain risk that this breaks something in a subtle way for anyone who still relies on libevent.

BTW, I expect the libevent driver to be phased out after the release of 0.8.6, unless someone raises their voice against it.

vitalka200:
@s-ludwig Yep. 

I found it to leak less memory than vibe-core at the time when I tested it on simple "Hello, World!" example and was slightly more performant.

I've tried vibe-core again and it looks pretty good but still uses slightly more CPU than libevent wrapper (for my use-case, mainly regex matching on kubernetes pod with 0.5 cpu and 512M ram and constant ~600req/sec ).

Thank you for such great framework!

PR #1332, opened 2015-11-12
added support for websocket clients

Kemonozume:
added ws/wss to the URL package
made changes to the Frame struct  and Websocket class to allow it to be
used as a server and client.

### DIFF
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -731,11 +731,12 @@ struct Frame {
 // copied from client.d not sure how to make visible for websockets.d so we avoid creating a new object
 private __gshared HTTPClientSettings defaultSettings = new HTTPClientSettings;
 
-private ubyte[] generateNewMaskKey()
+private ubyte[] generateNewMaskKey() 
 {
-    auto gen = Random(unpredictableSeed);
-    auto a  = uniform(0, 4294967295, gen);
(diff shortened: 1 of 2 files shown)


Kemonozume:
The ping/pong messages work properly, kinda missed the debug messages. 
Im not sure how to replace the auto gen = Random(unpredictableSeed); with the secure RNG from vibe.crypto.cryptorand.

source/vibe/http/websockets.d
```diff
@@ -64,6 +70,59 @@ class WebSocketException: Exception
...
+        req.requestURL = (url.localURI == "") ? "/" : url.localURI;
+        req.method = HTTPMethod.GET;
+        req.headers["Upgrade"] = "websocket";
+        req.headers["Connection"] = "Upgrade";
+        req.headers["Sec-WebSocket-Version"] = "13";
+        req.headers["Sec-WebSocket-Key"] = challengeKey;
+    });
+
+    auto key = "sec-websocket-accept" in res.headers;
+    assert(key, answerKey);
```
Geod24: You don't want an assert here, rather an enforce.

s-ludwig:
Thanks @Kemonozume, I think this looks good to merge now.

I'll also add a basic test project that sends a few messages between the client and server.

PR #1336, opened 2015-11-19
catch non Url Encoded Cookies

### LINKED ISSUE #1228: another cookie parsing problems
```
dub init -t vibe.d test
cd test
dub run &
curl localhost:8080 -H 'Cookie: __context={"SLT":"%f,*]","UID":""};'
```
and watch it crash:

```
400 - Bad Request

Bad Request

Internal error information:
object.Exception@../../../.dub/packages/vibe-d-0.7.24/source/vibe/textfilter/urlencode.d(179): invalid percent encoding
----------------
5   cookiefuckup                        0x000000010ca3e9e3 pure @safe bool std.exception.enforce!(Exception, bool).enforce(bool, lazy const(char)[], immutable(char)[], ulong) + 107
6   cookiefuckup                        0x000000010cb96b09 pure @safe void vibe.textfilter.urlencode.filterURLDecode!(std.array.Appender!(immutable(char)[]).Appender).filterURLDecode(ref std.array.Appender!(immutable(char)[]).Appender, immutable(char)[], bool) + 257
7   cookiefuckup                        0x000000010cb96280 @safe immutable(char)[] vibe.textfilter.urlencode.urlDecode(immutable(char)[]) + 112
8   cookiefuckup                        0x000000010cb7903d void vibe.http.server.parseCookies(immutable(char)[], ref vibe.http.common.CookieValueMap) + 369
9   cookiefuckup                        0x000000010cb7766b bool vibe.http.server.handleRequest(vibe.core.stream.Stream, vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo, ref vibe.http.server.HTTPServerSettings, ref bool) + 4527
10  cookiefuckup                        0x000000010cb763bd void vibe.http.server.handleHTTPConnection(vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo) + 425
11  cookiefuckup                        0x000000010cb75da6 void vibe.http.server.listenHTTPPlain(vibe.http.server.HTTPServerSettings).doListen(vibe.http.server.HTTPListenInfo, bool).__lambda3(vibe.core.net.TCPConnection) + 54
12  cookiefuckup                        0x000000010cada05b void vibe.core.drivers.libevent2_tcp.ClientTask.execute() + 1191
13  cookiefuckup                        0x000000010cacd644 void vibe.core.core.makeTaskFuncInfo!(void delegate()).makeTaskFuncInfo(ref void delegate()).callDelegate(vibe.core.core.TaskFuncInfo*) + 120
14  cookiefuckup                        0x000000010caca9e7 void vibe.core.core.CoreTask.run() + 299
15  cookiefuckup                        0x000000010cbbb3f0 void core.thread.Fiber.run() + 44
16  cookiefuckup                        0x000000010cbbafd6 fiber_entryPoint + 98
17  ???                                 0x0000000000000000 0x0 + 0

```

not sure if simply urldecoding  is the  correct thing to do. there is a few rfc, notably i found
http://tools.ietf.org/html/rfc6265 via http://stackoverflow.com/questions/1969232/allowed-characters-in-cookies

yannick:
catches urlDecode errors and just sanitizes the string;

people in the world wide web send all kind of weird cookies and its uncool if the
request crashes :/

fixes #1228

### DIFF
--- a/source/vibe/http/common.d
+++ b/source/vibe/http/common.d
@@ -669,4 +669,9 @@ struct CookieValueMap {
 			}
 		return null;
 	}
+
+	auto length()
+	{ 
+		return m_entries.length;
+	}
 }
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -1810,6 +1810,7 @@ private void parseRequestHeader(HTTPServerRequest req, InputStream http_stream, 
 
 private void parseCookies(string str, ref CookieValueMap cookies)
 {
+	import std.encoding : sanitize;
 	while(str.length > 0) {
 		auto idx = str.indexOf('=');
 		enforceBadRequest(idx > 0, "Expected name=value.");
@@ -1818,9 +1819,21 @@ private void parseCookies(string str, ref CookieValueMap cookies)
 
 		for (idx = 0; idx < str.length && str[idx] != ';'; idx++) {}
 		string value = str[0 .. idx].strip();
-		cookies[name] = urlDecode(value);
+		try {
+			cookies[name] = value.urlDecode();
+		} catch (Exception e) {
+			cookies[name] = value.sanitize();
+		}
 		str = idx < str.length ? str[idx+1 .. $] : null;
 	}
+}
+
+unittest
+{
+	auto cvm = CookieValueMap();
+	string nonUrlEncodedCookie = "test=%%pX;";
+	parseCookies(nonUrlEncodedCookie, cvm);
+	assert(cvm["test"] == "%%pX" );
 }
 
 shared static this()
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -1814,26 +1814,13 @@ private void parseCookies(string str, ref CookieValueMap cookies)
 	while(str.length > 0) {
 		auto idx = str.indexOf('=');
 		enforceBadRequest(idx > 0, "Expected name=value.");
-		string name = str[0 .. idx].strip();
+		string name = str[0 .. idx].strip().sanitize();
 		str = str[idx+1 .. $];
-
 		for (idx = 0; idx < str.length && str[idx] != ';'; idx++) {}
-		string value = str[0 .. idx].strip();
-		try {
-			cookies[name] = value.urlDecode();
-		} catch (Exception e) {
-			cookies[name] = value.sanitize();
-		}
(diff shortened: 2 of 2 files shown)


schuetzm:
Wouldn't it be better to respond with "400 Bad Request"? That would make it explicit that it's an error on the client side.

s-ludwig:
The problem is that cookies are actually not required to be URL encoded, the only requirement is to not contain illegal characters. In the HTTP/2 branch there are changes that makes this configurable when getting/setting the cookie value and I think that this is the right approach: decode not while parsing the HTTP header, but when actually requesting the cookie value. Then the user code can properly handle any decoding errors and we can also have a "raw" mode.

yannick:
ok i'll check that branch there. 
handling it on read is a good strategy.
 
@schuetzm not a good idea imho. often times services such as adservers write cookies on your domain.
and sometimes these contain weird stuff. thats actually how i discovered this. pretty nasty surprise if your service fails because of one cookie that you don't even care about.

s-ludwig:
The parsing code looks good. I'll see if I can extract the API-side from the other branch. We should also add some unit tests for such rather deep changes.

yannick:
i'm happy to write test and improve this implementation, but i think we should decide how far down this rabbit hole we want to go:

http://stackoverflow.com/questions/1969232/allowed-characters-in-cookies

shall we support illegal cookies that are supported by some browsers?
what about  cases l ike
```
Cookie: bar
```
```
Cookie: =bar
```
```
Cookie: foo=
```
i vote to *not* support this, since in a normal case vibe.d is in control of the cookies.
essentially sticking to rfc6265.
for the cases where someone needs to support weird cookies we should offer the possibility 
to have a middleware to do the parsing instead of this builtin method.

yannick:
@s-ludwig whats needed to get this in? :pray:

s-ludwig:
Some time ago I've made the hard work that was unfortunately necessary to support non-urlencoded cookies without breaking backwards compatibiilty (9262e62). On master, CookieValueMap now internally stores the raw value and decodes only once the urldecoded value is requested. Unfortunately, this change requires some really ugly things to ensure backwards compatibility. Your PR can now be applied, but it requires a slight modification, so that it stores the cookies in the `rawValue` field instead of in the `value` property.

s-ludwig:
Superseded by #1401.

PR #1339, opened 2015-11-28
Fix reading from ZlibInputStream with exact buffer size

villytiger:


s-ludwig:
Thanks! Especially good to have more (previously) failing test cases for this code, which used to be quite brittle due to zlib's awful and badly documented API.

PR #1340, opened 2015-11-29
explicit chunk control

yannick:
for the implementation of a S3 Library we needed more granular control over the http chunks.
( see https://github.com/tamediadigital/vibe-aws/tree/feature-s3 )

this PR should not break any existing api but just adds the methods to manipulate the chunks headers.

### DIFF
--- a/source/vibe/http/common.d
+++ b/source/vibe/http/common.d
@@ -427,11 +427,13 @@ final class ChunkedInputStream : InputStream {
 	Outputs data to an output stream in HTTP chunked format.
 */
 final class ChunkedOutputStream : OutputStream {
+	alias ChunkExtensionCallback = string delegate(in ubyte[] data);
 	private {
 		OutputStream m_out;
 		AllocAppender!(ubyte[]) m_buffer;
 		size_t m_maxBufferSize = 512*1024;
 		bool m_finalized = false;
+		ChunkExtensionCallback m_chunkExtensionCallback = null;
 	}
 
 	this(OutputStream stream, Allocator alloc = defaultAllocator())
@@ -449,18 +451,51 @@ final class ChunkedOutputStream : OutputStream {
 	/// ditto
 	@property void maxBufferSize(size_t bytes) { m_maxBufferSize = bytes; if (m_buffer.data.length >= m_maxBufferSize) flush(); }
 
+	/** A delegate used to specify the extensions for each chunk written to the underlying stream.
+	 	
+	 	The delegate has to be of type `string delegate(in const(ubyte)[] data)` and gets handed the
+	 	data of each chunk before it is written to the underlying stream. If it's return value is non-empty,
+	 	it will be added to the chunk's header line.
+
+	 	The returned chunk extension string should be of the format `key1=value1;key2=value2;[...];keyN=valueN`
+	 	and **not contain any carriage return or newline characters**.
+
+	 	Also note that the delegate should accept the passed data through a scoped argument. Thus, **no references
+	 	to the provided data should be stored in the delegate**. If the data has to be stored for later use,
+	 	it needs to be copied first.
+	 */
+	@property ChunkExtensionCallback chunkExtensionCallback() const { return m_chunkExtensionCallback; }
+	/// ditto
+	@property void chunkExtensionCallback(ChunkExtensionCallback cb) { m_chunkExtensionCallback = cb; }
+
+	private void append(scope void delegate(scope ubyte[] dst) del, ulong nbytes)
+	{
+		assert(del !is null);
+		auto sz = nbytes;
+		if (m_maxBufferSize > 0 && m_maxBufferSize < m_buffer.data.length + sz)
+			sz = m_maxBufferSize - min(m_buffer.data.length, m_maxBufferSize);
+
+		if (sz > 0)
+		{
+			m_buffer.reserve(sz);
+			m_buffer.append((scope ubyte[] dst) {
+					debug assert(dst.length >= sz);
+					del(dst[0..sz]);
+					return sz;
+				});
+		}
+	}
+
 	void write(in ubyte[] bytes_)
 	{
 		assert(!m_finalized);
 		const(ubyte)[] bytes = bytes_;
 		while (bytes.length > 0) {
-			auto sz = bytes.length;
-			if (m_maxBufferSize > 0 && m_maxBufferSize < m_buffer.data.length + sz)
-				sz = m_maxBufferSize - min(m_buffer.data.length, m_maxBufferSize);
-			if (sz > 0) {
-				m_buffer.put(bytes[0 .. sz]);
-				bytes = bytes[sz .. $];
-			}
+			append((scope ubyte[] dst) {
+					auto n = dst.length;
+					dst[] = bytes[0..n];
+					bytes = bytes[n..$];
+				}, bytes.length);
 			if (bytes.length > 0)
 				flush();
 		}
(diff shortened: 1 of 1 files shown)


mfrischknecht:
> this PR should not break any existing api

It might be necessary to note that I changed an implicit behavior in the `finalize()` method: It is now possible to call it multiple times without triggering an exception. The reason for this is that vibe.d's HTTP server internally always does this while I also used the method to transmit the final (0-sized) chunk. 
This is mainly due to the fact that S3 requires chunk signatures for every chunk (including the last) and I had to prevent the stream from sending a second 0-sized chunk when the `finalize()` Is called by the server.

s-ludwig:
I'm just a little worried about the API changes and wonder if a slightly more general solution would work. What about defining a callback `string delegate(const(ubyte)[] data) chunkExtensionCallback` that is called for every chunk? That would mean that the ordinary stream API continues to work in that scenario, too.

mfrischknecht:
I think that approach should work for the S3 use case. I'll have a go at the implementation and update the branch accordingly.

mfrischknecht:
Okay, I reverted the first couple of commits and added the delegate property instead. There are a couple of points worth looking at though.
To get rid of the reverts, I could squash the commits together into one if you are statisfied with the solution as is.

source/vibe/http/common.d
```diff
@@ -426,11 +426,13 @@ final class ChunkedInputStream : InputStream {
 	Outputs data to an output stream in HTTP chunked format.
 */
 final class ChunkedOutputStream : OutputStream {
+	alias ChunkExtensionCallback = string delegate(in ubyte[] data);
```
mfrischknecht: Note that I changed the delegate's signature to taking an argument of `in ubyte[]` instead of `const(ubyte)[]`. This should prove to be especially useful for a `VibeManualMemoryManagement` version later on (and, in fact, already is now, since one `write()` overload also takes `in ubyte[]`).

source/vibe/http/common.d
```diff
@@ -469,20 +488,25 @@ final class ChunkedOutputStream : OutputStream {
 	{
 		assert(!m_finalized);
 		if( m_buffer.data.length > 0 ) flush();
+		ubyte[] buffer = null;
+		auto getBytes = (size_t size) {
+			if (buffer is null || buffer.length < size)
+				buffer = new ubyte[](size);
+				return buffer[0..size];
+			};
```
mfrischknecht: This could easily be done using an appropriate allocator from memutils.

source/vibe/http/common.d
```diff
@@ -469,20 +488,25 @@ final class ChunkedOutputStream : OutputStream {
...
 			while( !data.empty ){
 				auto sz = data.leastSize;
 				assert(sz > 0);
-				writeChunkSize(sz);
-				m_out.write(data, sz);
-				m_out.write("\r\n");
-				m_out.flush();
+				auto bytes = getBytes(sz);
+				data.read(bytes);
+				write(bytes);
```
mfrischknecht: This change will result in the stream always using chunks of the size `maxBufferSize`, except for the last one. This is an implicit requirement for the S3 usecase (since chunks are not to allowed to be under a certain size).
Using the new extension mechanism it is also the easiest solution, since all data written to the stream has to be put into a buffer in this class anyway (in order to be able to call the delegate). It will however differ from the current chunking behavior. Is this a problem?

source/vibe/http/common.d
s-ludwig: Sounds good.

source/vibe/http/common.d
s-ludwig: Why not just reuse `m_buffer` here?

source/vibe/http/common.d
s-ludwig: I think that should be fine. If specific smaller chunk sizes are required, it's always possible to do something like `vibe.stream.operations.pipeRealtime`.

source/vibe/http/common.d
mfrischknecht: Mainly because I wanted to keep the buffer management in one place. But if I take the flushing out, I should be able to move that into it's own method.

s-ludwig:
> To get rid of the reverts, I could squash the commits together into one if you are statisfied with the solution as is.

Would be good, it's a nicely isolated change that fits well in a single commit.

source/vibe/http/common.d
s-ludwig: Flushing can indeed be removed here, it was just necessary to cleanly switch between buffered and stream modes. I just realized now that `m_buffer` is an `AllocAppender`. We'd need something like `AllocAppender.put(scope size_t delegate(scope T[] dst) del);`, where the delegate is called with `m_remaining` and the return value is the number of bytes written to that buffer.

source/vibe/http/common.d
mfrischknecht: Actually, I'm not sure if that would still be reasonable. I thought about writing a method that requests N bytes in the buffer and returns the according range in the buffer (so one could `stream.read(appendBytes(N))`), but the `Appender` struct seems not to be made for this case.
To be able to conveniently append data to the buffer, one seems to basically always have a range of items already allocated.

This could work, but I'm not sure if it is still efficient:
```{D}
private ubyte[] appendBytes(ulong maxSize)
{
    if (m_maxBufferSize > 0 && m_maxBufferSize < m_buffer.data.length + maxSize)
        maxSize = m_maxBufferSize - min(m_buffer.data.length, m_maxBufferSize);
    auto pos = m_buffer.data.length;
    for (ulong i = 0; i < maxSize; ++i)
        m_buffer.put(0);
    return m_buffer.data[pos..pos+maxSize];
}
```

Then I could basically do:
```{D}
auto bytes = appendBytes(sz);
if (bytes.length > 0)
    data.read(bytes);
else
    flush();
```

And for the `write(in ubyte[])` method it would be:
```{D}
while (bytes.length > 0) {
    auto sz = bytes.length;
    auto dst = appendBytes(sz);
    if (dst.length > 0)
    {
        dst[0..sz] = bytes[0..sz];
        bytes = bytes[sz..$];
    }
    else
        flush();
}
```

Do you perhaps have a better idea? Another variant could be to use initialize the appender with the according block size and use it like a normal array as long as `m_maxBufferSize > 0` (basically only using `put()` for unbounded chunk sizes).

mfrischknecht: Actually, I have been looking at the wrong thing all along - guess I didn't read the type toroughly enough and just assumed a `std.array.Appender`.  Yeah, if we could have a new method for this use case, that would be a nice solution.

source/vibe/http/common.d
mfrischknecht: As noted above, you can ignore the comment about appending bytes then (my bad). Should I have a look at that too and post another PR (would take some time though)?

source/vibe/http/common.d
s-ludwig: I think `AllocAppender` should definitely get extended. Either by adding an `append` method similar to the one you mentioned, or using the scheme that I mentioned in my reply to the wrong message thread:
> Flushing can indeed be removed here, it was just necessary to cleanly switch between buffered and stream modes. I just realized now that m_buffer is an AllocAppender. We'd need something like AllocAppender.put(scope size_t delegate(scope T[] dst) del);, where the delegate is called with m_remaining and the return value is the number of bytes written to that buffer.

The advantage of the callback based version would be that it could be made to work with `immutable(T)[]`, but using a method that just returns a buffer is of course much nicer to look at.

source/vibe/http/common.d
s-ludwig: I'll have quick look at that and commit that to master.

mfrischknecht: Thank you. Actually, once this `put()` overload is implemented, one could even build something like `AllocAppender.put(InputStream data, ulong size)` on top of it - this seems (at least to me) to be a much more obvious use case from an outside perspective.

s-ludwig: Done now (6ba887e28b84fc54ca11a732c0585146e1b22040) and Travis looks like it will pass.

source/vibe/http/common.d
```diff
@@ -449,18 +451,51 @@ final class ChunkedOutputStream : OutputStream {
...
+
+	 	Also note that the delegate should accept the passed data through a scoped argument. Thus, **no references
+	 	to the provided data should be stored in the delegate**. If the data has to be stored for later use,
+	 	it needs to be copied first.
+	 */
+	@property ChunkExtensionCallback chunkExtensionCallback() const { return m_chunkExtensionCallback; }
+	/// ditto
+	@property void chunkExtensionCallback(ChunkExtensionCallback cb) { m_chunkExtensionCallback = cb; }
+
+	private void append(scope void delegate(scope ubyte[] dst) del, ulong nbytes)
```
mfrischknecht: Okay, thanks again for the quick responses and extension of `AllocAppender`!
I integrated the `AllocAppender.append()` calls and squashed everything into a single commit. To homogenize the buffer handling in both `write()` overloads I implemented a new (private) `append()` method. 

Do you think there's further need for modifications (any obvious gotchas)?

s-ludwig:
Looks good! Thanks for doing the modifications!

PR #1341, opened 2015-11-30
Support IPv6 Urls

Marenz:
I tried running the unittests but somehow they all succeeded even though I deliberately broke one.
Still, the code seems to work with a simple test. I am secretly hoping for a CI that tests for me

### DIFF
--- a/source/vibe/inet/url.d
+++ b/source/vibe/inet/url.d
@@ -94,11 +94,31 @@ struct URL {
 					}
 
 					m_host = str[hs .. si];
-					auto pi = m_host.indexOfCT(':');
-					if(pi > 0) {
-						enforce(pi < m_host.length-1, "Empty port in URL.");
-						m_port = to!ushort(m_host[pi+1..$]);
-						m_host = m_host[0 .. pi];
+
+					auto findPort ( string src )
+					{
+						auto pi = src.indexOfCT(':');
+						if(pi > 0) {
+							enforce(pi < src.length-1, "Empty port in URL.");
+							m_port = to!ushort(src[pi+1..$]);
+						}
+						return pi;
+					}
+
+
+					auto ip6 = m_host.indexOfCT('[');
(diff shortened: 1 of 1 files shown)


s-ludwig:
The unit test issue is probably due to the split into sub packages. Need to use either `dub test --combined` or `dub test :core` now.

The diff looks good so far. The only thing I wonder is if `m_is_ipv6` should be replaced by a check for `.canFind(":")` in `toString()`, to save space, but more importantly to support setting the `.host` property to an IPv6-address after construction.

Marenz:
Updated to use `canFind` and extended unittest to setting `.host` to a ipv6

source/vibe/inet/url.d
```diff
@@ -92,11 +92,31 @@ struct URL {
 					}
 
 					m_host = str[hs .. si];
-					auto pi = m_host.indexOfCT(':');
-					if(pi > 0) {
-						enforce(pi < m_host.length-1, "Empty port in URL.");
-						m_port = to!ushort(m_host[pi+1..$]);
-						m_host = m_host[0 .. pi];
+
+					long findPort ( string src )
```
s-ludwig: This must be either `sizediff_t` or `auto` for the slice operation below to compile with 32-bits.

Marenz:
Changed return type to `auto`. Rebased on latest master.

s-ludwig:
Okay, thanks!

source/vibe/inet/url.d
```diff
@@ -242,7 +262,14 @@ struct URL {
 			dst.put(password);
 			dst.put('@');
 		}
+
+		import std.algorithm.searching : canFind;
```
s-ludwig: Just noticed that this causes the Travis-CI build to fail. Needs to be `import std.algorithm : canFind` for now.

Marenz:
importing `std.algorithm : canFind` now

Issue #500, opened 2014-02-06
Unix sockets' support


nazriel:
Discussion: http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/10870/

I have a WIP branch https://github.com/nazriel/vibe.d/commits/unix_sockets related to this (source/vibe/web/web.d  change is unrelated, sorry)

But I am not sure if I am going towards right direction API wise etc.
@s-ludwig could you look at this please?
Thanks!

JustDevZero:
First of all, let me say I hate the kinda +1, or "bump" posts, but this is one of that kind :/

@s-ludwig Sorry to bothering you, but there has been some progress about that stuff?

I tried searching over the code to see if there was something fairly implemented, but have given up without success.

This option would be extremely interesting, because the projects would easily coexist with existing ones on our machines, besides the improvement would involve.

TETYYS:
First of all, let me say I hate the kinda +1, or "bump" posts, but this is one of that kind :/

skoppe:
I am using it to talk to the docker daemon. It only works with libevent, but that is ok for me.

TETYYS:
very nice

liveresume:
Is it possible to use listenUDP() with a unix socket? Or is it tcp+http only?

skoppe:
only connectTCP. But it would be trivial to implement. On the other hand you could wait for @s-ludwig to rewrite vibe, I bet its gonna support it.

s-ludwig:
Should be possible for the 0.8.0-beta version with the `vibe-core` configuration for `vibe-d:core`. At least it's auto tested for the underlying `eventcore` library, just not yet in the vibe.d context.

s-ludwig:
> The functionality seems to be in Phobos - does vibe.d wrap the same stuff?

No, vibe.d goes directly down to the OS APIs to avoid the overhead of the class based `std.socket` module where possible.

liveresume:
@s-ludwig performance. always nice.  Will try beta out.  Just to confirm, `listenUDP()` should be possible with it?

s-ludwig:
Hm, I just noticed that I forgot to merge the USDS related changes to `NetworkAddress`. Once that is done, `listenUDP` with a USDS `NetworkAddress` should work (the function will be renamed to `listenDatagram` later on).

Geod24:
According to https://github.com/vibe-d/vibe.d/pull/1342#issuecomment-321856809 this is fixed and should be closed.

Issue #1326, opened 2015-11-07
Behavior of ConnectionStream.waitForData is unclear


mfrischknecht:
I have been looking at [ConnectionStream.waitForData()] on TCPConnection objects and found that the API documentation is not clear on some corner cases. After testing the behavior with a [simple test], it seems that the according behavior differs between backends<sup>1</sup>. 

Naively looking at the method, I would expect about the following behavior:
 1. If the method is passed a timeout of `0L` (the default value), it will wait indefinitely (since [InputStream.dataAvailableForRead] already delivers the information without the wait)
 2. If the connection is "empty" and gets closed, it will return false immediately (before the timeout occurs, not blocking up the fiber)
 3. If the connection contains unread data (whether it's still open or not), it returns true immediately

An implementation like this would basically allow for the most convenient loop that allows waiting for data:

```{D}
while(connection.waitForData(timeout))
{
    ubyte[] buffer = new ubyte[](connection.leastSize);
    connection.read(buffer);
    //Do stuff...
}
//Connection was closed or timed out
//All available data has been read
```

On a similar note, the maximum timeout of the method is not specified. Libasync seems to use `Duration.max` as the default value, whereas Libevent will throw exceptions stating that `int.max.seconds` is the maximum.

[ConnectionStream.waitForData()]: http://vibed.org/api/vibe.core.stream/ConnectionStream.waitForData
[simple test]: https://github.com/mfrischknecht/test-vibed-waitForData
[InputStream.dataAvailableForRead]: http://vibed.org/api/vibe.core.stream/InputStream.dataAvailableForRead

<sup>1</sup> On my Linux box, I was only able to build the Libevent and Libasync versions. Libev results in an error about a missing deimos module. Also, the behavior on Windows would be interesting to know, too.

Here's the behavior I observed:

##### Libevent
 1. Is not the case (waitForData(0L) returns immediately)
 2. True
 3. True

##### Libasync
 1. True
 2. True
 3. The test fails for a reason I can't explain. The socket I use won't transfer its payload although I use the flush() method explicitly (is this worth another issue?).

**Edit:**
##### win32
 1. True
 2. True
 3. True

s-ludwig:
[...]

PR #1342, opened 2015-12-01
[WIP] Unix sockets integration- Fixes rejectedsoftware/vibe.d#500

Geod24:
I had a shot at integrating unix sockets with Vibe.d

This is currently not in a mergeable state, for various reasons:
- You might want to discuss the API - I didn't use any connection pool because the number of sockets is more likely to be low, and it can be revisited / implemented on top if I'm proven wrong;
- I had a couple of issues related to the event driver:
  - rejectesoftware/vibe.d#1334
  - As you can see in the last commit, I switched from using `Duration.max` to `int.max.seconds` because it triggers an `assert` in libevent (see below). Right on point was rejectedsoftware/vibe.d#1326 but I think it was a little too `ConnectionStream`-specific. I would favor merging both `FileDescriptorEvent.wait` overloads, and give a special meaning of 0 => returns `true` immediately without yielding if the fd is readable, `false` otherwise, and `Duration.max` to wait indefinitely. Would that make sense / work ?
- `flush` is not yet implemented / tested;
- Several functions are still undocumented;

```
Task terminated with unhandled exception: Assertion failure
core.exception.AssertError@../../source/vibe/core/drivers/libevent2.d(478): Assertion failure
----------------
??:? _d_assert [0x8571ab]
??:? void vibe.core.drivers.libevent2.__assert(int) [0x81a847]
../../source/vibe/core/drivers/libevent2.d:478 void vibe.core.drivers.libevent2.Libevent2Driver.rescheduleTimerEvent(std.datetime.SysTime) [0x7bc474]
../../source/vibe/core/drivers/libevent2.d:421 void vibe.core.drivers.libevent2.Libevent2Driver.rearmTimer(ulong, core.time.Duration, bool) [0x7bbee6]
../../source/vibe/core/drivers/libevent2.d:823 bool vibe.core.drivers.libevent2.Libevent2FileDescriptorEvent.wait(core.time.Duration, vibe.core.driver.FileDescriptorEvent.Trigger) [0x7bdd1b]
../../source/vibe/core/unix.d:209 _D4vibe4core4unix10UnixSocket13startListenerMFDFC4vibe4core4unix10UnixStreamZvZ9__lambda2MFZv [0x7efa79]
../../source/vibe/core/core.d:486 void vibe.core.core.makeTaskFuncInfo!(void delegate()).makeTaskFuncInfo(ref void delegate()).callDelegate(vibe.core.core.TaskFuncInfo*) [0x7167b5]
../../source/vibe/core/core.d:1005 void vibe.core.core.CoreTask.run() [0x7b75f0]
??:? void core.thread.Fiber.run() [0x8b0add]
??:? fiber_entryPoint [0x8b087f]
??:? [0xffffffff]
Program exited with code 1
dub  16.06s user 1.78s system 95% cpu 18.677 total
```

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -28,6 +28,7 @@
 	"targetType": "library",
 	"sourcePaths": [],
 	"sourceFiles": ["source/vibe/d.d", "source/vibe/vibe.d"],
+	"excludedSourceFiles-windows": [ "source/vibe/core/unix.d" ],
 
 	"configurations": [
 		{"name": "libevent", "subConfigurations": {"vibe-d:core": "libevent"}},
--- a/source/vibe/core/unix.d
+++ b/source/vibe/core/unix.d
@@ -1,0 +1,498 @@
+/**
+ * Unix socket implementation compatible with the event driver
+ *
+ * Copyright:
+ * © 2015 RejectedSoftware e.K.
+ *
+ * Authors:
+ * Mathias 'Geod24' Lang
+ *
+ * License:
+ * Subject to the terms of the MIT license, as written in
+ * the included LICENSE.txt file.
+ */
+module vibe.core.unix;
+
+import core.time;
+import core.sys.posix.fcntl;
+import core.sys.posix.unistd : close, write;
+import core.sys.posix.sys.socket;
+
+import std.exception : enforce;
+
+import vibe.core.core;
+import vibe.core.driver;
+import vibe.core.log;
+import vibe.core.stream;
+
+
+/*******************************************************************************
+
+	Class to instantiate connections to/via unix sockets.
+
+	This class only support streaming (SOCK_STREAM) sockets.
+
+	Server example:
+	---
+	// Note: Path starting with '\0' are in the abstract linux namespace
+	auto server = UnixSocket.listen("\0test", &handleSingleConnection);
+	// After a while...
+	server.close(); // Stop listening on the socket for new connections
+	---
+
+	Client example:
+	---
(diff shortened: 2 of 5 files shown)


s-ludwig:
Interesting that you decided to implement this on top of `FileDescriptorEvent` instead of putting it into the driver, but I guess it makes sense from a code-reuse viewpoint. Eventually I'd like to revamp the whole `vibe.core.drivers` package, but until then I guess that should be fine.

Just the API should be harmonized a bit with the other socket related code (e.g. global `connectUnixSocket` instead of `UnixSocket.connect`), and the code style doesn't match the rest of the library in some aspects (mostly comment style and placement of private member variables).

> I would favor merging both FileDescriptorEvent.wait overloads, and give a special meaning of 0 => returns true immediately without yielding if the fd is readable, false otherwise, and Duration.max to wait indefinitely. Would that make sense / work ?

Makes sense! I'm still discovering such places from time to time and all of them should be changed to this scheme... maybe I should just search for all `0.seconds` occurrences....

I'd make this change a two-step process, though, to avoid breaking existing code: First let 0 still wait indefinitely, but emit a `logWarn` that explains the upcoming behavior change, then after a few releases we'll finally switch the behavior. I've already done it this way for the `ConnectionStream` changes.

skoppe:
See also #1547

s-ludwig:
Got UNIX socket support integrated now through additions to the libevent driver, as well as the vibe-core/eventcore based implementation.

PR #1345, opened 2015-12-02
Fix #1343 - re-enable std.concurrency integration

### LINKED ISSUE #1343: dmd 2.069.1 regression, std.concurrency functions sleep forever
I move my old project from dmd 2.066 to latest release. The problem: `std.concurrency` functions (
`receive`, `receiveOnly`, `receiveTimeout`) hang vibed's fibers.

Demo of issue
===========

dub.sdl
```
name "threads-test"
dependency "vibe-d" version="==0.7.27-alpha.1"
versions "VibeDefaultMain"
```

app.d
``` D
import std.stdio;
import std.concurrency : 
    stdReceiveOnly = receiveOnly
  , stdThisTid = thisTid
  , StdTid = Tid
  , stdSend = send
  , spawn;
import vibe.d;

void thread1(StdTid mainThread, StdTid otherThread) {

  writeln("thread1 send");
  otherThread.stdSend(stdThisTid);
  writeln("thread1 wait");
  stdReceiveOnly!StdTid();

  writeln("thread1 exit");
  mainThread.stdSend(true);
}

void thread2(StdTid mainThread) {
  writeln("thread2 wait");
  std.concurrency.receive((StdTid otherThread) {
      writeln("thread2 send");
      otherThread.stdSend(stdThisTid);
    });

  writeln("thread2 exit");
  mainThread.stdSend(true);
}

void test()
{
  auto tid2 = spawn(&thread2, stdThisTid);
	auto tid1 = spawn(&thread1, stdThisTid, tid2);

  stdReceiveOnly!bool();
  stdReceiveOnly!bool();
  writeln("test exit");
}

shared static this()
{
  settings.port = 8080;

  listenHTTP(settings, &handleRequest);
}

void handleRequest(HTTPServerRequest req,
                   HTTPServerResponse res)
{
  test();
  if (req.path == "/")
    res.writeBody("Hello, World!", "text/plain");
}
```

**Vibe-d**: 0.7.27-alpha.1
**dmd**: 2.069.1
**platform**: Fedora 23 x86_64

P.S. Dirty workaround is to run your stuff in separate thread and communicate with a fiber via shared memory.

s-ludwig:


Geod24:
Moving from https://github.com/rejectedsoftware/vibe.d/commit/d8b51c2f00a6d91042869de5d6e16d6380cad861

> I've reverted the changes and will start a branch. I didn't realize that the dependencies had become so deep in the meantime.

Had the same experience a few weeks ago. Also, one question is, are we absolutely certain that the scheduler will be set **first** ? What if the user doesn't import any vibe.d module from his main (he could use `extern` functions, and have the implementation in other modules. Would the shared ctor still run first ?

s-ludwig:
That should be the case, since the module constructor would still be in the dependency chain. Not sure if dynamic libraries work correctly, but since the initialization happens in `vibe.core.core`, otherwise this would not only break the std.concurrency integration, but all vibe.d core functionality.

s-ludwig:
I just went over the source again and all existing places use vibe.d's own "...Compat"  version of the message passing code, so there should be no functional change there. Since the tests are green now, including the message passing tests, I'll merge this now and tag a new pre-release for 0.7.27.

PR #1348, opened 2015-12-08
Busy wait in Task.init I/O causes cpu spike

etcimon:
Avoid a busy wait when running I/O outside fibers.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -145,7 +145,7 @@ final class LibasyncDriver : EventDriver {
 
 	int runEventLoopOnce()
 	{
-		getEventLoop().loop(0.seconds);
+		getEventLoop().loop(100.msecs);
 		processTimers();
 		getDriverCore().notifyIdle();
 		logTrace("runEventLoopOnce exit");
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -134,7 +134,7 @@ final class LibasyncDriver : EventDriver {
 
 	int runEventLoop()
 	{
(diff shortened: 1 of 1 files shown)


s-ludwig:
Does libasync recognize `Duration.max` there`? That would be the correct semantics in this case. 100 ms is functionally fine, too, but it would waste energy/battery if the process is idle.

etcimon:
Yes, I'm only worried if the task does a yield(). That could potentially cause it to freeze. Do you think this should be specifically forbidden just in case?

s-ludwig:
I don't understand, in which case would it freeze? Everytime at least one task calls `yield()` it must be guaranteed that `onIdle` will be called before going to sleep again (i.e. whenever the event queue gets empty). This should avoid any freezes.

etcimon:
I think it froze if yield was called twice in a row, because it won't loop back into notifyIdle, which is why I removed those in favor of manual events in the http/2 driver.

s-ludwig:
But it does loop: https://github.com/rejectedsoftware/vibe.d/blob/1b8c67ce6364d4a3cccba5b2ad3ee828d19bd364/source/vibe/core/core.d#L1192

etcimon:
So I think this must have been a false conclusion I made back when notifyIdle wasn't called in the right places in libasync. I think Duration.max is safe to use in that case

s-ludwig:
OK

s-ludwig:
Now `LibasyncDriver.exitEventLoop` just needs something to nudge the event loop.

s-ludwig:
Seems to be more than that - `dub test :core -c libasync` hangs after "Acquire event ID#32".

etcimon:
Maybe just AsyncNotifier with no handler. Strange that the event would make it hang...

PR #1349, opened 2015-12-08
Exit event loop in libasync using a signal

etcimon:


s-ludwig:
I've taken a closer look and the issue is with the unit test in `vibe.core.sync`, line 303 ff.

`t1` hangs in `sleep(20.msecs);` indefinitely, so that `t1.join()` never returns either.

etcimon:
Yes, the timer isn't being rescheduled because of a mistake I made in the driver. I'm testing it some more submitting a pull request.

etcimon:
Seems like it was a pretty serious issue

s-ludwig:
Looks nasty. Good to have this resolved!

PR #1352, opened 2015-12-15
Allow a serializer to be specified when ignoring an attribute

rjmcguire:


s-ludwig:
What worries me a little is that this connects a certain type to an output format. After all, a JSON serializer could output `Json`, `string`, `std.json : JSON` or any other type. Another potential issue is that it may actually be required to ignore different fields for the *same* format in different situations (e.g. when serializing a structure once for the DB and once for a REST API, but both times to JSON).

So taking that into account, maybe we could instead connect this to a certain serialization policy (`serializeWithPolicy`):

```
struct RESTPolicy {} // no additional customization in this policy

struct MyStruct {
    @ignore!RESTPolicy BsonObjectID _id;
    string foo;
    int bar;
}
```

The policy could maybe also support to encode this outside of the original `struct` declaration:
```
struct RESTPolicy {
    enum isIgnored(S, string fieldname) = is(S == MyStruct) && fieldname == "_id";
}

struct MyStruct {
    BsonObjectID _id;
    string foo;
    int bar;
}
```

This would mean that for it to be useful a few places, such as the MongoDB driver and the REST interface generator, would have to get means to customize the serialization process, but that is a rather important thing to support anyway.

rjmcguire:
On Thu, Dec 17, 2015 at 11:05 AM, Sönke Ludwig <notifications@github.com>
wrote:

> What worries me a little is that this connects a certain type to an output
> format. After all, a JSON serializer could output Json, string, std.json
> : JSON or any other type. Another potential issue is that it may actually
> be required to ignore different fields for the *same* format in different
> situations (e.g. when serializing a structure once for the DB and once for
> a REST API, but both times to JSON).
>
Agreed, it would be an issue if you used the same format in two places and
want different filtering.

> So taking that into account, maybe we could instead connect this to a
> certain serialization policy (serializeWithPolicy):
>
> struct RESTPolicy {} // no additional customization in this policy
>
> struct MyStruct {
>     @ignore!RESTPolicy BsonObjectID _id;
>     string foo;
>     int bar;
> }
>
> The policy could maybe also support to encode this outside of the original
> struct declaration:
>
> struct RESTPolicy {
>     enum isIgnored(S, string fieldname) = is(S == MyStruct) && fieldname == "_id";
> }
>
> struct MyStruct {
>     BsonObjectID _id;
>     string foo;
>     int bar;
> }
>
> This would mean that for it to be useful a few places, such as the MongoDB
> driver and the REST interface generator, would have to get means to
> customize the serialization process, but that is a rather important thing
> to support anyway.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rejectedsoftware/vibe.d/pull/1352#issuecomment-165388925>
> .
>

Nice, that would be better. Would it be a lot more changes in serializer.d
compared to the changes in this PR?
I didn't like having to tag the different Serializer drivers with the type
they output. It was just the shortest path for my needs.

s-ludwig:
See PR #1438.

PR #1357, opened 2015-12-18
make performLocked package(vibe)

9il:
BTW, why not to use `synchronized` instead of `performLocked`?

### DIFF
--- a/source/vibe/core/sync.d
+++ b/source/vibe/core/sync.d
@@ -94,7 +94,7 @@ struct ScopedMutexLock
 		Returns the value returned from $(D PROC), if any.
 */
 /// private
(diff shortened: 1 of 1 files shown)


s-ludwig:
Unfortunately we can't use `package(...)` yet, as the 2.066 frontend is still supported. The reason for not using `synchronized` is the transition to making `synchronized` and `Mutex` `nothrow`, which, due to `TaskMutex` deriving from `Mutex`, caused a lot of difficult to resolve situations, so that we had to switch to an own version of `synchronized` for the time being.

The transition is currently on hold, but until everything is resolved, it is probably a good idea to keep `performLocked`.

s-ludwig:
Okay, 2.066.0 isn't supported anymore. Merging.

PR #1358, opened 2015-12-22
Added HTTP 451 status code and message

andrewbenton:
IETF RFC:
https://tools.ietf.org/html/draft-ietf-httpbis-legally-restricted-status-04

### DIFF
--- a/source/vibe/http/status.d
+++ b/source/vibe/http/status.d
@@ -46,6 +46,7 @@ enum HTTPStatus {
 	requestedrangenotsatisfiable = 416,
 	expectationFailed            = 417,
 	tooManyRequests              = 429,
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'm wondering if the constant should get some kind of prefix/suffix to signal the draft status, so that there won't be any backwards compatibility issues in case of any unexpected numeric constant changes in the final RFC.

andrewbenton:
That makes sense.  It should be pretty solid though since the IESG approved the publication.  The PR could just sit for a while.

s-ludwig:
Just noticed that this is an RFC now: https://tools.ietf.org/html/rfc7725

Merging in.

PR #1361, opened 2016-01-01
Significant latency reduction in Redis

etcimon:
This made my latency go down from ~400ms to ~100ms in the browser when requests required multiple cache subrequests.


s-ludwig:
Sounds fine. `StreamOutputRange` should make sure that data is sent in reasonable chunks, although we might want to extend it so that the buffer size can be configured.

PR #1365, opened 2016-01-06
DirectoryWatcher optimisation

9il:


source/vibe/core/file.d
```diff
@@ -400,13 +402,18 @@ interface DirectoryWatcher {
...
 
 		Returns:
 			If the call completed successfully, true is returned.
 	*/
-	bool readChanges(ref DirectoryChange[] dst, Duration timeout = dur!"seconds"(-1));
+	bool readChanges(
+		Appender!(DirectoryChange[]) app, /+Appender has ref semantic+/
+		Duration timeout = dur!"seconds"(-1));
+	/// ditto
+	bool readChanges(out DirectoryChange[] dst, Duration timeout = dur!"seconds"(-1));
```
9il: `ref` was replaced with `out`, because previous implementations contains `dst.length = 0;`

s-ludwig:
I don't remember how exactly the current API and semantics resulted, but I think we should go for an interface that can work without allocations. An obvious way to do this for the existing API would be to add a call to `dst.assumeSafeAppend();` after the `dst.length = 0;`. More precise semantics would only overwrite `dst` up the its original length.

For a new API I'd either go straight with a range based interface (would have to use a range object), or simply use a scoped callback that accepts each change - probably the latter because it usually means less boilerplate code. The `Appender` solution can then be implemented on top of that.

PR #1366, opened 2016-01-08
Disable 'Bit 0x20' mixed case option in dns resolve of libevent2 driver to make it working like other drivers

soarqin:
I found that not all DNS servers support the 'Bit 0x20' security mechanism (randomize query names to avoid hacking responses).
Libevent2 enables it by default(all other drivers do not enable it) and sometimes fail on resolveHost(), so it is better to disable it by default to avoid such problems.


s-ludwig:
Interesting, didn't know about this. Kind of a pity, but there seems to be no way around it other than a server white list.

PR #1372, opened 2016-01-13
Fix logicalProcessorCount for OSX

carlor:
The current code uses C-style `sizeof` and `NULL`, and declares the wrong function name.


s-ludwig:
Oh, totally forgot to test on OSX afterwards! Thanks!

PR #1377, opened 2016-01-19
parseAttributes rework

burner:
to accept html attributes common in angular2

https://angular.io/cheatsheet


burner:
@s-ludwig I'm not sure why the build fails. The output of travis does not really help

s-ludwig:
There is a flaky test for the libev driver. I'll probably remove the libev driver altogether - it was supposed to stay there to compare against libevent, but since I'm comparing against a navtive implementation now anyway, that doesn't make much sense anymore.

burner:
thank you for the quick answer

PR #1378, opened 2016-01-21
add setsockopt(SO_REUSEPORT) for listening sockets in libevent2/libev drivers

soarqin:
socket option SO_REUSEPORT is introduced on linux 3.9, it is different from SO_REUSEPORT in BSD systems, see man socket(7) for details.

I found that libasync enables it by default but libevent2/libev drivers should set it in vibe.d own drivers.

### DIFF
--- a/source/vibe/core/drivers/libev.d
+++ b/source/vibe/core/drivers/libev.d
@@ -316,6 +316,21 @@ final class LibevDriver : EventDriver {
 			logError("Error enabling socket address reuse on listening socket");
 			return null;
 		}
+		version(Posix){
+			enum{ SO_REUSEPORT = 15 }
+			if( setsockopt(listenfd, SOL_SOCKET, SO_REUSEPORT, &tmp_reuse, tmp_reuse.sizeof) ){
+				do{
+					version(linux){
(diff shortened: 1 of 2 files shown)


s-ludwig:
I think we should add a new flag `TCPOptions.reusePort` for this, due to security and usability considerations for apps that don't use this. Maybe a parallel PR against Druntime to add `SO_REUSEPORT` there is also a good idea, as well as wrapping the `enum` definition in a `static if (!is(typeof(SO_REUSEPORT)))`. Otherwise LGTM!

soarqin:
D cannot check linux kernel version in CTFE so that there is no way to do static check on SO_REUSEPORT.
That's why libasync use such codes to support port reuse.

s-ludwig:
I didn't mean this to check the kernel features, but to avoid compilation errors in the future, when Druntime also adds a definition of `SO_REUSEPORT`.

soarqin:
I see, it seems like we also need to send a PR to libasync for TCPListenOptions.reusePort support

soarqin:
Close it.
I will give a new PR with all libevent/libev/libasync drivers supported as soon as that etcimon/libasync#42 is merged

PR #1379, opened 2016-01-21
add kernel-level distribution support(SO_REUSEPORT) for linux

soarqin:
libasync does not have setOption for AsyncTCPListener so I do think it is impossible to support TCPListenOptions.reusePort in libasync driver at present.
Therefore I only add it to libevent/libev drivers.

### DIFF
--- a/source/vibe/core/drivers/libev.d
+++ b/source/vibe/core/drivers/libev.d
@@ -317,6 +317,7 @@ final class LibevDriver : EventDriver {
 			return null;
 		}
 		version(linux) {
+			import vibe.core.drivers.utils;
 			if (setsockopt(listenfd, SOL_SOCKET, SO_REUSEPORT, &tmp_reuse, tmp_reuse.sizeof)) {
 				// ignore invalid and not supported errors
 				if (errno != EINVAL && errno != ENOPROTOOPT) {
--- a/source/vibe/core/drivers/libev.d
+++ b/source/vibe/core/drivers/libev.d
@@ -316,6 +316,15 @@ final class LibevDriver : EventDriver {
 			logError("Error enabling socket address reuse on listening socket");
 			return null;
 		}
+		version(linux) {
+			if (setsockopt(listenfd, SOL_SOCKET, SO_REUSEPORT, &tmp_reuse, tmp_reuse.sizeof)) {
+				// ignore invalid and not supported errors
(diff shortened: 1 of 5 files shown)


etcimon:
It's only fair that it be enabled by default, I don't see any web server going without it. Which is why I re-enabled it by default in libasync directly.

soarqin:
@etcimon
1. port reuse is restricted to same user, setsockopt would fail if listen to the same port from another user
2. if the socket listend earlier was not set SO_REUSEPORT, setsockopt would fail
These 2 failures are not the same as that of listen errors without port reusing, although it is okay if we don't care about different types of socket errors

s-ludwig:
It's also important to keep in mind that not all servers are HTTP servers, and in practice there are a lot of HTTP servers, too, which get along without request distribution and which wouldn't work even correctly in that case. I'd instead prominently document this approach.

BTW, it probably makes sense to deprecate the (unsafe) `TCPOptions.distribute` and `HTTPServerOptions.distribute` options. `runWorkerTask({ listenHTTP(...); })` together with `SO_REUSEPORT` would then be the direct replacement.

etcimon:
> if the socket listend earlier was not set SO_REUSEPORT, setsockopt would fail
I'm not sure I understand correctly. Which setsockopt command would fail?

> in practice there are a lot of HTTP servers, too, which get along without request distribution and which wouldn't work even correctly in that case. I'd instead prominently document this approach.

Agree, it will have to be documented a bit more. Is there a downside to having it on by default though? The only difference is that the application won't crash at start if another one was already listening

soarqin:
@etcimon  If first socket on the port was not set SO_REUSEPORT, try to set it on the later ones would fail

s-ludwig:
It can lead to unexpected and hard to detect behavior if a previous program instance keeps running in the background and may lead to data corruption if the application isn't ready for concurrency. For example it may lose data when doing a read-modify-write cycle for on-disk data without locking. I'd just want to keep enabling non-obvious behavior explicit. If you need to scale a service, you have to put work into that and adding an `options |= HTTPServerOption.reusePort;` isn't really worth mentioning.

etcimon:
True, I'll disable the default for the next version =)

soarqin:
I removed reusePort from default options, wait for new release of libasync then.

s-ludwig:
Okay, thanks! I'll merge now (the libasync driver can be adjusted later).

PR #1380, opened 2016-01-22
Libevent read buffer

s-ludwig:
This makes `Libevent2TCPConnection` use a `FixedRingBuffer` based read buffer instead of the one built into libevent, which is considerably slower (~18% slower when tested against a simple HTTP server). Currently, the libevent buffer still exists and removing it requires further work (using an `event*` instead of a `bufferevent*`).


s-ludwig:
Merging to include this in the beta.

PR #1385, opened 2016-01-28
Implement Diet template caching.

s-ludwig:
This is the implementation of an idea by Martin Tschierschke for speeding up Diet template compilation. This allows downstream projects to define `-version=VibeCacheDietTemplates` to let the compiled Diet template (D code) be cached in the "views" folder, so that it gets picked up on the next compiler run instead of recompiling the template again. This can vastly improve compile time and memory use during edit-compile-run cycles (~5x speedup for the templates of vibed-org).

Note that the cached source code is written at application initialization time. The program must be run from the root of the package directory, so that the cached files get written into the right "views" folder. [DIP81](http://wiki.dlang.org/DIP81) could avoid this shortcoming (even better would of course be a rewrite of the CTFE engine, which would remove the need for such workarounds altogether).

### DIFF
--- a/source/vibe/templ/diet.d
+++ b/source/vibe/templ/diet.d
@@ -257,7 +257,7 @@ private template dietParser(string template_file, int base_indent, TRANSLATE...)
 	}
 
 	enum hash = computeTemplateHash();
-	enum cache_file_name = "-cached-"~template_file~"-"~hash.to!string~".d";
+	enum cache_file_name = "_cached_"~template_file~"_"~hash.to!string~".d";
 
 	static if (use_cache && is(typeof(import(cache_file_name)))) {
 		enum dietParser = import(cache_file_name);
--- a/source/vibe/templ/diet.d
+++ b/source/vibe/templ/diet.d
@@ -71,9 +71,9 @@ void compileDietFileIndent(string template_file, size_t indent, ALIASES...)(Outp
 	// Generate the D source code for the diet template
 	//pragma(msg, dietParser!template_file(indent));
 	static if (is(typeof(diet_translate__)))
-		mixin(dietParser!(template_file, diet_translate__)(indent));
+		mixin(dietParser!(template_file, indent, diet_translate__));
 	else
-		mixin(dietParser!template_file(indent));
+		mixin(dietParser!(template_file, indent));
 }
 
 /// compatibility alias
@@ -110,7 +110,7 @@ void compileDietFileCompatV(string template_file, TYPES_AND_NAMES...)(OutputStre
 
 	// Generate the D source code for the diet template
 	//pragma(msg, dietParser!template_file());
-	mixin(dietParser!template_file(0));
+	mixin(dietParser!(template_file, 0));
 }
 
 /// compatibility alias
@@ -147,7 +147,7 @@ template compileDietFileMixin(string template_file, string stream_variable, size
 		"import vibe.stream.wrapper;\n" ~
 		"OutputStream stream__ = "~stream_variable~";\n" ~
 		"auto output__ = StreamOutputRange(stream__);\n" ~
-		dietParser!template_file(base_indent);
+		dietParser!(template_file, base_indent);
 }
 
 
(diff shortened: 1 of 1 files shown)


EmTee70:
Hint: The naming of the cached "-.....dt....d" files starting with - (minus) is not so good, because you need special quoting to open them even "ls *"  breaks.     ls: Ungültige Option -- e

s-ludwig:
Good, point. I'll switch to an underscore.

EmTee70:
Now a newbie question, I am using dub.sdl and tried to place 
versions "VibeDefaultMain"
versions "VibeCacheDietTemplates" 
what may be clear dos not work, but how to put a list in .sdl?
There was a moment when the -...template...dt...d files where created, so I could give the hint, but the linking failed.  Or in an other way how to call -version=VibeCacheDietTemplates ?

EmTee70:
Looking in the code, I tried to understand the hashing mechanism you are using. Why not comparing 
max(modification_datetime(all_files)) with modification_datetime(cached_file) and than decide?
It seams to me, that by every change there are created new _cached-..HASH...d-files filling up the directory?

s-ludwig:
Both,
    versions "VibeDefaultMain"
    versions "VibeCacheDietTemplates"
and
    versions "VibeDefaultMain" "VibeCacheDietTemplates"
should work. If indeed it doesn't, there must be a bug in the SDLang package recipe parser. Which were the linker errors that you got?

s-ludwig:
> Looking in the code, I tried to understand the hashing mechanism you are using. Why not comparing
max(modification_datetime(all_files)) with modification_datetime(cached_file) and than decide?
It seams to me, that by every change there are created new _cached-..HASH...d-files filling up the directory?

There are two reasons:
- The modification time is not available at compile-time, only the file contents can be read with `import(filename)`
- It's possible that the same template file is instantiated with different parameters, causing the generated code to be different. In the current implementation that's only the indentation level, but it should also include the translation context any maybe other parameters.

The side-effect that the folder fills up with files over time is kind of unfortunate, but maybe that can be worked around. Inside of the generated `shared static this` module constructors, instead of directly creating the files, a list of files could be assembled, and then all files not in that list would be removed from the "views/" folder. Now the only difficulty is to determine when the list is actually complete.

EmTee70:
Thank you. Ok, when using the syntax you mentioned **it compiles, but the parameter is ignored**,
when omitting the VibeDefaultMain-Part than the_cache...dt..d files are created, but the missing of main is complained, what seams logical: 
       /usr/lib/gcc/x86_64-linux-gnu/4.8/../../../x86_64-linux-gnu/crt1.o(.text+0x20): error: undefined reference to 'main'

s-ludwig:
Strange, both work for me. There is even a DUB [test case](https://github.com/D-Programming-Language/dub/blob/c98dd9e9739665863dd4b1e090d5e0dda3234b8e/source/dub/recipe/sdl.d#L382) for this. You could run `dub build -v` to see what flags actually end up on the command line, maybe something else goes wrong.

EmTee70:
One Error:
Found dependency vibe-d:core ~cache-diet-templates
Version selection for dependency botan (botan) of vibe-d:core is missing.

command line:
dmd -c -of.dub/build/application-debug-linux.posix-x86_64-dmd_2069-AA667810FA06613A16AD14A86F12084A/mysql-browse.o -debug -g -w -version=VibeCacheDietTemplates -version=VibeDefaultMain -version=VibeLibeventDriver -version=Have_mysql_browse -version=Have_mysql_d -version=Have_vibe_d -version=Have_vibe_d_core -version=Have_libevent -version=Have_openssl -version=Have_vibe_d_data -version=Have_vibe_d_utils -version=Have_vibe_d_diet -version=Have_vibe_d_http -version=Have_vibe_d_mail -version=Have_vibe_d_mongodb -version=Have_vibe_d_redis -version=Have_vibe_d_web -Isource/ -I../../../.dub/packages/mysql-d-0.3.2/source/ -I../../../.dub/packages/vibe-d-cache-diet-templates/source/ -I../../../.dub/packages/libevent-2.0.1_2.0.16 -I../../../.dub/packages/openssl-1.1.4_1.0.1g -Jviews source/app.d ../../../.dub/packages/vibe-d-cache-diet-templates/source/vibe/appmain.d -vcolumns

???

EmTee70:
But: Your additional hint to **speed up compilation by using gold linker** was great. Took me some time, but in the end I changed the link from /usr/bin/ld to point to ld.gold, linking time halved. 
sudo apt-get install binutils-gold did not hat a effect, exclaiming that no changes are performed, because something was set up by hand. (?)

s-ludwig:
The command line looks alright and the "botan" notice shouldn't matter. Either there is something really strange going on with the intermediate build results, or something about the results is simply mixed up.

One thing that doesn't fit is that since the cached files are only created once the application is actually *executed*, it is impossible that it would work without the "VibeDefaultMain", since the executable wasn't even built in that case. My cautious guess what actually happened would be that the executable from the first build with both versions ran and that it actually worked correctly.

EmTee70:
When i remove all -cache..dt.d files and use both version flags, the program compiles fine,
but no new cache files are written
Without VibeDefaultMain, the cache files are created, but than no linking.

s-ludwig:
But did you actually run the program after deleting the files? They won't be created during the build process.

EmTee70:
Hello Sönke, do you know that sitting to long behind the monitor at night sometimes produces strange results :-) Sorry, I have not yet woken up.......

s-ludwig:
The "correct" way to switch the linker, I think, is to use `update-alternatives`: `update-alternatives --install "/usr/bin/ld" "ld" "/usr/bin/ld.gold" 20`. It should also be possible to change it only for DMD using /etc/dmd.conf, but I'm not sure how.

s-ludwig:
> Hello Sönke, do you know that sitting to long behind the monitor at night sometimes produces strange results :-) Sorry, I have not yet woken up.......

Know that too well! ;-)

EmTee70:
Because my mini project compiles quite fast, I tried to compare compile times with 
time dub build and than suddenly the sun was rising :-).
What about an special output from dub  
when compiling .. template ... "using cached file",
my initial idea was to import the _cache_.dt.d files as small modules, 
so even the 
Compiling diet template 'index.dt'... 
would not occur but something like _cache_...index.dt....d is up to date?

Is this theoretically possible?

s-ludwig:
It's a little difficult, because the caching happens at a lower level than the code that outputs the "Compiling ..." message. The message could be moved to the lower level, too, but then there would also be messages when included templates get compiled. It could probably be solved by passing an additional boolean template parameter, though.

rjmcguire:
Just for future reference;  the correct flag for dub with DMD compiler is:
 lflags "-fuse-ld=gold"

s-ludwig:
This is now superseded by https://github.com/rejectedsoftware/diet-ng/commit/fa5f9e54747804201431850299a2f1263d1b52e3 (i.e. implemented in diet-ng) - it doesn't make much sense to put more work into the old implementation at this point.

PR #1390, opened 2016-02-01
Fix ManualEvent Leak/Range Error

etcimon:
This fixes a potential range error that might have gone undetected for a while.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -859,7 +859,7 @@ final class LibasyncManualEvent : ManualEvent {
 
 		bool signal_exists;
 
-		if (s_eventWaiters.length <= m_instance) {
+		if (s_eventWaiters.length <= m_instance + 1) {
 			expandWaiters();
 		}
 		logTrace("Acquire event ID#%d", m_instance);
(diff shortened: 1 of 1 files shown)


etcimon:
Actually, I don't think this fixes anything. Not what I was looking for anyways

etcimon:
Well, it seems to have a positive effect overall on the server load for some reason so I'm reopening

source/vibe/core/drivers/libasync.d
```diff
@@ -893,7 +893,7 @@ final class LibasyncManualEvent : ManualEvent {
 	bool amOwner()
 	{
 		import std.algorithm : countUntil;
-		if (s_eventWaiters.length <= m_instance) return false;
+		if (s_eventWaiters.length <= m_instance + 1) return false;
```
s-ludwig: Wouldn't this cause `amOwner` to always return `false` for the latest `ManualEvent` that was created?

source/vibe/core/drivers/libasync.d
```diff
@@ -859,7 +859,7 @@ final class LibasyncManualEvent : ManualEvent {
 
 		bool signal_exists;
 
-		if (s_eventWaiters.length <= m_instance) {
+		if (s_eventWaiters.length <= m_instance + 1) {
```
s-ludwig: And this should cause `expandWaiters` to be called repeatedly for every `acquire` call on the latest `ManualEvent`.

etcimon:
Yes you're right, this isn't the fix I was looking for. I'm getting a range violation strangely in doWait after a lot of traffic has gone through, so there's probably something wrong somewhere in here.

PR #1392, opened 2016-02-02
Change counter used for BsonObjectID to get initialized with a random…

machindertech:
… value as per the documentation

See https://docs.mongodb.org/manual/reference/object-id/#overview


s-ludwig:
Thanks, LGTM! The LDC test failure should be unrelated.

PR #1393, opened 2016-02-02
Fixed build error in release mode

machindertech:


Geod24:
The autotester's not too happy about it. What was the issue in release ?

s-ludwig:
Yes, the error message would be interesting, as well as possibly a test case that can be added to the Travis-CI script. It currently calls `dub build -b release --combined`, but maybe that template doesn't get instantiated.

machindertech:
We were getting the following error when building our project in 64-bit release mode on Linux:
../../dub/vibe.d/source/vibe/internal/meta/traits.d(153,31): Error: variable t used before set

s-ludwig:
Wow, February, this got totally lost. I see no issue in merging this. Will be in 0.7.30.

PR #1394, opened 2016-02-02
Fix Bson.opt not working with Bson[]

machindertech:


s-ludwig:
Pushed a modified fix which keeps calling `opt` on `const(Bson)` working and adds a unit test.

PR #1396, opened 2016-02-02
Fix RedisReply leaving connection in unusable state if an exception o…

machindertech:
…ccurs during initialization

This is required because if an exception is thrown in a constructor, the
destructor will not be called.

### DIFF
--- a/source/vibe/db/redis/redis.d
+++ b/source/vibe/db/redis/redis.d
@@ -1273,6 +1273,7 @@ struct RedisReply(T = ubyte[]) {
 		auto ctx = &m_conn.m_replyContext;
 		assert(!ctx.initialized);
 		ctx.initialized = true;
+		scope (failure) drop();
 
 		auto ln = cast(string)m_conn.conn.readLine();
 


s-ludwig:
AFAICS, `drop()` would only do something if reading a multi-reply failed. But the only way that could throw an exception is if the count is not a valid `long` string. But then then the reply length would still stay `0` and `drop` still wouldn't do anything. 

I think the correct thing to do would be to instead close the connection, but only for the two cases of "unknown reply" and an invalid multi-reply length string. What do you think?

machindertech:
The issue we were having was that when an exception is thrown, the ctx.refCount stays incremented and then the next RedisReply fails on the assert(ctx.refCount == 0).  Any kind of exception in initialize() will cause the assert to be triggered for the next reply.

s-ludwig:
Can you try the [issue1412_redis_reply_exception](https://github.com/rejectedsoftware/vibe.d/tree/issue1412_redis_reply_exception) branch? I've more or less implemented the approach outlined above there.

machindertech:
The issue1412_redis_reply_exception branch seems to fix the issues that we were having, so we're closing this pull request.

PR #1397, opened 2016-02-02
Fix 32-bit compile error

machindertech:


Geod24:
Shouldn't the parameter be made `size_t` instead ? A wild cast without any check is not that good.

s-ludwig:
Correct fix is in 3f3d14e961fea6b4476868b264138d233d24447f

PR #1398, opened 2016-02-02
Change Redis eval and evalSHA functions to allow a const array for keys

machindertech:


s-ludwig:
Unfortunately this resulted in totally obscure linker errors, so I had to revert it: https://travis-ci.org/rejectedsoftware/vibe.d/jobs/108150891

Using `scope const(string)[]` didn't solve it either. Unfortunately, I don't get those linker errors locally, so I can't even run dustmite.

PR #1399, opened 2016-02-02
[Libasync] ManualEvent - Fix Race Condition and Optimize Space Usage

etcimon:
The allocations were inconsistent with binary increments and were hard to debug. This fixes it.

### DIFF
--- a/source/vibe/core/drivers/libasync.d
+++ b/source/vibe/core/drivers/libasync.d
@@ -805,20 +805,22 @@ final class LibasyncManualEvent : ManualEvent {
 		Array!(void*) ms_signals;
 
 		core.sync.mutex.Mutex m_mutex;
+
+		@property size_t instanceID() nothrow { return atomicLoad(m_instance); }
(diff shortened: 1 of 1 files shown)


etcimon:
Finally found the issue. Race conditions caused connections to fail unexpectedly

s-ludwig:
Are those `+1`/`-1` operations stil relevant, then? They look kind of strange in code (at least in those places), although they seem to be semantically correct.

etcimon:
They were originally intended to start the offsets at 0 for s_eventWaiters, which allows it to occupy space in factors of 2 which optimizes the use of the array container. Other than that, it's not very relevant I guess.

s-ludwig:
I'd just usually do such arithmetic within `generateID` and `recycleID`, so that it is close together in a central place. With only a single call to each function, it doesn't matter that much in this case, but usually the current form would be relatively error prone.

s-ludwig:
BTW, there is a `nothrow` related compilation error on Travis.

s-ludwig:
Looks good!

Issue #1388, opened 2016-01-30
Bad Request using [::1]


srpelissier:
From bash prompt:

```
dub init
dub

Performing "debug" build using dmd for x86_64.
vibe-d 0.7.26: target for configuration "libevent" is up to date.
test ~master: target for configuration "application" is up to date.
To force a rebuild of up-to-date targets, run again with --force.
Running ./test 
Listening for requests on http://::1:8080
Listening for requests on http://127.0.0.1:8080
Please open http://127.0.0.1:8080/ in your browser.
```

When opening [::1]/8080 in Firefox, instead of _Hello, world_, the following occurs:

```
400 - Bad Request

Bad Request

Internal error information:
std.conv.ConvException@/usr/local/include/std/conv.d(2002): Unexpected end of input when converting from type string to type uint
----------------
/usr/local/include/std/conv.d:58 pure @safe uint std.conv.parse!(uint, immutable(char)[]).parse(ref immutable(char)[]) [0x844034]
/usr/local/include/std/conv.d:1934 pure @safe ushort std.conv.parse!(ushort, immutable(char)[]).parse(ref immutable(char)[]) [0x843e6c]
/usr/local/include/std/conv.d:1723 pure @safe ushort std.conv.toImpl!(ushort, immutable(char)[]).toImpl(immutable(char)[]) [0x7d8928]
/usr/local/include/std/conv.d:298 pure @safe ushort std.conv.to!(ushort).to!(immutable(char)[]).to(immutable(char)[]) [0x7d88f3]
../../../.dub/packages/vibe-d-0.7.26/vibe-d/source/vibe/http/server.d:1598 bool vibe.http.server.handleRequest(vibe.core.stream.Stream, vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo, ref vibe.http.server.HTTPServerSettings, ref bool) [0x81c18d]
../../../.dub/packages/vibe-d-0.7.26/vibe-d/source/vibe/http/server.d:1487 void vibe.http.server.handleHTTPConnection(vibe.core.net.TCPConnection, vibe.http.server.HTTPListenInfo) [0x81b905]
[...]

PR #1402, opened 2016-02-05
Better parsing of host/port

cyisfor:
One nitpick I have with vibe is that it can't parse IPv6 addresses in urls (because of colons between []). The parsing of that stuff was kind of clunky too, using nonspecific array splits and trying to examine those to see if there's a good host/port. So, I rewrote the host/port parser. Now it does IPv6 addresses just fine. Verified it with an app of my own. I haven't tested against a huge comprehensive list of all sorts of possible URLs, but that'd be a task in of itself I think. 

IPv6 addresses can have variable amounts of colons, and numbers between colons are often optional too, which is why those difficult to parse addresses are surrounded by [] in URLs, so all I had to do was search for ].

tl;dr with this patch, if you have an IPv6 address, then http://[1:2::2::33]:45/ should work now.

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -1555,6 +1555,75 @@ private void handleHTTPConnection(TCPConnection connection, HTTPListenInfo liste
 	logTrace("Done handling connection.");
 }
 
+struct HostPort {
+	string host;
+	ulong port;
+};
+
+HostPort parseHostPort(string s) {
+	HostPort res;
+	const(char)[] arr = s.to!(const(char)[]);
+	enum {
+		BEFORE_BRACKET,
+		IN_BRACKET,
+		AFTER_BRACKET
+	}
+	auto state = BEFORE_BRACKET;
+	ulong bracket, colon, close_bracket;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Basically looks good, but I'd like to make this a bit more compact - the `import` and the `struct HostPort` definition could go into `parseHostPort` to make it more self-contained. You are using a manual state machine, is there a reason why using `find` or `indexOf` wouldn't work? It would squash everything together considerably:
```
if (s.startsWith('[')) {
    auto idx = s.indexOf(']');
    enforce(idx > 0, "Missing closing ']' for IPv6 address.");
    ret.host = s[1 .. idx];
    s = s[idx+1 .. $];
} else {
    auto idx = s.indexOf(':');
    if (idx < 0) idx = s.length;
    enforce(idx > 0, "Missing host name.");
    ret.host = s[0 .. idx];
    s = s[idx .. $];
}
if (s.startsWith(':')) ret.port = s[1 .. $].to!ushort;
```

Corresponding issue: #1388

PR #1404, opened 2016-02-05
Bson: Adds std.uuid.UUID support

denizzzka:


s-ludwig:
Do you happen to know the binary format that is required for the new UUID type `0x4`? The bson module still uses the old code `0x3`, which seems to be defined as an application specific representation (so this PR is fine in principle), but I'm wondering if we shouldn't instead directly opt for `0x4` instead of producing legacy data that results in possibly non-interoperable database contents.

denizzzka:
Plan for this PR:

* Remove conversion of std UUIDs to old 0x3 type
* Add conversion of std UUID to new 0x4 type
* Both 0x3 and 0x4 types of Bson UUIDs should be readable to std UUIDs

Sounds good?

denizzzka:
Or just merge it, and after we can insensibly make changes in the future.

s-ludwig:
If you are willing/have the time to do the former, I'd definitely prefer that, but since there are at least no backwards compatibility issues with doing the latter first, I see no real reason to generally rule it out either. The only downside is that it generates data with an obsolete format.

denizzzka:
Not obsolete, just old

s-ludwig:
Ok, let's say "deprecated" ;)

denizzzka:
It is better for nothing

Also, I am interested in 0x4 UUID support and will implement it later. Promise! :з

denizzzka:
It is need to add also toJson conversion for Bson UUIDs?

s-ludwig:
Hm, right, I didn't think about this. It would definitely be nicer to emit a normal UUID string instead of base64 encoding the binary representation. It also needs to be handled in `BsonSerializer` (`writeValueH`, `readValue` and `getBsonTypeID`).

WebFreak001:
@s-ludwig I think that such a change could be done in a separate PR, no need to let this rot for another year

denizzzka:
PR valid and can be merged, CI accidentally failed and can be restarted.

s-ludwig:
Okay thanks! Makes sense as a step forward.

PR #1417, opened 2016-02-11
Fix #1310: dub test fails

### LINKED ISSUE #1310: dub test fails
With my current project 'dub test' with vibe v0.7.25 passes 
But with the new v0.7.26-beta.1 'dub test' fails with
../../../../../tmp/dub_test_root-f035dc5c-69b6-464f-a027-658a95093915.d(17,12): Error: only one main allowed

The contents of said file are here http://pastebin.com/URfR3Evg

jacob-carlborg:
Make sure "VibeDefaultMain" is not set when "version(unittest)"
is set.

### DIFF
--- a/source/vibe/appmain.d
+++ b/source/vibe/appmain.d
@@ -20,6 +20,8 @@
 */
 module vibe.appmain;
 
+version (unittest){}
+else
 version (VibeDefaultMain):
 
 /**


s-ludwig:
Usually `dub test` should automatically exclude the "appmain.d" file from the build (because it's the "mainSourceFile"). I've just tested this with a trivial test application and it worked correctly there. Do you have a failing test case?

jacob-carlborg:
```
dub init bar -t vibe.d && cd bar && touch source/foo.d && dub test
```
Dub: DUB version 0.9.24, built on Sep 20 2015
vibe.d: 0.7.27

jacob-carlborg:
The comment you wrote in #1310 seems to still be an issue: https://github.com/rejectedsoftware/vibe.d/issues/1310#issuecomment-152432102. Dub does not include `mainSourceFile` from the project I'm building but it does include the one from `vibe.d`. Perhaps this is an issue with Dub.

This is the output of `dub test --verbose`:

```
dmd -c -of.dub/build/__test__library__-unittest-posix.osx-x86_64-dmd_2070-97A975557F90C54E09BB7A9AA3EA7951/__test__library__.o -debug -g -unittest -w -version=VibeDefaultMain -version=VibeCustomMain -version=VibeLibeventDriver -version=Have_bar -version=Have_vibe_d -version=Have_vibe_d_core -version=Have_libevent -version=Have_openssl -version=Have_vibe_d_data -version=Have_vibe_d_utils -version=Have_vibe_d_diet -version=Have_vibe_d_http -version=Have_vibe_d_mail -version=Have_vibe_d_mongodb -version=Have_vibe_d_redis -version=Have_vibe_d_web -Isource/ -I../../../.dub/packages/vibe-d-0.7.27/source/ -I../../../.dub/packages/libevent-2.0.1_2.0.16 -I../../../.dub/packages/openssl-1.1.4_1.0.1g -Jviews ../../../../../var/folders/6l/cmgd0cb51_j_bhxzwks_w99r0000gn/T/dub_test_root-5b99d558-11ae-4b21-b448-67641eb7c558.d source/foo.d ../../../.dub/packages/vibe-d-0.7.27/source/vibe/appmain.d -vcolumns
../../../../../var/folders/6l/cmgd0cb51_j_bhxzwks_w99r0000gn/T/dub_test_root-5b99d558-11ae-4b21-b448-67641eb7c558.d(9,12): Error: only one main allowed
FAIL .dub/build/__test__library__-unittest-posix.osx-x86_64-dmd_2070-97A975557F90C54E09BB7A9AA3EA7951/ __test__library__ executable
dmd failed with exit code 1.
```

At the end of the first line you can see `../../../.dub/packages/vibe-d-0.7.27/source/vibe/appmain.d`.

jacob-carlborg:
To reproduce this issue it's necessary to have at least one file in `source` except for `app.d`.

s-ludwig:
Okay, I see now. This is basically a general issue with the `VibeDefaultMain` mechanism. It also happens if you try to use the package as a library target and want to define your own main in the dependent package. The current solution is to define the configurations manually:
```
configuration "application" {
    targetType "executable"
    versions "VibeDefaultMain"
}
configuration "library" {
    targetType "library"
}
```

This has bothered be for a while now and I'm not yet sure how to best improve this in general. The fix using `version(unittest)` is at least an improvement for `dub test`, but it doesn't solve the library issue and at the same time it can break cases where using vibe.d's main is actually desired (e.g. `dub build -b unittest`).

Any ideas?

jacob-carlborg:
> Any ideas?

* Change Dub to ignore `mainSourceFile` for all dependencies unless building an executable
* Will adding the above configurations to vibe.d work?

jacob-carlborg:
* Generate the above configurations when creating a new vibe.d project

schveiguy:
Just ran into this: https://forum.rejectedsoftware.com/groups/rejectedsoftware.dub/thread/8185/

IMO, this looks like an issue with dub. If dub's default vibe.d initialization does not allow me to write unit tests, then something is wrong.

jacob-carlborg:
@s-ludwig I've found some more issues with this [1] . This needs to be solved. Could we merge this for now until we can come up with a better solution?

[1] https://forum.rejectedsoftware.com/groups/rejectedsoftware.dub/post/8306

jacob-carlborg:
@s-ludwig What's the original idea behind using a shared module constructor and a main function inside vibe.d?

Why doesn't the user supply the main function, which calls a function in vibe.d? Something like this:

```d
import vibe.d;

void hello(HTTPServerRequest req, HTTPServerResponse res)
{
    res.writeBody("Hello, World!");
}

int main()
{
    settings.port = 8080;
    listenHTTP(settings, &hello);

    logInfo("Please open http://127.0.0.1:8080/ in your browser.");

    return runVibe(); // new function which does whatever the main function in vibe.d is doing now
}

```

s-ludwig:
The idea was to hide any explicit setup calls, to make it impossible to misuse and to appeal to scripting language framework users. Using `runEventLoop()` instead of `runVibe()` works already, except for some extra bits. But I think providing a `runVibe()`-like function actually makes sense now that `VibeDefaultMain` is opt-in.

(I'll have a proper look at this issue complex later today)

jacob-carlborg:
Could this be merged as a temporary workaround until we can find a better solution. Or would cause any other problems?

schveiguy:
I have to say, I find the way vibe.d library defines the main function very off-putting. You are already creating an app.d file which sets up the default site (with dub init). Why not have the app.d file contain the main function? Most other frameworks I have used (Visual C++, ios, etc.) generate an equivalent file that "starts" the program, which you can then tinker with or override.

s-ludwig:
Using this change set would mainly break `dub build -b unittest` or similar ways to perform a unittest build. I agree that multiple points mentioned so far make sense:
- Let DUB avoid to build vibe.d's main function - either by removing the mainSourceFile from the build, or by removing the VibeDefaultMain version. I currently lean towards the latter as a quick and safe (intermediate) fix, because the semantics of mainSourceFile are more complex
- Implement a single API entry point as an alternative to vibe.d's default main function
- Change the default vibe.d app template to output the two separate configurations above - or a main function instead of the module constructor

I'm not particularly keen on keeping the default main mechanism, because most of the reasons for its introduction don't exist anymore. The only potential benefit is that this takes the burden to define platform specific main functions off from the application developer (`WinMain` or the WinRT main function), but there are arguably better places for that.

s-ludwig:
Okay, two of three points are now implemented. Adjusting the default DUB project will just have to wait until a new vibe.d release is out.

jacob-carlborg:
Seems to be working 👍. Thanks for the fix. Closing this now since it's obsolete.

PR #1428, opened 2016-02-15
don't split log lines on LF

MartinNowak:
- it might make sense to have newlines in a log message
  (e.g. when printing a backtrace) and they shouldn't
  be split into multiple log messages, e.g. when using syslog
- the splitting was poorly done anyhow (only on the first LF)

### DIFF
--- a/source/vibe/core/log.d
+++ b/source/vibe/core/log.d
@@ -195,6 +195,12 @@ struct LogLine {
 class Logger {
 	LogLevel minLevel = LogLevel.min;
 
+	/** Whether the logger can handle multiple lines in a single beginLine/endLine.
+
+	   By default log text with newlines gets split into multiple log lines.
+	 */
+	protected bool multilineLogger = false;
+
 	private {
 		LogLine m_curLine;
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'd like to make this change optional - for example by adding a `@property bool multilineLogger` to `Logger`, which defaults to return `false`. Especially the HTML logger wouldn't work well with multiple lines per "line", as well as a custom logger that I have in an external code base (I could of course easily adapt it, but maybe there are more `Logger`s in the wild).

Interesting that the bug with splitting only at the first line separator slipped through so long. I never noticed that in the log output.

MartinNowak:
made it configurable

s-ludwig:
Okay, thanks! Also for fixing the line splitting along the way. The only issue seems to be that `splitter().front` is called unconditionally and fails for empty input strings.

MartinNowak:
Fixed.
I didn't care about splitting on CR and what not.

s-ludwig:
Yeah, I think it's completely OK to require the user to use UNIX line endings here. LGTM.

Issue #229, opened 2013-05-28
serveStaticFiles and URLRouter can't handle encoded URLs


MartinNowak:
For example the following request isn't match.

```
GET /foo%20bar
```

PR #1436, opened 2016-02-21
Do not URL-decode `req.path`! Decode `req.params` instead

AndreyMZ:
The commit fixed the following problem.

**Problem description:**
It is wrong to URL-decode `req.path`.

**STR:**
1.  Assume the following code:
   
   ``` d
   class MyWebInterface
   {
       @path("/items/:ITEM_ID")
       void getItem(HTTPServerRequest req, HTTPServerResponse res)
       {
           string itemId = req.params["ITEM_ID"];
           assert(itemId == "foo/bar");
           res.writeBody(itemId);
       }
   }
   ```
2.  Make request `http://localhost/items/foo%2Fbar`

**Actual result:**

```
404 - Not Found

Not Found

Internal error information:
No routes match path '/items/foo%2Fbar'
```

Because actually router tries to match path `"/items/foo/bar"`

**Expected result:**
`itemId` equals to `"foo/bar"`.

### DIFF
--- a/http/vibe/http/router.d
+++ b/http/vibe/http/router.d
@@ -263,7 +263,7 @@ final class URLRouter : HTTPServerRequestHandler {
 				if (r.method != method) return false;
 
 				logDebugV("route match: %s -> %s %s %s", req.path, r.method, r.pattern, values);
-				foreach (i, v; values) req.params[m_routes.getTerminalVarNames(ridx)[i]] = v;
+				foreach (i, v; values) req.params[m_routes.getTerminalVarNames(ridx)[i]] = urlDecode(v);
 				if (m_computeBasePath) req.params["routerRootDir"] = calcBasePath();
 				r.cb(req, res);
 				return res.headerWritten;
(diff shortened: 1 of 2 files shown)


s-ludwig:
This would be a breaking change if done as-is. The `req.path` was changed to be URL-decoded specifically to address an earlier issue: #229

A backwards-compatible possibility to get this behavior would be to add a `@property bool matchRawPath` to `URLRouter` and let it match `req.pathString` instead, if set to true.

AndreyMZ:
The description of #229 is poor: ther is no steps, actual result, expected result. I cannot understand what was the problem? Do you know more details about it than is written?

s-ludwig:
It was basically the following:
```
router.get("/foo bar/", ...);
```
Expected: The above route should be matched if a request `GET /foo%20bar` is made to the router.
Actual: It doesn't match the route, but it does match `router.get("/foo%20bar", ...);` instead.

The same situation was true for the file server: It would have looked for a file named "foo%20bar" instead of "foo bar"

---

I think at this point there are two possible ways to improve the situation:

1. Make the behavior proposed here optional - this in the end just provides two suboptimal solutions (one has the issue that "/" cannot be part of a path element and the other has the issue that the matching depends on how the client chooses to encode the characters).
2. Decode the URL in two steps, first everything but "/", then match, then decode "/" within the matched path elements - this appears to be the "correct" solution, if there is one, but it will be difficult to implement in an efficient way.

WebFreak001:
Maybe instead also encode the url (except for `/`) inside the router.match functions? Then my PR I made would actually fix all of the noted issues.

s-ludwig:
Closing in favor of the discussion in #1635.

PR #1438, opened 2016-02-23
Make all serialization attributes specifiable per policy. Fixes #1352.

s-ludwig:
The attributes `@ignore`, `@optional`, `@name`, `@byName` and `@asArray` now take a serialization policy as an optional template argument. With that specified, they only have an effect if `(de)serializeWithPolicy` is called with the same policy.


s-ludwig:
Poor people with common nick names... I didn't expect then to be referenced from the commit message.

PR #1444, opened 2016-03-01
disable botan usage with VibeNoSSL

MartinNowak:


MartinNowak:
It seems some bug in dub is still causing optional dependencies to be downloaded and build.
I though I can select optional dependencies through my dub.selections.json.

MartinNowak:
That's apparently a bug in the development version of dub.

dariusc93:
+1

s-ludwig:
I'll have a look at the DUB master issue. Since 0.9.24 handles optional dependencies in a totally unacceptable way (at least for the default build path that is affected in this case), this change will have to be postponed until 0.9.25 is reasonably widespread.

One thing that should be changed, though, is to add `"default": true` for openssl, so that the default build still supports TLS.

s-ludwig:
DUB master works fine for me. I initially thought I had the same problem, but it turned out that it simply still selected a previous version of vibe.d, thus still having openssl as non-optional.

MartinNowak:
Done, we really should release a new dub version soon, the last release is already 6 month old, and master has heavily diverged since then.

s-ludwig:
Yep, I had too little time recently to make progress on issues that I initially wanted to get into 0.9.25. However, since that is probably going to stay true for another few weeks, we can just as well make another release in-between. Would be good to get https://github.com/D-Programming-Language/dub/pull/762 merged and I'll have to try to remember the issues I had with the current `dub describe --data=...`.

MartinNowak:
Reduced the sope of the PR to just the botan change.
Making openssl optional was already done while switching to SDL.

s-ludwig:
Okay, should be good to go with 1.0.0 out now.

PR #1446, opened 2016-03-06
Fixed documentation for HTTPServerRequest.params

WebFreak001:
http://vibed.org/api/vibe.http.server/HTTPServerRequest.params

it has no documentation because the comment documents the import rather than the `params` variable


s-ludwig:
Oh, good catch, thanks!

Issue #1420, opened 2016-02-12
no stack trace on uncaught exception in task


John-Colvin:
```
% cat source/app.d
import vibe.d;
void main()
{
    runTask((){throw new Exception("");});
    runEventLoop();
}
% dub
[ ... ]
Task terminated with uncaught exception: 
```

No sign of any stack trace.

jdeschenes:
@John-Colvin 

It's very similar to the issue in #1421. In these cases, the message should definitively send the full traceback and not only the error message as it does right now.

s-ludwig:
It doesn't make sense to present a stack track to the user by default. The application can be run with `--vv` or call `setLogLevel(LogLevel.debug_);` to enable debug messages and stack traces will appear.

marcioapm:
I think, ideally, stack traces should be on a separate log level, or perhaps just controlled with a version instead.

With LogLevel.debug_ you get bombarded with vibe's internal logging, which as a library user is not very useful.

Would maybe having a LogLevel.internal or what have you, where all vibe.d debug logging would go be more useful? This way users could use debug_ for application-level debugging and callstacks.

jdeschenes:
Hi,

Logging stack traces in production is invaluable when a user has never recovered from the exception. It is a big indication of a bug and helps greatly in investigating the issue. Setting the default logging to debug_ is simply not viable in that case as there is too much verbosity.

as for @marcioapm's suggestion, I think this is a good idea. Having a lower tier of logging for vibe.d's internal would help when debugging issues.

s-ludwig:
Just noticed that I didn't submit my earlier reply... So I think that many of the `logDebug` calls within vibe.d should actually be `logDebugV`, at least those that definitely are happening in a server context. That would solve the noise issue without yet another log level.

In addition to that, it maybe makes sense to provide a default error logging mechanism within the HTTP server, which can currently be achieved by overriding the `errorPageHandler`.

However, both #1421 and this issue have in common that their exception handling is not ideal. #1421 should ideally have the callback marked as `nothrow`, as there is nothing that the WebSocket code can reasonably do with an exception anyway, other than closing the connection, which it does anyway. 

For `runTask`, there are two options that I was thinking of - either require task functions to also be `nothrow` (way too much breakage at this point), or rethrow uncaught exceptions when another task `join`s the throwing task. However, the latter has a lot of corner cases that need to be handled, it requires additional bookkeeping, and at some point there might still be the situation that nobody handled the exception.

[...]

PR #1450, opened 2016-03-08
Fix #1421: Changed the LogLevel when an uncaught exception to a WebSocket is sent

### LINKED ISSUE #1421: Exception gets swallowed
I modified `examples/websocket` to throw an exception on connection, like this:

```
import vibe.core.log;
import vibe.http.fileserver : serveStaticFiles;
import vibe.http.router : URLRouter;
import vibe.http.server;
import vibe.http.websockets : WebSocket, handleWebSockets;

shared static this()
{
	auto router = new URLRouter;
	router.get("/ws", handleWebSockets(&handleWebSocketConnection));

	settings.port = 8080;
	listenHTTP(settings, router);
}

void handleWebSocketConnection(scope WebSocket socket)
{
    logInfo("hello");
    throw new Exception("");
}
```

When I connect to the socket, `hello` is printed, but there is no sign of the exception and the program continues uninterrupted.

jdeschenes:
When an exception is caught when handling a websocket connection, a logError should be sent instead with the traceback to aid in debugging the actual problem.

### DIFF
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -200,7 +200,7 @@ void handleWebSocket(scope WebSocketHandshakeDelegate on_handshake, scope HTTPSe
 	try {
 		on_handshake(socket);
 	} catch (Exception e) {
-		logDiagnostic("WebSocket handler failed: %s", e.msg);
+		logError("WebSocket handler failed: %s", e);
 	}
 	socket.close();
 }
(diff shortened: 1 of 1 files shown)


jdeschenes:
Before this gets pulled, issue #1420 is also related to this.

Is there any design decision to only spit out the message and not the full traceback when an uncaught error occurs?

Uncaught errors should definitively be sent to the logger with the traceback.

If you want I can do both of the fixes in the same PR.

Let me know.

s-ludwig:
Going the `nothrow` route as discussed: 8668fd0e03d45f15a136eaa63f54d2335a863407
The log level has also been changed to warning until the non-`nothrow` overloads have been removed.

PR #1453, opened 2016-03-10
Fixed StaticRoute working incorrectly with the same headerParam on mu…

IrenejMarc:
…ltiple methods in interface

When the same interface had multiple functions all using the same headerParam, the comparator for AnySatisfy was reused from previously-compared functions, resulting in completely incorrect results

Some details in the code below.

Example:
```d
import vibe.d;

interface API
{
	Collection!FooCollection foos();
}

class APIImpl : API
{
	private FooCollection _foos;

	this()
	{
		_foos = new FooCollectionImpl;
	}

	Collection!FooCollection foos()
	{
		return Collection!FooCollection(_foos);
	}
}

enum fooHeader = headerParam("foo", "FooHeader");
interface FooCollection
{
	struct CollectionIndices
	{
		long _id;
	}


// foo is the second param here
	@fooHeader
		string getSomething(long asd, string foo);

// BOTH the second param and the third one will now be headerParams("<fieldname>", "FooHeader")
	@fooHeader
		string get(long _id, int bar, string foo);
}

class FooCollectionImpl : FooCollection
{

	string getSomething(long asd, string foo)
	{
		return foo;
	}

	string get(long _id, int bar, string foo)
	{
		return foo;
	}
}

shared static this()
{
	settings.port = 8080;

	auto router = new URLRouter;
	router.registerRestInterface(new APIImpl);

	listenHTTP(settings, router);
}
```

### DIFF
--- a/source/vibe/web/internal/rest/common.d
+++ b/source/vibe/web/internal/rest/common.d
@@ -260,7 +260,7 @@ import vibe.web.rest;
 
 				// Comparison template for anySatisfy
 				//template Cmp(WebParamAttribute attr) { enum Cmp = (attr.identifier == ParamNames[i]); }
-				alias CompareParamName = GenCmp!("Loop", i, parameterNames[i]);
+				alias CompareParamName = GenCmp!("Loop"~func.mangleof, i, parameterNames[i]);
 				mixin(CompareParamName.Decl);
 
 				StaticParameter pi;


s-ludwig:
I'm just wondering if using the manged name as an identifier is a good idea. For functions with D linkage (admittedly almost always the case), it should work, but for C++ linkage it might not for example. Safer might be to use `fi.stringof`.

IrenejMarc:
Seeing as this is not a GenCmp change, but just a change in the StaticRoute, that applies to only REST interfaces, is it even possible to have C++ linkage for those methods?

s-ludwig:
In theory it is, but admittedly highly unlikely. Ideally, GenCmp should be extended to support nested indices directly, I'll do that later when I get some time. But for now, the mangleof fix should do.

PR #1454, opened 2016-03-11
Issue1452 crash for tcp close at shutdown

### LINKED ISSUE #1452: Segfault on connection close
0.7.28
Ubuntu 64-bit
libevent2

Whenever a connection is closed, logs get spammed with the following *forever*:
```
[781D1FE4:00000000 2016.03.09 11:23:26.491 dia] Actively closing TCP connection
core.exception.InvalidMemoryOperationError@src/core/exception.d(679): Invalid memory operation
----------------
core.exception.InvalidMemoryOperationError@src/core/exception.d(679): Invalid memory operation
----------------
core.exception.InvalidMemoryOperationError@src/core/exception.d(679): Invalid memory operation
----------------
...
```

Will stay on 0.7.25 for now

s-ludwig:
Fixes the `InvalidMemoryOperationError` by avoiding the assertion and instead writing a warning message to the console. Fixes #1452.

### DIFF
--- a/source/vibe/core/drivers/libevent2_tcp.d
+++ b/source/vibe/core/drivers/libevent2_tcp.d
@@ -160,7 +160,7 @@ package final class Libevent2TCPConnection : TCPConnection {
 
 		if (!getThreadLibeventEventLoop()) {
 			import std.stdio;
-			stderr.writefln("Warning: Dangling TCP connection to %s left alive at shutdown. "
+			stderr.writefln("Warning: Attempt to close dangling TCP connection to %s at shutdown. "
 				~ "Please avoid closing connections in GC finalizers.", m_remoteAddress);
 			return;
 		}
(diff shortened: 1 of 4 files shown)


marcioapm:
What about the case of someone for example destroying the `ConnectionPool` in the middle of the program run? Won't that leave all the sockets open in libevent?

I was thinking if it would work to have `Connection` and `ConnectionPool` have a finalizer that would destroy their `Connection`. Vibe.d would then keep track of all `ConnectionPool` alive and destroy them right before libevent. This way there would never be dangling connections and from the user perspective, it just works.

For this to work, `Connection` must have either a finalizer or a close method.

This "requirement" would also allow `ConnectionPool` to be able to shrink, which at the moment, it can't.

s-ludwig:
Unfortunately finalizers are not allowed to even touch GC allocated memory at all, according to the  language spec. So at least those cannot be used, it would have to be a separate `dispose()` style method. But my plan for the upcoming major release is to convert all resources into reference counted structs anyway, so that this issue is generally gone (well, as long as those structs are not stored within classes at least).

Extending `ConnectionPool` to gain more insight into a connection generally sounds like a good idea. It just needs to be worked out how to best make the transition in terms of backwards compatibility (probably using simple duck-typing).

PR #1456, opened 2016-03-17
add sqlite mime-type

UplinkCoder:


s-ludwig:
I was a little worried about the uniqueness of the file extension "db", but I guess it's today by far most popular for sqlite databases and it will have a small impact anyway. However, what about adding "s3db" to that list, it is mentioned quite frequently on the internet.

UplinkCoder:
good idea

UplinkCoder:
@s-ludwig done

s-ludwig:
Thanks, merging.

PR #1468, opened 2016-04-04
Support non-string keys in `parseJson`

Dicebot:
This is necessary to be able to parse JSON created by browsers via
JavaScript `JSON.stringify`.

Right now lack of this prevents using `vibe.web.rest` with browser
JavaScript clients relying on `JSON.stringify`. Though it is worth
noting that JSON grammar at http://www.json.org doesn't list it
is legit.


s-ludwig:
Which browser does that happen on? Even if the spec for `stringify` is rather complex and I didn't read it in all detail, I'm pretty sure that the key is defined to be always quoted.

PR #1469, opened 2016-04-06
special exception for connection errors in libevent

yannick:
ideas here:
a) make the errors catchable so it can be ignored or more silently logged
b) add some information about the peer if possible.

*PR is probably not usable as is, but i want to discuss the Idea.*

reason: i see lots of such errors in production (mainly from mobile ios devices):
```
 vibe.core.drivers.utils.ConnectionException@../vibe.d/source/vibe/core/drivers/libevent2_tcp.d(432): Connection error while writing to TCPConnection. Remote: 127.0.0.1:55774
[EF29AF15:EF2B1315 2016.04.06 18:15:51.581 ERR] ----------------
[EF29AF15:EF2B1315 2016.04.06 18:15:51.581 ERR] ../vibe.d/source/vibe/core/drivers/libevent2_tcp.d:432 void vibe.core.drivers.libevent2_tcp.Libevent2TCPConnection.checkConnected(bool) [0x8e45bc]
[EF29AF15:EF2B1315 2016.04.06 18:15:51.581 ERR] ../v
```

### DIFF
--- a/source/vibe/core/drivers/libevent2_tcp.d
+++ b/source/vibe/core/drivers/libevent2_tcp.d
@@ -348,7 +348,7 @@ package final class Libevent2TCPConnection : TCPConnection {
 		logTrace("evbuffer_add (fd %d): %d B", m_ctx.socketfd, bytes.length);
 		auto outbuf = bufferevent_get_output(m_ctx.event);
 		if( bufferevent_write(m_ctx.event, cast(char*)bytes.ptr, bytes.length) != 0 )
-			throw new Exception("Failed to write data to buffer");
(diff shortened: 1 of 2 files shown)


WebFreak001:
@s-ludwig no comment?

s-ludwig:
Having proper separate exception types is definitely a pressing issue for the library in general. The problem is that it is going to be very difficult to switch from a custom exception hierarchy to exceptions defined in Phobos once they become available. However, if the depth of the hiearchy within vibe.d is kept to 1 and all those exception classes are marked final, it should be possible without breaking normal user code.

In this case, the question is if this should derive from `std.socket.SocketException`. Also, it should either be defined in a common place for vibe-d:core and vibe-core, or just in vibe-core so that no forward-compatibility issues get introduced with this.

Issue #1462, opened 2016-03-29
File descriptors leak


yazd:
The following program leaks file descriptors just by importing `vibe.core.core`. Removing the import results in a constant number of file descriptors: 4.

```
import std.stdio;
import std.parallelism;
import std.range;
import std.algorithm;
import std.file;

import vibe.core.core;

void main()
{
    writeln(__VERSION__);

    foreach (i; 0 .. 10)
    {
        TaskPool t = new TaskPool();
        t.finish(true);
        destroy(t);

        writeln("Number of file descriptors: ", dirEntries("/proc/self/fd/", SpanMode.breadth, false).walkLength);
    }
}
```

Example output (the numbers are always the same):

```
2070
Number of file descriptors: 24
Number of file descriptors: 39
Number of file descriptors: 54
Number of file descriptors: 69
Number of file descriptors: 84
Number of file descriptors: 99
Number of file descriptors: 114
Number of file descriptors: 129
Number of file descriptors: 144
Number of file descriptors: 159
```

Tested on Ubuntu 64bit 4.2.0-34-generic and dmd 2.070.0.
Fails on vibe-d versions 0.7.28 and 0.7.29-alpha.2.
Works correctly on vibe-d version 0.7.26.

yazd:
Maybe related to https://github.com/rejectedsoftware/vibe.d/issues/1244

yazd:
Looks like valgrind can trace file descriptors as well.
Output when creating one `TaskPool` with 1 worker thread.

```
==2417== FILE DESCRIPTORS: 8 open at exit.
==2417== Open AF_INET socket 12: <unbound> <-> unbound
==2417==    at 0x657F067: socket (syscall-template.S:81)
==2417==    by 0x50676D2: ??? (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
==2417==    by 0x5069A8C: evdns_base_nameserver_ip_add (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
==2417==    by 0x5069D74: ??? (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
==2417==    by 0x506CE34: evdns_base_resolv_conf_parse (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
==2417==    by 0x506D173: evdns_base_new (in /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5.1.9)
[...]

PR #1472, opened 2016-04-12
Fix issue 1462 - fix file leak due to extra driver setup

yazd:
Fixes https://github.com/rejectedsoftware/vibe.d/issues/1462 but requires Phobos fix https://github.com/D-Programming-Language/phobos/pull/4191.

If this is pulled before Phobos has the fix, an assertion will fail:

```
core.exception.AssertError@vibe.d/source/vibe/core/concurrency.d(1227): Assertion failure
----------------
??:? _d_assert [0x6340cf]
??:? void vibe.core.concurrency.__assert(int) [0x60113b]
vibe.d/source/vibe/core/concurrency.d:1227 nothrow vibe.core.sync.TaskCondition vibe.core.concurrency.VibedScheduler.newCondition(core.sync.mutex.Mutex) [0x601388]
??:? @trusted std.concurrency.MessageBox std.concurrency.MessageBox.__ctor() [0x647793]
??:? @trusted std.concurrency.Tid std.concurrency.thisTid().__dgliteral1() [0x646f66]
??:? @property @safe std.concurrency.Tid std.concurrency.thisTid() [0x646f16]
??:? void std.concurrency.unregisterMe() [0x683620]
??:? void std.concurrency.ThreadInfo.cleanup() [0x647028]
??:? void std.concurrency._staticDtor265() [0x646aa8]
??:? void std.concurrency.__moddtor() [0x646ab8]
```

So maybe this needs to be versioned before pulling.

### DIFF
--- a/core/vibe/core/concurrency.d
+++ b/core/vibe/core/concurrency.d
@@ -1279,7 +1279,11 @@ package class VibedScheduler : Scheduler {
 			if (LibasyncDriver.isControlThread)
 				return null;
 		}
-		setupDriver();
+		// due to useless creation of a new Condition during std.concurrency module
+		// destructor, setupDriver call is necessary in versions prior to 2.072
+		static if (__VERSION__ < 2072) {
+			setupDriver();
+		}
 		return new TaskCondition(m);
 	}
 }


s-ludwig:
Interesting. Did you test if this change is necessary at all, with the Phobos fix present? Without actually running the code, my guess would have been that the creation of the extra `MessageBox` lead to `newCondition` being called. But on the other hand, if nothing crashes at startup or during Thread creation, removing the `setupDriver` call should be fine anyway.

yazd:
Yes, this is unneeded with the Phobos fix present. I'll wait until the next dmd beta is prepared to be sure that the fix is included and then I can version this if you are ok with it.

s-ludwig:
Does this still make sense to change? In `vibe-core`, there is no `setupDriver` call anymore and the Phobos fix is in 2.072.0. Since 2.071.0 is still supported, a `static if (__VERSION__ >= 2072)` would still be required.

yazd:
Updated.

vitalka200:
Hi @s-ludwig / @yazd,
Any chance it can be merged?

s-ludwig:
@vitalka200 I think this really should just be closed, as the earliest supported version of DMD nowadays is 2.076.1 and the Phobos fix was released with 2.072.0.

PR #1477, opened 2016-04-20
Fix support for client response

9il:


yannick:
imho its should even accept ALL Content-Encoding's and _optionally_ decode the ones known. 
maybe also _optionally_ throw if its a unknown encoding.

i don't see any reason why you might NOT want to store a already gzipd file on your disk directly before further processing it.

yannick:
i think decoding and checking Content-Type should be 2 optional settings.
- you might want to choose to save a decoded stream directly to disk
- you might want to decode known encodings but still save unknowns as is!

s-ludwig:
I'll merge this first and then have a closer look at #1478.

PR #1478, opened 2016-04-20
add DecodeContent options for bodyReader

yannick:
give the Consumer the option to decide if the http payload should be transparently decoded or not
and what to do with unknown encodings.

supersedes https://github.com/rejectedsoftware/vibe.d/pull/1477

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -766,6 +766,11 @@ final class HTTPClientResponse : HTTPResponse {
 		int m_maxRequests;
 	}
 
+	enum DecodeContent {
+		Strict, // Known encodings are decoded, unknown throw
+		Known,  // Known encodings are decoded, unknown are forwarded untouched
+		None 		// No Content is decoded at all
+	}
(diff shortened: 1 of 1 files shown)


9il:
rebase please

yannick:
Needs refactoring.

```
s-ludwig ╡  I think it needs two changes: decodeContent shouldn't be a
         ╡ parameter for bodyReader, as it will oly have an effect on the first call, which may be unexpected, and there must be some kind of indication to the user if the result
         ╡ is decoded or not.                                                       
```

s-ludwig:
So my proposal for the first issue would be to simply add a method `setContentDecodeMode` (I'd also call the enum accordingly). For determining if decoding is effective, I'm not yet sure.

The obvious solution would be to add a property `isContentDecoded`. This would have to be determined up front, before the first call to `bodyReader` and after each call to `setContentDecodeMode` to reduce the risk of usage mistakes. Using a special return type for `bodyReader` instead of a plain `InputStream` would be another possibility, and has the advantage that there would still be only one place to determine how to decode. However, any additional runtime overhead must be avoided.

PR #1488, opened 2016-05-07
Fixed writeJsonBody not to always overwrite statusCode to 200

IrenejMarc:
It was impossible to set the status code of any RestInterface functions, even through `@after` because it always got overwritten by writeJsonBody which defaulted to 200


s-ludwig:
Can you move the doc comment to the first overload and put a `/// ditto` to the second one? Otherwise LGTM.

IrenejMarc:
All done. My bad.

s-ludwig:
Okay, thanks!

Issue #1490, opened 2016-05-12
registerWebInterface() redirect ignores X-Forwarded-Port header


Mihail-K:
Minimal case to reproduce,
```d
@path("/users")
class Users {
    void index() {
        render!"index.dt";
    }
}

shared static this() {
    URLRouter router = new URLRouter;

    // . . .

    router.registerWebInterface(new Users);
}
```

Requests made to `/users` are redirected to `/users/`, however, the redirect includes the port number the vibe app accepted the request on. This is problematic for environments where the internal port number (which vibed is bound to) does not match the external port number (which requests arrive on). For example if the app listens on port 5000, but incoming requests are actually accepted on port 80.

I've set up a Heroku app that demonstrates this behavior. Note that it is running on a free dyno, and may take a while to spin up.

Works correctly: http://testrig.herokuapp.com/users/
Problem redirect: http://testrig.herokuapp.com/users
(May redirect to something like: http://testrig.herokuapp.com:48669/users/)

In the case of Heroku, this is particularly problematic as there is no control over what the internal port number is. It is assigned (via environment variable) whenever the app is deployed, and is usually different every time.

s-ludwig:
Currently the server expects the "X-Forwarded-Host" header to be set by the proxy server - if that is given, it will ignore the local port. This works for simple setups, but should definitely be made customizable in the future. Does Heroku set any similar headers, or can it be made to?

Mihail-K:
It appears Heroku does not set the `X-Forwarded-Host` header, but instead all of these,
https://devcenter.heroku.com/articles/http-routing#heroku-headers

Other major HTTP servers appear to support these, but this is probably the cause of the issue.

Mihail-K:
Resolved by #1491

PR #1491, opened 2016-05-12
#1490 Add support for X-Forwarded-Port header.

Mihail-K:
Honours the `X-Forwarded-Port` header, when present.

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -707,28 +707,53 @@ final class HTTPServerRequest : HTTPRequest {
 	@property URL fullURL()
 	const {
 		URL url;
-		auto fh = this.headers.get("X-Forwarded-Host", "");
-		if (!fh.empty) {
-			url.schema = this.headers.get("X-Forwarded-Proto", "http");
-			url.host = fh;
-		} else {
-			if (!this.host.empty) url.host = this.host;
-			else if (!m_settings.hostName.empty) url.host = m_settings.hostName;
-			else url.host = m_settings.bindAddresses[0];
 
-			if (this.tls) {
-				url.schema = "https";
-				if (m_port != 443) url.port = m_port;
-			} else {
-				url.schema = "http";
-				if (m_port != 80) url.port = m_port;
+		// Set URL host segment.
+		if (auto xfh = this.headers.get("X-Forwarded-Host")) {
+			url.host = xfh;
+		} else if (!this.host.empty) {
+			url.host = this.host;
+		} else if (!m_settings.hostName.empty) {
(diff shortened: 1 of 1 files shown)


source/vibe/http/server.d
```diff
@@ -707,28 +707,53 @@ final class HTTPServerRequest : HTTPRequest {
...
+			url.schema = "https";
+		} else {
+			url.schema = "http";
+		}
 
-			if (this.tls) {
-				url.schema = "https";
-				if (m_port != 443) url.port = m_port;
+		// Set URL port segment.
+		Nullable!ushort port;
```
s-ludwig: Could simply use `ushort port = 0;` here, because 0 isn't a valid port anyway.

source/vibe/http/server.d
```diff
@@ -707,28 +707,53 @@ final class HTTPServerRequest : HTTPRequest {
...
+		if (auto xfp = this.headers.get("X-Forwarded-Port")) {
+			try {
+				port = xfp.to!ushort;
+			} catch (ConvException) {
+				// Value was not ushort.
```
s-ludwig: Another possibility would be to respond with a 400 error here. May be the better option in case of e.g. proxy configuration errors. But I'd say that there should be a diagnostic log message here at least.

Mihail-K: Alright. I've added a logger-warning in case of a bad port number. I'm hesitant to serve a 400 error from here though, considering this is in `HTTPServerRequest`, which shouldn't really be sending responses, should it?

s-ludwig: You are right, I forgot that this happens after the general request parsing, within the user's request callback.

s-ludwig:
Okay, thanks! Merging in.

source/vibe/http/server.d
```diff
@@ -707,28 +707,47 @@ final class HTTPServerRequest : HTTPRequest {
...
 
-			if (this.tls) {
-				url.schema = "https";
-				if (m_port != 443) url.port = m_port;
-			} else {
-				url.schema = "http";
-				if (m_port != 80) url.port = m_port;
+		// Set URL schema segment.
+		if (auto xfp = this.headers.get("X-Forwarded-Proto")) {
+			url.schema = xfp;
```
MartinNowak: I'd argue this should set the port to the default port for that schema for cases where X-Forwarded-Port isn't set.

s-ludwig: Setting no port (defaut of `0`) is considered to be implicitly equal to setting the default port. Are you thinking of something in particular that is broken?

PR #1492, opened 2016-05-13
 Http listener

gedaiu:
I wanted to create a new http listener, but it was complicated to do without duplicating a lot of code, so I decided to try to change some things to make the code easier to read and reuse. 

This is still work in progress, please don't merge it yet, since I will add more changes to it... 

It would be nice if I could get some feedback if this is a good ideea or not... or maybe there is someone else working on the same thing?

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -79,17 +79,20 @@ HTTPListener listenHTTP(HTTPServerSettings settings, HTTPServerRequestDelegate r
 {
 	enforce(settings.bindAddresses.length, "Must provide at least one bind address for a HTTP server.");
 
-	HTTPServerContext ctx = g_contextCollection.create(settings, request_handler);
-
 	// if a VibeDist host was specified on the command line, register there instead of listening
 	// directly.
 	if (s_distHost.length && !settings.disableDistHost) {
 		listenHTTPDist(settings, request_handler, s_distHost, s_distPort);
-	} else {
-		listenHTTPPlain(settings);
+	}
+	else
+	{
+		auto listener = new HTTPListenerPlain(settings, request_handler);
+		listener.start();
+
+		return listener;
 	}
 
-	return HTTPListener(ctx.id);
+	return new HTTPListener(settings, request_handler);
 }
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'm currently a bit reluctant with refactorings in the HTTP code, because I'll have to merge the HTTP/2 branch soon and there are already lots of rather difficult to resolve conflicts. Are there any functional changes that we can apply for now?

gedaiu:
There are no functional changes... I wanted to write another http request parser and I figured out that in order to add this parser, I need to make some changes here.

anyway, you can merge that http/2 branch and I will update my code later

s-ludwig:
I just realized that I did a similar refactoring as part of #1929, although I didn't go quite as far. The main difference is that the changes I did allow to create a custom listener from the outside instead of providing lower level functions for doing so (e.g. `HTTPListenerPlain`). Does that fit the purpose that you had in mind, or do we need to integrate something additional from your changes?

BTW, it has become pretty certain by now that the HTTP/2 branch can not simply be merged. Instead, I'm slowly working on a complete redesign of the HTTP package to better match the new stream model of vibe-core and to generally remove some of the performance issues inherent to the current design. For this reason, refactoring the existing code is okay again as long as it doesn't contradict the new, unfinished design.

gedaiu:
this pull request it's kind of old... I think that the best option is to close it. Right now I am a little bit busy with some other projects, but I can try to see what I can get from the HTTP/2 branch and make a new pull request. How does that sound?

s-ludwig:
I'd leave the current HTTP/2 branch aside for now. Do you have anything from the PR that you'd like to see in the future? At the very least I would keep that in mind when thinking about the redesign. Once that is roughly done, this functionality can then also be implemented on master immediately, just so that it doesn't accidentally introduce a forward compatibility issue with the redesigned API.

gedaiu:
I think that at that time I wanted to organise a little bit the code... In the end I think I wanted to have some sort of middleware, which can be used for adding features like CORS that are sometimes needed for all the requests. 

But this was a long time ago...

Issue #1496, opened 2016-05-20
Easy DOS with crafted WebSocket packet


mathias-lang-sociomantic:
https://github.com/rejectedsoftware/vibe.d/blob/85178cfd3f1d8299c6e73956ef9eb4a31225b488/source/vibe/http/websockets.d#L755-L758

Step 1: Read the length from the network
Step 2: Assert that the given length is less than the maximum amount of virtual memory on your system
Step 3: Allocate a buffer of that length

Issue #1475, opened 2016-04-14
Exception in lev_[un]lock_mutex: Unknown lock handle


mathias-lang-sociomantic:
I'm hitting the following `assert` in a test of mine:
https://github.com/rejectedsoftware/vibe.d/blob/0.7.28/source/vibe/core/drivers/libevent2.d#L1360

The repo, while easy, is not straightforward:
```bash
dub init random --type=vibe.d && cd random
# You need this repository and branch, as there is a change to avoid duplicate main
git clone git@github.com:mathias-lang-sociomantic/dmd-dlang.git submodules/dmd
cd submodules/dmd && git checkout origin/vibed-bug-repro && cd -
```

Add the following to dub.sdl :
```
versions "VibeDefaultMain" "LibDDMD"

sourcePaths "source/" "submodules/dmd/src/"
importPaths "source/" "submodules/dmd/src/"
stringImportPaths "submodules/dmd/src/"
excludedSourceFiles "submodules/dmd/src/objc.d"
excludedSourceFiles "submodules/dmd/src/checkwhitespace.d" "submodules/dmd/src/idgen.d"
excludedSourceFiles "submodules/dmd/src/libmscoff.d" "submodules/dmd/src/scanmscoff.d"
excludedSourceFiles "submodules/dmd/src/scanmach.d" "submodules/dmd/src/libmach.d"
excludedSourceFiles "submodules/dmd/src/scanomf.d" "submodules/dmd/src/libomf.d"
preBuildCommands "make -C submodules/dmd/ -f posix.mak"
lflags "submodules/dmd/src/glue.a" "submodules/dmd/src/backend.a"
libs "stdc++"
```

I also use `"VibeDebugCatchAll"` to not get flooded by messages.
Run `dub` twice (the first time will generate `id.d` but it won't properly be picked up, the second time it'll work).

Local output:
```
Running ./random 
Listening for requests on http://[::1]:8080/
Listening for requests on http://127.0.0.1:8080/
Please open http://127.0.0.1:8080/ in your browser.
Exception in lev_lock_mutex: Unknown lock handle
Exception in lev_unlock_mutex: Unknown lock handle
Exception in lev_lock_mutex: Unknown lock handle
Exception in lev_unlock_mutex: Unknown lock handle
[Ctrl+C]
^CException in lev_lock_mutex: Unknown lock handle
Exception in lev_unlock_mutex: Unknown lock handle
Received signal 2. Shutting down.
[...]

PR #1497, opened 2016-05-20
Put some band-aid on WebSocket

mathias-lang-sociomantic:
@s-ludwig : After looking at the code in this module, I believe it needs serious work before being ready for production.

This P.R. will mitigate some of the issues I encountered, but things like https://github.com/rejectedsoftware/vibe.d/issues/1496 make it completely unusable for any serious application.

I tested this with a crippled version of my Slack bot, however I was blocked by https://github.com/rejectedsoftware/vibe.d/issues/1475 (as only an old frontend, such as GDC's, produce correct code, but Vibe.d's master is not compatible with 2.067 anymore). I'll try to invest some time in tracking the issues in the near future.

### DIFF
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -386,9 +386,9 @@ final class WebSocket {
 		On the JavaScript side, the text will be available as message.data (type string).
 		Throws: WebSocketException if the connection is closed.
 	*/
-	void send(string data)
+	void send(scope const(char)[] data)
 	{
-		send((scope message){ message.write(cast(ubyte[])data); });
+		send((scope message){ message.write(cast(const ubyte[])data); });
 	}
 
 	/**
@@ -423,7 +423,7 @@ final class WebSocket {
 			code = Numeric code indicating a termination reason.
 			reason = Message describing why the connection was terminated.
 	*/
-	void close(short code = 0, string reason = "")
+	void close(short code = 0, scope const(char)[] reason = "")
 	{
 		//control frame payloads are limited to 125 bytes
 		assert(reason.length <= 123);
@@ -435,7 +435,7 @@ final class WebSocket {
 				frame.isServer = m_isServer;
 				frame.opcode = FrameOpcode.close;
 				if(code != 0)
-					frame.payload = std.bitmanip.nativeToBigEndian(code) ~ cast(ubyte[])reason;
+					frame.payload = std.bitmanip.nativeToBigEndian(code) ~ cast(const ubyte[])reason;
 				frame.fin = true;
 				frame.writeFrame(m_conn);
 			});
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -692,7 +692,14 @@ final class IncomingWebSocketMessage : InputStream {
 
 private immutable s_webSocketGuid = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
 
-enum FrameOpcode {
+/**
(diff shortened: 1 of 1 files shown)


source/vibe/http/websockets.d
```diff
@@ -386,9 +386,9 @@ final class WebSocket {
 		On the JavaScript side, the text will be available as message.data (type string).
 		Throws: WebSocketException if the connection is closed.
 	*/
-	void send(string data)
+	void send(const(char)[] data)
```
s-ludwig: Could even be weakened to `scope const(char)[]`.

source/vibe/http/websockets.d
```diff
@@ -423,7 +423,7 @@ final class WebSocket {
 			code = Numeric code indicating a termination reason.
 			reason = Message describing why the connection was terminated.
 	*/
-	void close(short code = 0, string reason = "")
+	void close(short code = 0, const(char)[] reason = "")
```
s-ludwig: Same here.

source/vibe/http/websockets.d
mathias-lang-sociomantic: Since `scope` is not implemented, is it really worth it ?

s-ludwig:
Looks good. Yeah, for #1496 we need some user defined frame/message size limits. A `WebSocketSettings` class probably makes sense anyway, so that things like automatic keep-alives can be enabled.

Generally, this module (from my side) unfortunately has been sitting there without being used for anything apart from a few tests and a tutorial and I guess it shows...

source/vibe/http/websockets.d
s-ludwig: It's at least a documentation cue, and I *really*  hope that it gets implemented one day, or at least that more illegal cases are detected (but it actually needs to be implemented correctly so that `@safe` isn't broken).

mathias-lang-sociomantic:
@s-ludwig : Updated with:

```diff
diff --git a/source/vibe/http/websockets.d b/source/vibe/http/websockets.d
index baeb366..fbd9965 100644
--- a/source/vibe/http/websockets.d
+++ b/source/vibe/http/websockets.d
@@ -389,7 +389,7 @@ final class WebSocket {
                On the JavaScript side, the text will be available as message.data (type string).
                Throws: WebSocketException if the connection is closed.
        */
-   void send(const(char)[] data)
+ void send(scope const(char)[] data)
        {
                send((scope message){ message.write(cast(const ubyte[])data); });
        }
@@ -426,7 +426,7 @@ final class WebSocket {
                        code = Numeric code indicating a termination reason.
                        reason = Message describing why the connection was terminated.
        */
-   void close(short code = 0, const(char)[] reason = "")
+ void close(short code = 0, scope const(char)[] reason = "")
        {
                //control frame payloads are limited to 125 bytes
                assert(reason.length <= 123);
```

mathias-lang-sociomantic:
I let it run over the week-end with the GC disabled. It did not work obviously, but it didn't crash either like it used to.

I just got a stream of: `Task terminated with uncaught exception: Operating on closed TCPConnection.`

mathias-lang-sociomantic:
I got a spurious [failure](https://travis-ci.org/rejectedsoftware/vibe.d/jobs/132283840), now it's green.

mathias-lang-sociomantic:
Ping @s-ludwig

s-ludwig:
Sorry, forgot this again while waiting for Travis to finish..

mathias-lang-sociomantic:
Thanks!

PR #1499, opened 2016-05-24
libevent2: Don't expose debug global data as extern(C)

mathias-lang-sociomantic:
There is no reason for them to be extern(C).
However, they can be easily corrupted / misinterpreted as
their name would be quite trivial in the resulting binary.

Related to rejectedsoftware/vibe.d#1475


mathias-lang-sociomantic:
Got another spurious failure with LDC + libevent: https://travis-ci.org/rejectedsoftware/vibe.d/builds/132643340

mathias-lang-sociomantic:
This one as well @s-ludwig ? ;)

s-ludwig:
Guess so ^^

PR #1500, opened 2016-05-24
fix serialization conditions order

tchaloupka:
I can't get to work custom policy to deserialize enum values as I described in [forum](https://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/30890/#post-30898).

So I tried to find out why and found there is a bug in current implementation.

Problem is with the static if conditions order where even if the enum policy condition passed, it is not used as it is too far bottom and other condition gets prioritized before intended function.

With this PR it works ok.

Although as I looked on the other conditions, it seems to me, that they are wrong too, ie isCustomSerializable!T, isISOExtStringSerializable!T should probably also go upper. But I did not dig deeper.

Sample code, which works for me is this:
```D
import std.stdio;
import vibe.d;

enum Color {red, green, blue}

template UpperCasePolicy(T) if (is(T == Color)) {
     import std.conv, std.string : toLower, toUpper;

     Color fromRepresentation(string value) {return value.toLower.to!Color;}

     string toRepresentation(Color value) {return to!string(value).toUpper;}
}

struct Test {
     @byName!UpperCasePolicy Color color;
}

void main()
{
	Test test = {Color.red};
	assert(test.serializeWithPolicy!(JsonSerializer, UpperCasePolicy).toString == `{"color":"RED"}`);
	assert(Json(["color": Json("GREEN")]).deserializeWithPolicy!(JsonSerializer, UpperCasePolicy, Test) == Test(Color.green));
}
```

I can't get it to compile with JsonStringSerializer - as mentioned on forum (`Error: template vibe.data.serialization.serializeWithPolicy cannot deduce function from argument types !(JsonStringSerializer, UpperCasePolicy)(Test)`)

### DIFF
--- a/source/vibe/data/serialization.d
+++ b/source/vibe/data/serialization.d
@@ -297,7 +297,10 @@ private void serializeImpl(Serializer, alias Policy, T, ATTRIBUTES...)(ref Seria
 
 	alias TU = Unqual!T;
 
-	static if (is(TU == enum)) {
+	static if (isPolicySerializable!(Policy, TU)) {
+		alias CustomType = typeof(Policy!TU.toRepresentation(TU.init));
+		serializeImpl!(Serializer, Policy, CustomType, ATTRIBUTES)(serializer, Policy!TU.toRepresentation(value));
+	} else static if (is(TU == enum)) {
 		static if (hasPolicyAttributeL!(ByNameAttribute, Policy, ATTRIBUTES)) {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Makes sense indeed! The other ones should be fine, though, since they only apply to structs/classes and only built-in types/`Nullable` are tested in the conditions above. I'll have a look at the remaining issue.

PR #1508, opened 2016-06-02
Correct the client Websocket handshake

Geod24:
The RFC specify in section 4.1 (Opening Handshake, Client Requirements) that:
"The value of this header field MUST be a nonce consisting of a randomly
selected 16-byte value that has been base64-encoded"

However `randomUUID.toString` would return a 36 chars string.
As a result, the base64 encoding was 48 chars long instead of 24.


Geod24:
Added a commit that do case insensitive comparison of the protocol string in the http client (caused `Upgrade: WebSockets` to be wrongly rejected).

s-ludwig:
LGTM!

PR #1509, opened 2016-06-04
Don't force immutability for string algorithms

Geod24:
When taking a string argument for validation, 'const(char)[]' should be used instead of 'string'.
That's because the former only promise the caller it won't modify the reference,
while the later ask for the strong promise (from the caller) that this argument shall never be modified.

In addition, some of those algorithm will either return a portion of the input string or a newly
allocated one.  This may result in a lot of otherwise avoidable allocations.  The runtime decision also
makes it impossible to use in `@nogc` code.
Ultimately, this code could be modified to take a buffer as parameter.
For the time being those functions were turned into templates.

Note: This is a rather shallow pass.

### DIFF
--- a/source/vibe/textfilter/urlencode.d
+++ b/source/vibe/textfilter/urlencode.d
@@ -16,10 +16,12 @@ import std.exception;
 import std.format;
 
 
-/** Returns the URL encoded version of a given string.
-*/
-string urlEncode(string str, string allowed_chars = null)
-@safe {
+/**
+ * Returns:
+ *   the URL encoded version of a given string, in a newly-allocated string.
+ */
+T[] urlEncode(T)(T[] str, const(char)[] allowed_chars = null) if (is(T[] : const(char)[]))
+{
 	foreach (char c; str) {
 		switch(c) {
 			case '-':
@@ -31,7 +33,7 @@ string urlEncode(string str, string allowed_chars = null)
 			case '~':
 				break;
 			default:
-				auto dst = appender!string();
+				auto dst = appender!(T[]);
 				dst.reserve(str.length);
 				filterURLEncode(dst, str, allowed_chars);
 				return dst.data;
@@ -40,12 +42,12 @@ string urlEncode(string str, string allowed_chars = null)
 	return str;
 }
 
-unittest {
+@safe unittest {
 	string s = "hello-world";
 	assert(s.urlEncode().ptr == s.ptr);
 }
 
-private bool isCorrectHexNum(string str)
+private auto isCorrectHexNum(const(char)[] str)
 @safe {
 	foreach (char c; str) {
 		switch(c) {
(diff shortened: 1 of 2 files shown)


s-ludwig:
Right, the range based overloads are intended for allocation-free operation. Adding a more classical buffer based optimization shouldn't generally hurt, but it won't solve the `@nogc` issue, unless they are implemented as a separate overload and insufficient buffer space causes a failure instead of reallocating.

source/vibe/textfilter/urlencode.d
```diff
@@ -40,12 +42,12 @@ string urlEncode(string str, string allowed_chars = null)
...
 }
 
-unittest {
+@safe unittest {
 	string s = "hello-world";
 	assert(s.urlEncode().ptr == s.ptr);
 }
 
-private bool isCorrectHexNum(string str)
+private auto isCorrectHexNum(const(char)[] str)
```
s-ludwig: Why `auto`?

source/vibe/utils/string.d
```diff
@@ -108,29 +108,29 @@ bool anyOf(string str, string chars)
...
+inout(char)[] stripRightA(inout(char)[] s)
 @safe pure nothrow {
 	while (s.length > 0 && (s[$-1] == ' ' || s[$-1] == '\t'))
 		s = s[0 .. $-1];
 	return s;
 }
 
 /// ASCII whitespace trimming (space and tab)
-string stripA(string s)
+inout(char)[] stripA(inout(char)[] s)
```
s-ludwig: Didn't ever use `inout` for static/global functions, this is neat to have.

source/vibe/textfilter/urlencode.d
Geod24: Attribute inference for recent compilers

Geod24: It's present since 2.068: http://dlang.org/changelog/2.068.0.html#attribinference3

s-ludwig: Ugh okay, that's rather ugly having to obfuscate the return value just to get attribute inference. It's just a private function, so it's not that important, though.

source/vibe/textfilter/urlencode.d
```diff
@@ -62,7 +64,7 @@ private bool isCorrectHexNum(string str)
 
 /** Checks whether a given string has valid URL encoding.
 */
-bool isURLEncoded(string str, string reserved_chars = null)
+auto isURLEncoded(const(char)[] str, const(char)[] reserved_chars = null)
```
s-ludwig: But in this case I'd rather put the attributes manually, so that both, the return type and the attributes appear in the documentation.

Geod24: Makes sense. Updated to `bool` again :)

s-ludwig:
Okay, thanks! Merging in.

Issue #1507, opened 2016-06-02
Missing port information on HTTP request


Geod24:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23

> A "host" without any trailing port information implies the default port for the service requested (e.g., "80" for an HTTP URL).

Currently, Vibe.d's http client does not respect this.

PR #1510, opened 2016-06-04
Provide port information when the HTTP request isn't on the default port

Geod24:
Fixes rejectedsoftware/vibe.d#1507

### DIFF
--- a/source/vibe/http/client.d
+++ b/source/vibe/http/client.d
@@ -84,7 +84,13 @@ HTTPClientResponse requestHTTP(URL url, scope void delegate(scope HTTPClientRequ
 			}
 			if (settings.proxyURL.schema !is null)
 				req.requestURL = url.toString(); // proxy exception to the URL representation
-			req.headers["Host"] = url.host;
+
+			// Provide port number when it is not the default one (RFC2616 section 14.23)
+			if (url.port && url.port != url.defaultPort)
+				req.headers["Host"] = format("%s:%d", url.host, url.port);
+			else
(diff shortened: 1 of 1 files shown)


Geod24:
@s-ludwig : This will create another allocation when the port is provided.
I am **not so** fan of it (this kind of allocation could be avoided), although currently there are many places which use `~` so its in line with current practice.
Since you're doing a major refacto, do you have some plan for those ?

etcimon:
Maybe lazy loading these headers only when the user calls req.headers, because otherwise I don't see how it's necessary to preload the container when you can get away with writing them to the stream directly when time comes.

Geod24:
@s-ludwig : Ping

s-ludwig:
A simple solution for the current architecture would be to allocate the buffer on the stack, within `requestHTTP`. I'm planning to add an allocation-less lower layer for the HTTP implementation, which would consist of some kind of callback based system that can be used to write headers directly to the wire. This would be the obvious candidate for a revised version.

BTW, what do you think about https://github.com/dlang/dub/pull/866?

Geod24:
Thanks !

> I'm planning to add an allocation-less lower layer for the HTTP implementation, which would consist of some kind of callback based system that can be used to write headers directly to the wire. This would be the obvious candidate for a revised version.

That sounds very promising. Feel free to ping me when you have something ready.

> BTW, what do you think about dlang/dub#866?

Sorry for the late answer. Since I'm subscribed to many thing dlang wise, I may miss some of the notifications.

PR #1511, opened 2016-06-04
Fixup PR #1508: Do case-insensitive comparison of protocol upgrade

Geod24:
The original P.R. only fixed one of the overload.


Geod24:
Ping @s-ludwig

PR #1524, opened 2016-06-17
vibe.core.concurrency: Update references to Isolated template

CyberShadow:
At least it seems like it refers to the same thing.

Not sure how to do the linking correctly here.


s-ludwig:
Thanks. The reference should work (just "Isolated" in backticks works, too, because it's in the same scope).

PR #1526, opened 2016-06-23
Added VibeDataNoOpDispatch version 

intrica:
This should not break existing code but allows us to define a version VibeDataNoOpDispatch. This prevents the compilation of the opDispatch functions in both Bson and Json.

We are semi-regularly running into issues with people using UFCS on a Json object. Instead of calling the function, it uses opDispatch and more often than not returns Json.undefined. This compiles without a problem and sometimes drives you mad when tests fail and you can't see the (not so) obvious issue.

Considering they are deprecated anyway, this allows a nice way for people to throw away this functionality immediately and make better use of things like UFCS.

### DIFF
--- a/source/vibe/data/bson.d
+++ b/source/vibe/data/bson.d
@@ -669,17 +669,22 @@ struct Bson {
 		return 0;
 	}
 
-	/** Deprecated, please use `opIndex` instead.
+	// If VibeDataNoOpDispatch is declared, don't include the depracted functions. This allows UFCS to be used on a Bson object
+	version (VibeDataNoOpDispatch) {
+	}
+	else {
+		/** Deprecated, please use `opIndex` instead.
 
-		Allows to access existing fields of a JSON object using dot syntax.
+			Allows to access existing fields of a JSON object using dot syntax.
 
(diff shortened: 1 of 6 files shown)


s-ludwig:
I just noticed a few days ago that DMD doesn't even evaluate the `deprecated` attribute for `opDispatch` (anymore?), which is really unfortunate. The best thing that I can see for continuing the deprecation path would be to put a `pragma(msg)` inside. Would that also be sufficient in your case? But I wouldn't mind adding the `version` block either.

intrica:
We noticed the lack of deprecated message as well. While a pragma message might be handy, I know we have been ready for a long time to ditch the functions as they are a little dangerous. 

For example, In our controllers we have some functions like validateResponse(Json). We have had to remember to call them by calling validateReponse(jsonPayload) but every now and again somebody will accidentally call jsonPayload.validateResponse - which compiles. We'd love to see the version block included as it's a big pain point.

I've also added a second commit as there were a few spots in the vibe code that were still relying on opDispatch.

s-ludwig:
Okay, I'll merge this when Travis is done and will then add the pragma afterwards.

PR #1532, opened 2016-07-17
Fix vibe.inet.urltransfer.download leaving stale sockets around

Yoplitein:


s-ludwig:
Good catch, I never used `download` for anything serious (just `requestHTTP` directly) and didn't notice this. The only thing is that like this it would mean that keep-alive connections won't work anymore for the case where a custom `HTTPClient` is passed to `download` - something like a `bool is_custom_client = client !is null;` flag is needed.

s-ludwig:
Okay, thanks!

PR #1533, opened 2016-07-17
url router test utility

gedaiu:
Here is a proposal for making testing routes easier. What do you think?

### DIFF
--- a/source/vibe/http/test.d
+++ b/source/vibe/http/test.d
@@ -1,0 +1,349 @@
+/**
+	A simple way to test routes added to a URLRouter
+
+	Copyright: © 2012-2014 RejectedSoftware e.K.
+	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
+	Authors: Szabo Bogdan
+*/
+module crate.request;
+
+import vibe.inet.url;
+import vibe.http.router;
+import vibe.data.json;
+
+import vibe.stream.memory;
+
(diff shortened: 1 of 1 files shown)


s-ludwig:
Looks like a really useful idea. One thing I was wondering (sorry if this is obvious form the implementation, I still have to take a closer look) is if it would be feasible to implement basically the same API, but only using directly `HTTPServerResponse`, by employing some UFCS. Instead of just `URLRouter`, it would also make sense to allow any `HTTPServerRequest(Delegate/Function/Handler)` as the target.

gedaiu:
I'm not sure I understand... where do you expect to send a `HTTPServerRequest`? As I know a server request is represented as an object not as a `Delegate/Function/Handler`.

I will try to remove the test server class and use a `HTTPServerResponse`. I know that I am using this way of testing inside a project and for some reason I was not able to use the `HTTPServerResponse`. I will try again and post here more details :)

gedaiu:
ping @s-ludwig

s-ludwig:
Sorry for losing track of this. I just got reminded of it due to recent [forum thread](https://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/39536/).

So since `URLRouter` itself inherits from `HTTPServerRequestHandler`, it would be a simple generalization to let `request(...)` take a `HTTPServerRequestHandler`, as well as any of the other handler types.

Issue #1505, opened 2016-06-01
Improve and test Websocket


Geod24:
Following recent discussions, just creating an issue to track that.
There seem to be an obvious test suite for websockets: https://github.com/crossbario/autobahn-testsuite
AFAICS, the module code quality could be improved, and so could the documentation. Finally, ping/pong from the client side looks broken.

Geod24:
@s-ludwig : Might be a bit early to ask, but would you consider a split into two module (client/server, as HTTP does) ?

s-ludwig:
I guess it depends on the size of the API. If it becomes necessary just because of implementation size, I'd rather go for a `http.internal` package (same rationale as for `web.internal`.

Geod24:
My original thought was that it could be worthwhile to separate client and server.
However, since it's a bidirectional connection, and after another half day of fiddling, I'd rather wait a bit.

The current client approach seems to rely on the implementer knowing which kind of message it receives - binary or text -.  I don't know how well that assumption will do in the real world, but as part of testing, I'd like a way to dispatch binary and text message to separate method, and it's not there. From a quick look, the same issue seems present in the server as well.

For reference, this is the kind of API Java provides: https://github.com/crossbario/autobahn-android/blob/e5abf7c64c9327ff2515a831d4e6125f716b69ee/Demo/TestSuiteClient/src/de/tavendo/autobahn/testsuiteclient/MainActivity.java#L104-L126

Geod24:
Solved in #1534

PR #1534, opened 2016-07-17
Websockets: Add a client test again the autobahn testsuite + improvements

Geod24:
Here's a couple of improvements to the Websockets implementation.

First, a new 'test', which is not run by default, against the autobahn testsuite (see https://github.com/rejectedsoftware/vibe.d/issues/1505 ).
Currently, it passes 517 out of 519 test cases (it blocks on the 518th, haven't investigated yet).

The other improvements are mostly cosmetic / performance: reduce the size of the `Frame`, document everything, provide a better encapsulation...

There's a change to the API, where 'send' with a message doesn't assume the data to be text anymore, and `send` didn't accept `const(ubyte)[]` - even though it was just forwarding to a const-accepting method.

### DIFF
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -20,7 +20,9 @@ if [ ${BUILD_EXAMPLE=1} -eq 1 ]; then
 fi
 if [ ${RUN_TEST=1} -eq 1 ]; then
     for ex in `\ls -1 tests/`; do
-        echo "[INFO] Running test $ex"
-        (cd tests/$ex && dub --compiler=$DC && dub clean)
+        if [ -r test/$ex/dub.json ] || [ -r test/$ex/dub.sdl ]; then
+            echo "[INFO] Running test $ex"
+            (cd tests/$ex && dub --compiler=$DC && dub clean)
+        fi
     done
 fi
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -409,7 +409,7 @@ final class WebSocket {
 		On the JavaScript side, the text will be available as message.data (type Blob).
 		Throws: WebSocketException if the connection is closed.
 	*/
-	void send(ubyte[] data)
+	void send(in ubyte[] data)
 	{
 		send((scope message){ message.write(data); }, FrameOpcode.binary);
 	}
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -1,7 +1,7 @@
 /**
 	Implements WebSocket support and fallbacks for older browsers.
 
-    Standards: $(LINK2 https://tools.ietf.org/html/rfc6455, RFC6455)
+	Standards: $(LINK2 https://tools.ietf.org/html/rfc6455, RFC6455)
 	Copyright: © 2012-2014 RejectedSoftware e.K.
 	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
 	Authors: Jan Krüger
@@ -311,7 +311,9 @@ final class WebSocket {
 		uint m_lastPingIndex;
 		bool m_pongReceived;
 		bool m_pongSkipped;
-        SystemRNG m_rng;
+		/// The entropy generator to use
+		/// If not null, it means this is a server socket.
+		SystemRNG m_rng;
 	}
 
 	/**
@@ -574,7 +576,7 @@ final class WebSocket {
 */
 final class OutgoingWebSocketMessage : OutputStream {
 	private {
-        SystemRNG m_rng;
+		SystemRNG m_rng;
 		Stream m_conn;
 		FrameOpcode m_frameOpcode;
 		Appender!(ubyte[]) m_buffer;
@@ -586,7 +588,7 @@ final class OutgoingWebSocketMessage : OutputStream {
 		assert(conn !is null);
 		m_conn = conn;
 		m_frameOpcode = frameOpcode;
-        m_rng = rng;
+		m_rng = rng;
 	}
 
 	void write(in ubyte[] bytes)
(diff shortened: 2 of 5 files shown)


Geod24:
Fixed the failures. Now I only get the spurious ones.

mathias-lang-sociomantic:
Ping @s-ludwig

Geod24:
Ping @s-ludwig

source/vibe/http/websockets.d
```diff
@@ -418,7 +425,8 @@ final class WebSocket {
 		Sends a message using an output stream.
 		Throws: WebSocketException if the connection is closed.
 	*/
-	void send(scope void delegate(scope OutgoingWebSocketMessage) sender, FrameOpcode frameOpcode = FrameOpcode.text)
```
s-ludwig: There should be an additional compatibility wrapper here to avoid breaking code:

    /// Compatibility overload - will be removed soon.
    deprecated("Need to specify an explicit FrameOpcode.")
    void send(scope void delegate(scope OutgoingWebSocketMessage) sender)
    {
        send(sender, FrameOpcode.text);
    }

source/vibe/http/websockets.d
```diff
@@ -654,6 +662,27 @@ final class IncomingWebSocketMessage : InputStream {
...
+	 *
+	 * This function is helpful if one wish to process frames by frames,
+	 * or minimize memory allocation, as `peek` will only return the current
+	 * frame data, and read requires a pre-allocated buffer.
+	 *
+	 * Returns:
+	 * `false` if the current frame is the final one, `true` if a new frame
+	 * was read.
+	 */
+	bool nextFrame()
```
s-ludwig: Nitpick: `skipFrame()` would be more consistent with `skip()`, which is used by other stream types.

source/vibe/http/websockets.d
```diff
@@ -697,108 +727,234 @@ final class IncomingWebSocketMessage : InputStream {
 	}
 }
 
+/// Magic string defined by the RFC for challenging the server during upgrade
+private static immutable s_webSocketGuid = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
```
s-ludwig: Does `static` actually have an effect for `private` symbols?

source/vibe/http/websockets.d
```diff
@@ -697,108 +727,234 @@ final class IncomingWebSocketMessage : InputStream {
...
 struct Frame {
-	bool fin;
-	FrameOpcode opcode;
+	/**
+	   Contains the first byte of the frame:
+	   - The FIN bit
+	   - The 3 reserved bits RSV1, RSV2, RSV3
+	   - The 4-bits opcode
+	*/
+	ubyte fin_rsv_opcode;
```
s-ludwig: This could probably benefit from using `bitfields` instead of manual bit fiddling, but both are generally fine for me in this case.

source/vibe/http/websockets.d
```diff
@@ -697,108 +727,234 @@ final class IncomingWebSocketMessage : InputStream {
...
+			stream.read(data[0 .. 2]);
+			length = bigEndianToNative!ushort(data[0 .. 2]);
+		} else if (length == 127) {
+			stream.read(data);
+			length = bigEndianToNative!ulong(data);
+
+			// RFC 6455, 5.2, 'Payload length': If 127, the following 8 bytes
+			// interpreted as a 64-bit unsigned integer (the most significant
+			// bit MUST be 0)
+			enforceEx!WebSocketException(!!(length >> 63),
```
s-ludwig: Should be just a single `!` here.

travis-ci.sh
```diff
@@ -20,7 +20,9 @@ if [ ${BUILD_EXAMPLE=1} -eq 1 ]; then
 fi
 if [ ${RUN_TEST=1} -eq 1 ]; then
     for ex in `\ls -1 tests/`; do
-        echo "[INFO] Running test $ex"
-        (cd tests/$ex && dub --compiler=$DC && dub clean)
+        if [ -r test/$ex/dub.{json,sdl} ]; then
```
s-ludwig: Bash complains for Travis-CI:
> '[' -r test/vibe.core.net.1376/dub.json test/vibe.core.net.1376/dub.sdl ']'
> ./travis-ci.sh: line 23: [: test/vibe.core.net.1376/dub.json: binary operator expected

Should probably use `[ -r test/$ex/dub.json} ] || [ -r test/$ex/dub.sdl ]`.

source/vibe/http/websockets.d
mathias-lang-sociomantic: `static immutable` does: It means the variable is a CT-known constant that will end up in the data section. It is quite useful for arrays, for example:

```D
static immutable data = [1, 2, 3, 4];

void process (immutable int[]) @nogc;

void main () @nogc
{
    process(data);
}
```

If one used `enum` here, `main` couldn't be `@nogc`.

travis-ci.sh
mathias-lang-sociomantic: Oops, good catch !

source/vibe/http/websockets.d
s-ludwig: I always thought that this will always happen for `immutable` literals (at the least for string literals, not sure about arrays).

mathias-lang-sociomantic: Actually at module level `immutable x = y` still have the same behavior. Do you want me to revert that change ?

s-ludwig: Nah, doesn't really matter, I was just wondering.

mathias-lang-sociomantic:
Rebased on master and addressed your comments

mathias-lang-sociomantic:
Oh, thank you very much ! Sorry this went off the radar a bit, I don't actively work on that project currently, since I had many issues (notably the fact that the bot would only survive a minute).
I should try again to see if everything is fixed now.

s-ludwig:
Good that you mentioned it again! I actually remembered that I observed the same issue in one of my projects (it does an auto-reconnect anyway, so it didn't seem so important) and now pushed a fix for this: #1848

PR #1537, opened 2016-07-20
`assert(0, "msg")` instead of `assert("msg")`

CromFr:
`static assert("msg")` will never fail


s-ludwig:
Wow, that went undiscovered for a long time now. Thanks!

PR #1542, opened 2016-08-04
Let serializers get a type traits struct instead of the raw type.

s-ludwig:
This is a breaking change for any serializer implementations. The expectation though is that the only existing seriailzer implementations are currently the ones in vibe.d itself.

The traits struct in particular carries UDA information, which allows serializers to support serializer specific UDAs. The code has also been refactored a bit to reduce line length/visual noise.

### DIFF
--- a/source/vibe/data/bson.d
+++ b/source/vibe/data/bson.d
@@ -1347,34 +1347,34 @@ struct BsonSerializer {
 		return ret;
 	}
 
-	void beginWriteDictionary(T)()
+	void beginWriteDictionary(Traits)()
 	{
 		writeCompositeEntryHeader(Bson.Type.object);
 		m_compositeStack ~= m_dst.data.length;
 		m_dst.put(toBsonData(cast(int)0));
 	}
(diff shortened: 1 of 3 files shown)


s-ludwig:
Cleaned up the diff and added `beginWriteDocument`/`endWriteDocument`, `beginReadArrayEntry`/`endReadArrayEntry` and `beginReadDictionaryEntry`/`endReadDictionaryEntry`.

skoppe:
The changes look pretty straightforward. I see no issues. (btw good idea to separate T => Traits rename)

s-ludwig:
Thanks for reviewing! I still often do not manage to focus on doing a single thing at a time, so I had to rework the commit history quite heavily to get it to a reviewable state ;-)

BTW, I'll get to #1547 shortly (still in vacation, but maybe tomorrow when we are on the ferry), but it looks like what I had in mind for such an implementation.

skoppe:
Hehe, I thought the best way to get your attention would be to review something :)

PR #1545, opened 2016-08-10
Add unix-socket support for connectTCP for libevent

skoppe:


skoppe:
oops. added binary

examples/unix_socket/dub.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "unit-socket-example",
+	"description": "Example for sending http requests to unix sockets",
+	"dependencies": {
+		"vibe-d:http": {"path": "../../"},
+		"vibe-d:web": {"path": "../../"},
+		"vibe-d:core": {"path": "../../"}
+	},
+	"versions": ["VibeCustomMain", "VibeNoSSL"],
+	"subConfigurations": { "vibe-d:core": "libasync"}
```
s-ludwig: `"VibeCustomMain"` can be dropped here (it's the default now).

The libasync based driver is selected here, although the changes are for the libevent one - is this intended? Does the libasync driver work?

source/vibe/core/drivers/libevent2_tcp.d
```diff
@@ -83,9 +83,12 @@ package final class Libevent2TCPConnection : TCPConnection {
...
 		if( ctx.remote_addr.family == AF_INET ) ptr = &ctx.remote_addr.sockAddrInet4.sin_addr;
-		else ptr = &ctx.remote_addr.sockAddrInet6.sin6_addr;
-		evutil_inet_ntop(ctx.remote_addr.family, ptr, m_peerAddressBuf.ptr, m_peerAddressBuf.length);
-		m_peerAddress = cast(string)m_peerAddressBuf[0 .. m_peerAddressBuf[].indexOf('\0')];
+		else if (ctx.remote_addr.family != AF_UNIX ) ptr = &ctx.remote_addr.sockAddrInet6.sin6_addr;
+		else ptr = &ctx.remote_addr.sockAddrUnix.sun_path;
+
+		if (evutil_inet_ntop(ctx.remote_addr.family, ptr, m_peerAddressBuf.ptr, m_peerAddressBuf.length) !is null)
+			m_peerAddress = cast(string)m_peerAddressBuf[0 .. m_peerAddressBuf[].indexOf('\0')];
+		// TODO: do we need to set m_peerAddress in else case? (when family is AF_UNIX)
```
s-ludwig: The peer address is used in some places to construct URLs, but I'm not sure how much sense it makes in this context. Probably safe to skip it for now.

examples/unix_socket/source/app.d
```diff
@@ -0,0 +1,13 @@
+import vibe.inet.url;
+import vibe.http.client;
+import vibe.stream.operations;
+import std.stdio;
+
+void main()
+{
+	URL url = URL("unix:///var/run/docker.sock") ~ Path("containers/json");
```
s-ludwig: This should be "http+unix:///var/run/docker.sock...", so that it's still possible to select the proper application protocol. Is there any precedence for such a URL? From what I've found, and so that the URL has a unique string representation (`URL(url.toString()) == url`), the best way to represent it is `"http+unix://%2Fvar%2Frun%2Fdocker.sock/containers/json"` and then using `urlDecode` on the `url.host` value before putting it in the UNIX address struct.

source/vibe/core/net.d
```diff
@@ -91,25 +94,28 @@ TCPListener listenTCP_s(ushort port, void function(TCPConnection stream) connect
...
-	addr.port = port;
+	if (addr.family != AF_UNIX)
+		addr.port = port;
 	NetworkAddress bind_address;
 	if (bind_interface.length) bind_address = resolveHost(bind_interface, addr.family);
 	else {
-		bind_address.family = addr.family;
-		if (addr.family == AF_INET) bind_address.sockAddrInet4.sin_addr.s_addr = 0;
-		else bind_address.sockAddrInet6.sin6_addr.s6_addr[] = 0;
+		bind_address.family = addr.family;// == AF_UNIX ? AF_INET : addr.family;
```
s-ludwig: left-over comment ^

source/vibe/core/net.d
```diff
@@ -91,25 +94,28 @@ TCPListener listenTCP_s(ushort port, void function(TCPConnection stream) connect
...
 }
 /// ditto
 TCPConnection connectTCP(NetworkAddress addr, NetworkAddress bind_address = anyAddress)
 {
 	if (bind_address.family == AF_UNSPEC) {
-		bind_address.family = addr.family;
-		if (addr.family == AF_INET) bind_address.sockAddrInet4.sin_addr.s_addr = 0;
-		else bind_address.sockAddrInet6.sin6_addr.s6_addr[] = 0;
-		bind_address.port = 0;
+		bind_address.family = addr.family;// == AF_UNIX ? AF_INET : addr.family;
```
s-ludwig: another one ^

Issue #1544, opened 2016-08-08
GzipOutputStream with requestHTTP


skoppe:
I am hitting some limits of plain text transfer and would like to gzip it. Any possibility of using the GzipOutputStream with requestHTTP?

If not implemented yet, I was thinking about using a similar method like in the HTTPClientResponse. There the OutputStream type for the bodyReader gets determined by some headers (chunked, gzip, limited, etc.).

For the HTTPClientRequest it would then work like this: if you would set the appropriate gzip header before calling the bodyReader, it would then be instantiated with a GzipOutputStream upon calling the bodyReader property.

How 'bout that?

skoppe:
Nevermind, this doesn't make sense.

PR #1546, opened 2016-08-11
set gzip output stream in HTTPClientRequest's bodyWriter()

skoppe:
when content-encoding is set to gzip
#1544


skoppe:
Are the travis builds flaky or what?

skoppe:
Nevermind, gzipping requests obviously doesn't make sense.

Issue #1442, opened 2016-02-24
Empty Json serializes to "undefined"


SmorkalovG:
Why is Json().toString "undefined"? It is not a valid json, I can not pass it to the client.

SmorkalovG:
Accidentally created the issue before wrote body. Sorry.
I am trying to use REST interface generator. The result of some method is:
struct Test {
	ushort port;
	string path;
	Json arguments;
}
If method returns Test(8081, "/", Json()) then serialized json is {"port":8081,"path":"/","arguments":undefined}. It isn't valid. May be it is better to serialize Json() to {} or null?
If not, is there another way to add optional field to struct with arbitrary data?

s-ludwig:
Try to use `Json arguments = Json.emptyObject;` in this case. `Json.undefined` is used, because everything else would be arbitrary, it could otherwise mean any of `null`, `{}` or `[]`, but which of these makes sense depends on the situation.

If you need an optional field, you have two other possibilities:
- Use the `@optional` attribute on the field in question, if optionality on the receiving side is enough (it will always be serialized when serializing)
- Use `Nullable!T`. In this case, `null` will be used to denote the missing field.

It would be nice if Phobos also had an `Optional!T` type. I'd add support for it, so that the field is completely omitted if the `Optional` isn't set.

SmorkalovG:
Thank you for answer. Now I understand how to solve my problem.
But I still don't understand why it is possible to get invalid json as result of serialization. In my point of view it is a design problem. I think nothing should lead to invalid json. And there are not many ways to get rid of this behavior. Json().toString must throw an exception or return valid json (i.e. {} or null). Are there any reasons why it is "undefined"? Because it is human readable?

s-ludwig:
Oh, you are absolutely right, I missed that part. There are related tickets (#222, #958) and a partial PR (#1323). I think the best solution would be to follow JavaScript's rules for this - if an object has a field with an undefined value, skip it during serialization and else output `null` instead.

SmorkalovG:
Sorry, I thought PR fixes this situation. It is better if issue will be opened until it is fixed.

s-ludwig:
Fixed by https://github.com/rejectedsoftware/vibe.d/pull/1548

Issue #958, opened 2015-01-18
data.json: nan should not be serialized to "undefined"


tomykaira:
Problem: The serialization object including `NaN` is invaild JSON.

```d
auto j = Json.emptyObject;
j["float"] = Json(double.init);
assert(j.toString == `{"float":undefined}`);  // literal "undefined" is invalid
```

According to [existing unittests](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/data/json.d#L1721), this is intended behavior, but IMO, not acceptable as implementation of JSON serializer.
Serialization must not create invalid JSON data without error notification.

I understand it is difficult to approach from compatibility point of view, but I appreciate if one of following fixes are introduced.

- Throw an exception because it is `NaN` input is not permitted [1]
- Serialize `NaN` to `null` (as ECMAScript does [2])

1. [RFC 4627 - The application/json Media Type for JavaScript Object Notation (JSON)](http://tools.ietf.org/html/rfc4627) 2.4. Numbers
2. [PDF: Final draft Standard ECMA-262 edition 5.1, March 2011 (Rev. 6)](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf) p.208

NfNitLoop:
There's a similar issue with the default `undefined` value in Json: 

~~~
Json j;
assert(j.toString == "undefined")
~~~

These kinds of values are time bombs for JSON in D.  They give neither compile-time nor runtime errors, they just result in invalid output that can't be parsed by REST clients!

intrica:
Okay so when I did this, I also made another change to our local fork of vibe.d which I've had to reapply for every upgrade. I wasn't sure if people would like the behaviour, but here is my logic.

For a JSON object, a value that is not set should simply not appear in the structure. I don't particularly see why we should have 'null' in there, anymore than we should have undefined. If the key isn't set, it shouldn't exist. For a value, I agree that null is the better value to set to. 

We worked with this by modifying writeJsonString. A NaN is output as null, and ignored if it is part of an object.

Now some people may not like this idea as if something is missing from the structure, it needs to be optional. But if you are going to allow NaN to be serialised, then it really is an optional value anyway.

If you guys are okay with the idea of just modifying the serialised version of the string, then I'm happy to submit my branch as a pull request.

s-ludwig:
Fixed by https://github.com/rejectedsoftware/vibe.d/pull/1548

[...]

PR #1548, opened 2016-08-13
nan -> null not undefined

John-Colvin:
should add some more tests, but I believe this is a necessary change. Either this or throw an exception on serialising a nan. `undefined` is not a valid json literal and should never appear in any JSON output.


s-ludwig:
Thanks, this should have been fixed long ago. Fixes #1442 and #958. There is also a related PR that I lost track of (#1323, sorry @rjmcguire for having missed your comment). Can you integrate that change here, too (i.e. the same change, but in the `JsonStringSerializer`)?

John-Colvin:
ok, done. I guess I need to test that, not sure where

s-ludwig:
Thanks! `assert(serializeToJsonString(double.nan) == "null");` should be a valid test (I'd place it directly below the `JsonStringSerializer` struct).

John-Colvin:
done

s-ludwig:
LGTM!

Issue #1529, opened 2016-07-07
HTTPServerRequest needs some custom storage


schveiguy:
Currently, there is the `params` member, which can store custom strings to be passed through the chain of processors.

However, there are plenty of use cases where I want to store something other than a string in the request data. For instance, I want to store the user id in the session, and early on in the routing, I want to fetch all the information about the user from the database, and store that User object in the request, for further access.

I don't want to cache the user data in the session, since it may change in the DB. And I don't want to fetch that data from the database at every step in the router chain.

What I'm going to do for now, is to use the `params` member, with the value "string" being 0 length and pointing at some object that I really know what it is. And just add casts everywhere.

This should work, but god is it ugly! `HTTPServerRequest` desperately needs some mechanism to store custom types according to an id or string. I understand from searching the forums that this is a recognized problem, and the backwards compatibility requires that we don't muck with how params works. I propose to add a new variable that is a `Variant[string]`, or even `void*[string]`, called `context` or something like this (hell, even if we have a `void * context` member), so that I can simply pass data around.

bheads:
Java has attributes as part of the ServletRequest Spec:

https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/ServletRequest.html#setAttribute(java.lang.String, java.lang.Object)

This would be a very useful addition!

thaven:
The Variant[string] (or a Variant typed DictionaryList) is probably the most useful. void* would be bad, because it can store only one value, so multiple libraries using it would cause trouble for sure!

s-ludwig:
Fixed by 9d0a75b838b53bf281f21a0dd8209b508a2a0b67.

PR #1550, opened 2016-08-16
Add context field to HTTPServerRequest

thaven:
The context field allows application specific data to be passed down the chain of request processors.
Solves rejectedsoftware/vibe.d#1529

### DIFF
--- a/source/vibe/http/server.d
+++ b/source/vibe/http/server.d
@@ -627,6 +627,20 @@ final class HTTPServerRequest : HTTPRequest {
 		*/
 		DictionaryList!(string, true, 8) params;
 
+        import std.variant : Variant;
+        /** A map of context items for the request.
+
+            This is especially useful for passing application specific data down
+            the chain of processors along with the request itself.
+
+            For example, a generic route may be defined to check user login status,
+            if the user is logged in, add a reference to user specific data to the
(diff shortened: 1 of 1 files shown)


wilzbach:
@thaven your PR needs a rebase.

@s-ludwig what's your opinion on this? On #1529 there are many people in favor of a way to store custom information within the current request.

s-ludwig:
Definitely makes sense, I even remember having implemented this once, but for some reason it then got deferred. I'm just trying to remember now which name I've used and whether "context" is any better or worse.

One thing regarding the implementation: `DictionaryList!(Variant, true, 4)` would probably be a better idea than `Variant[string]`, so that a few entries can be set without additional GC allocations.

schveiguy:
> One thing regarding the implementation

Sounds good, I really look forward to this update! @s-ludwig, you can just edit the PR branch, no? It's possible @thaven isn't working on this any more as the PR is so old.

thaven:
I do not have much spare time at the moment, so feel free to edit the PR.
@s-ludwig: Using DictionaryList seems a good idea to me.

s-ludwig:
@thaven: Alright, I'll change to `DictionaryList` and merge!

PR #1551, opened 2016-08-18
disallow defining VibeCustomMain and VibeDefaultMain together

John-Colvin:
I don't know how, but I ended up with both of these defined and it caused a headache, perhaps this will help others see the problem quickly


s-ludwig:
Good point. This check in similar form was actually there a while ago, but it got lost when `VibeCustomMain` became a no-op.

PR #1563, opened 2016-09-16
ConnectionPool's LocalTaskSemaphore Intermittent Infinite Wait Bug

etcimon:
When the semaphore reached the lock limit, tasks can compete for a lock but the waiter sometimes ended up having its lock snatched. This fixes the issue by freeing the lock only if no waiter was called.


s-ludwig:
LGTM!

Issue #1564, opened 2016-09-18
rest js client wrong code generation


deviator:
If I want use `Collections` or simple classes in js client I get errors.
Generated js code have more that one mistake:
```
IModel = new function() {
var toRestString = function(v) { return v; } // bad serialization for structures (js-objects) if use HTTP GET
calculator = new function() { // here "calculator" must be "this.calculator" because it's a method of IModel
var toRestString = function(v) { return v; } 
  this.calc = function(_name, points, on_result, on_error) {
    var url = "http://127.0.0.1:8080/calculator/:name/calc"; // here no replacement of ":name" parameter
...
```

PR #1566, opened 2016-09-18
fix some rest js client problems

### LINKED ISSUE #1565: error in serveRestJSClient
[here](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/web/rest.d#L246) and [here](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/web/rest.d#L253) not specified template parameter `I`

deviator:
fix #1565
some problem from #1564 fix also

### DIFF
--- a/source/vibe/web/internal/rest/jsclient.d
+++ b/source/vibe/web/internal/rest/jsclient.d
@@ -37,7 +37,7 @@ import std.conv : to;
 	auto inner1 = indent ~ indentStep;
 	auto inner2 = inner1 ~ indentStep;
 
-	output.put(indent ~ "var toRestString = function(v) { return v; }\n");
+	output.put(indent ~ "var toRestString = function(v) { return JSON.stringify(v); }\n");
 
 	foreach (i, SI; intf.SubInterfaceTypes) {
 		output.put("\n");
@@ -70,17 +70,19 @@ import std.conv : to;
 				else output.formattedWrite("encodeURIComponent(toRestString(%s))", p.text);
 			}
 		} else {
-			import std.algorithm;
-			auto rpn = route.parameters.map!"a.name".filter!(a=>a[0]=='_').map!"a[1..$]";
-			if (rpn.save.count == 0)
-				output.formattedWrite("%s", Json(concatURL(intf.baseURL, route.pattern.dup)));
-			else {
-				import std.array;
-				char[] sink = route.pattern.dup;
-				foreach (param; rpn)
-					sink = replace(sink, ":" ~ param, "${_" ~ param ~ "}");
-				output.formattedWrite("`%s`", concatURL(intf.baseURL, sink.idup)); // use `` instead ""
-			}
+			import std.algorithm : filter, map;
+			import std.range : chain;
+			import std.array : replace;
+			import std.typecons : tuple;
+			auto rpn = route.parameters
+				.map!"a.name"
+				.filter!(a => a[0] == '_')
+				.map!"a[1..$]"
+				.map!(a => tuple(":"~a, "${_"~a~"}"));
+			char[] sink = route.pattern.dup;
+			foreach (p; rpn)
+				sink = replace(sink, p[0], p[1]);
+			output.formattedWrite("`%s`", concatURL(intf.baseURL, sink.idup)); // use `` instead ""
 		}
 		output.put(";\n");
 
@@ -97,7 +99,7 @@ import std.conv : to;
 		if (route.bodyParameters.length) {
 			output.put(inner1 ~ "var postbody = {\n");
 			foreach (p; route.bodyParameters)
-				output.formattedWrite("%s%s: toRestString(%s),\n", inner2, Json(p.fieldName), p.name);
+				output.formattedWrite("%s%s: %s,\n", inner2, Json(p.fieldName), p.name);
 			output.put(inner1 ~ "};\n");
 		}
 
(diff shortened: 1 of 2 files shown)


deviator:
ping @s-ludwig

s-ludwig:
The Travis failure is due to a flaky test - I've restarted the test run now and will remove that specific test.

Changes look good, except for the mentioned if/else swap. Maybe the three additional parameters could be grouped inside of a `JSRestInterfaceSettings` class - it seems like more options could come in handy later on.

deviator:
@s-ludwig it's ok?

WowVeryLogin:
Ping @s-ludwig I've encountred bug, described in issue #1564 and it's very important for owr company to continue to work on project, so pls fix this stuff as soon as you can

s-ludwig:
Looks good to merge, thanks! I'll just make the `parent` parameter private for now/remove it from the settings class, so that it doesn't become a burden w.r.t. backwards compatibility.

source/vibe/web/internal/rest/jsclient.d
```diff
@@ -30,81 +58,98 @@ import std.conv : to;
...
-				if (!p.isParameter) output.serializeToJson(p.text);
-				else output.formattedWrite("encodeURIComponent(toRestString(%s))", p.text);
+				fout.put(" + ");
+				if (!p.isParameter) fout.serializeToJson(p.text);
+				else fout.formattedWrite("encodeURIComponent(toRestString(%s))", p.text);
 			}
-			output.put(";\n");
 		} else {
-			output.formattedWrite("    var url = %s;\n", Json(concatURL(intf.baseURL, route.pattern)));
+			auto rpn = route.parameters
```
s-ludwig: Stupid new review feature has eaten my earlier comment...

So this part is strange, because the upper part of the if statement is supposed to be doing what you did in the `else` part. Maybe there is an issue with `route.hasPathPlaceholders`.

deviator: I was look to this part, but was confusing by [this](https://github.com/rejectedsoftware/vibe.d/blob/master/source/vibe/web/internal/rest/common.d#L176) and I don't touch this, because I think change in this part can have more complex effects. Is this a bug or only first `id` parameter can be a placeholders by design? If it by design it's doubtful (or I don't understand somethings).

s-ludwig: That was indeed broken. I've committed a fix for the `hasPathPlaceholders` computation: https://github.com/rejectedsoftware/vibe.d/commit/f23d7740e7e459e6a890cf6d7ef64bb1b5640402
And also one for the JS URL computation: https://github.com/rejectedsoftware/vibe.d/commit/10e37edd880e3790e5cc6ff7ee89e3018348b3ad

Can you remove the changes in the else block from the PR and rebase?

deviator:
ping @s-ludwig

source/vibe/web/internal/rest/jsclient.d
```diff
@@ -31,85 +59,94 @@ import std.conv : to;
...
-			output.serializeToJson(burl.toString()[0 .. $-1]);
+			fout.serializeToJson(burl.toString()[0 .. $-1]);
 			// and then assemble the full path piece-wise
 			foreach (p; route.fullPathParts) {
-				output.put(" + ");
-				if (!p.isParameter) output.serializeToJson(p.text);
-				else output.formattedWrite("encodeURIComponent(toRestString(%s))", p.text);
+				fout.put(" + ");
+				if (!p.isParameter) fout.serializeToJson(p.text);
+				else fout.formattedWrite("encodeURIComponent(%s)", p.text);
```
s-ludwig: I believe the `toRestString` must conceptually stay here, even if the function itself is currently not implemented. At some point it must be made equal to the one defined in the D code base.

deviator: I implement this function as [JSON.stringify](https://github.com/rejectedsoftware/vibe.d/pull/1566/files/5745ca3a0af6dbd5e56b3dd08619850f5e69de19#diff-00965ba02c942ce6f16f4106fcd28900R68), because if you want use objects as a parameter of get request you get `object.Object` in params instead `{"x":2, "y":3}` for example. If use it in this case result url has quotes symbols. Maybe need add `toGetParameter` for get request?

s-ludwig: The correct semantics are these:
```
	string toRestString(Json value)
	{
		switch (value.type) {
			default: return value.toString();
			case Json.Type.Bool: return value.get!bool ? "true" : "false";
			case Json.Type.Int: return to!string(value.get!long);
			case Json.Type.Float: return to!string(value.get!double);
			case Json.Type.String: return value.get!string;
		}
	}
```
I.e. mostly strings are special cased to omit the quotes. Otherwise it matches `JSON.stringify`.

deviator: All of calls `toRestString` returns value to `encodeURIComponent`, maybe `encodeURIComponent` must calls inside of `toRestString`?

s-ludwig: That could indeed be done.

deviator:
I beginner in js and can make a mistake, but on my examples it's code works.

s-ludwig:
Looks good, except that arrays also need to be stringified (I'd reverse the condition and stringify everything that is *not* a string).

deviator:
[Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof) arrays `typeof` is `"object"`

s-ludwig:
Oh, alright, didn't expect that! I'll merge then once Travis has finished. Thanks for all the quick follow-up changes!

PR #1572, opened 2016-09-25
Fix FixedRingBuffer.removeAt

tchaloupka:
There is a problem when removing from the start index

### DIFF
--- a/source/vibe/utils/array.d
+++ b/source/vibe/utils/array.d
@@ -374,8 +374,9 @@ struct FixedRingBuffer(T, size_t N = 0, bool INITIALIZE = true) {
 	void removeAt(Range r)
 	{
 		assert(r.m_buffer is m_buffer);
+		if (r.m_start == m_start) { popFront(); return; }
 		if( m_start + m_fill > m_buffer.length ){
-			assert(r.m_start >= m_start && r.m_start < m_buffer.length || r.m_start < mod(m_start+m_fill));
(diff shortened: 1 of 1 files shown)


source/vibe/utils/array.d
```diff
@@ -376,7 +376,7 @@ struct FixedRingBuffer(T, size_t N = 0, bool INITIALIZE = true) {
 		assert(r.m_buffer is m_buffer);
 		if( m_start + m_fill > m_buffer.length ){
 			assert(r.m_start >= m_start && r.m_start < m_buffer.length || r.m_start < mod(m_start+m_fill));
-			if( r.m_start > m_start ){
+			if( r.m_start >= m_start ){
```
s-ludwig: Would probably be a nice optimization to do `popFront` in case of `r.m_start == m_start`.

s-ludwig:
Sounds like this might have been responsible for some obscure bugs!

source/vibe/utils/array.d
tchaloupka: Good point, fixed

s-ludwig: Thanks! Will merge when Travis is done.

tchaloupka:
I just squashed the PR.

s-ludwig:
Just discovered that GitHub now also offers this as an option for how to merge the PR, which is pretty nice for avoiding to bother the PR authors with this!

PR #1574, opened 2016-09-26
Add build configuration for Meson

ximion:
Hello!
I am going to use Vibe.d in a project which is build using Meson, and also would like to make Vibe.d available in Debian/Ubuntu at some point.

Especially for the latter, we need some more sophisticated build system than Dub currently is. We need to install built software system-wide, generate proper pkgconfig files, allow to inject custom compile flags for the respective D compiler instead of using upstream's settings and most importantly using dependencies that are part of the system exclusively instead of downloading them from the web.

Dub can't do that (bugs filed with nothing happening), but Meson can do that and more.
Other options for this task would have been Automake or CMake, but Meson is much easier to use for this purpose (compiling D code) than the other two. You can find more information about Meson at http://mesonbuild.com/, at time it is being worked on to likely replace Automake in GNOME projects.

The above patch is not 100% complete, I focused on the most important bits:
 * Compile Vibe.d modules as (shared) libraries
 * Install libraries and pkgconfig files for them
 * Install the Vibe source code for other projects to use
 * Add `unittest {}` support by generating a test binary and executing it
 * Find & use system copies of OpenSSL and LibEvent D bindings, if they are not found, automatically download & use a Git copy.

Missing pieces are:
 * Windows support (I have no way to test that, maybe it already works)
 * Options, e.g. disable tests or compile without OpenSSL or switch LibEvent for a different implementation.

This patch can be tested by:
 * Obtaining Meson (`sudo apt install meson`, or just Git-cloning it)
 * Ensure you have at least version `0.34.0` to ensure D support is present
 * Run `mkdir build && cd build`, then run `meson ..`
 * Run `ninja`
(`ninja` is a build-tool similar to make, but parallelized by default and much simpler, it is used to build several bigger projects, like Google Chrome)

Known issues:
 * Since the D compilers at time don't output Make-compliant depfiles (hello LDC, DMD!), or have broken support for it (hello, GDC!) Ninja sometimes doesn't recompile pieces that need rebuilding (e.g. because a template or function has changed) - in that case, a `ninja clean && ninja` can work around that. Aside from that, I am not aware of any issues.

I went for one very large Meson configuration here - if wanted, this can be split up into more, smaller parts.

Thanks for considering this PR!

Cheers,
    Matthias

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,421 @@
+project('Vibe.d', 'd')
+
+source_root = meson.source_root()
+build_root = meson.build_root()
+
+project_version      = '0.7.30'
+project_version_name = '0.7.30~beta.1'
+project_soversion    = '0'
+
+pkgc = import('pkgconfig')
+
+#
+# Sources
+#
+src_dir = include_directories('source/')
+
+vibe_main_src = [
+    'source/vibe/appmain.d',
+    'source/vibe/d.d',
+    'source/vibe/vibe.d'
+]
+install_headers(vibe_main_src, subdir: 'd/vibe')
+
+vibe_core_src = [
+    'source/vibe/core/task.d',
+    'source/vibe/core/sync.d',
+    'source/vibe/core/concurrency.d',
+    'source/vibe/core/file.d',
+    'source/vibe/core/log.d',
+    'source/vibe/core/connectionpool.d',
+    'source/vibe/core/args.d',
+    'source/vibe/core/stream.d',
+    'source/vibe/core/drivers/libasync.d',
+    'source/vibe/core/drivers/utils.d',
+    'source/vibe/core/drivers/native.d',
+    'source/vibe/core/drivers/threadedfile.d',
+    'source/vibe/core/drivers/libevent2.d',
+    'source/vibe/core/drivers/winrt.d',
+    'source/vibe/core/drivers/timerqueue.d',
+    'source/vibe/core/drivers/win32.d',
+    'source/vibe/core/drivers/libevent2_tcp.d',
+    'source/vibe/core/net.d',
+    'source/vibe/core/core.d',
+    'source/vibe/core/driver.d'
+]
+install_headers(vibe_core_src, subdir: 'd/vibe/core')
+
+vibe_crypto_src = [
+    'source/vibe/crypto/passwordhash.d',
+    'source/vibe/crypto/cryptorand.d'
+]
+install_headers(vibe_crypto_src, subdir: 'd/vibe/crypto')
+
+vibe_inet_src = [
+    'source/vibe/inet/path.d',
+    'source/vibe/inet/url.d',
+    'source/vibe/inet/mimetypes.d',
+    'source/vibe/inet/webform.d',
+    'source/vibe/inet/urltransfer.d',
+    'source/vibe/inet/message.d'
+]
+install_headers(vibe_inet_src, subdir: 'd/vibe/inet')
+
+vibe_stream_src = [
+    'source/vibe/stream/botan.d',
+    'source/vibe/stream/counting.d',
+    'source/vibe/stream/taskpipe.d',
+    'source/vibe/stream/base64.d',
+    'source/vibe/stream/zlib.d',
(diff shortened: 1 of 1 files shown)


rikkimax:
There is one big issue you have not addressed.
No D compiler to my knowledge supports multiple versions of itself and along with it other compilers mixing of binaries produced.
How are is that being handled with Meson? If this is not handled there is no point pulling this.

Also reminder bindings are D code. If Meson cannot handle source packages that are D, it won't handle this use case. Dub already supports using the system linker to link against e.g. libevent although neither it or OpenSSL should be required these days.

ximion:
@rikkimax Since LDC is the only completely free compiler that can output shared libraries, this problem doesn't exist (yet) for distros.
I could make the Meson build definition suffix the Vibe libraries with a compiler-specific tag though to work around this issue, if you would like that.
The issue definitely exists for static libraries.

Meson obviously can handle source packages that are in D, but usually needs those to have a Meson build-definition as well. Maybe we could add some dub glue as a Meson module at some point.
In this regard though it's still better than using Automake (or CMake) which don't work well with that scheme at all.

s-ludwig:
Leaving the question of whether pre-compiled binaries make sense for D at the moment aside, I have no objections against adding a Meson build file, as long as anybody uses it and reports issues. Just the main.d file seems slightly out of place. Can Meson use the `-main` switch of DMD (assuming that LDC and GDC even have an equivalent switch)?

Regarding the DUB issues, I have them all on my short TODO list, but it's been a while since I've been able to work on that. It's a little sad that there isn't more active participation for DUB (i.e. PRs), considering the number of users, but eventually I'll get to it, too. In case of https://github.com/dlang/dub/issues/838 it would be great, if you could sketch a proposal of how the directory structure should look like. I got a lot of feedback for that in the past and some of it contradicts each other, and I'm just half familiar with the latest conventions on Linux systems.

s-ludwig:
BTW, a meson generator module for DUB would be a good idea and should be pretty straight forward (see e.g. https://github.com/dlang/dub/blob/master/source/dub/generators/cmake.d for the CMake equivalent).

ximion:
@s-ludwig Nice :) Well, I will use it, definitely. The incompatible ABIs are st time the biggest problem with D adoption in distros, and it can only be solved if the D Foundation standardizes an ABI and all compilers adopt it.
I would definitely use the Meson piece, and help keeping it up to date (but it would still be appreciated if you could take a look at the file-lists in there to see if they are up-to-date. Meson and Ninja require the lists to quickly scan files and have deterministic output).
I also need meson to conditionally generate files at build-time, e.g. to generate D classes from GIR (GObject repository) of GObject C libraries - which is super awesome, because it allows you to bind to a lot of C stuff without needing extra D bindings.

To be precise, I want to try out the Mongo bindings for a MongoDB backend for the appstream-generator, and need the full Vibe.d for a web interface to a package and software bundle management solution, which all should end up in Debian eventually - I hope this works out, D is amazing to work with when it works, but the toolchain at time has some issues which make it quite painful often, and makes it hard to convince others to invest time in it (while they could just use Python or C with rock-solid, well-tested and maintained tools)

Sorry, this comment went slightly OT :P

ximion:
@s-ludwig 
> Can Meson use the `-main` switch of DMD (assuming that LDC and GDC even have an equivalent switch)?

Yes, it can - I adapted the patch.

s-ludwig:
Okay, thanks! I'll try to remember looking at the file list when necessary, but fortunately such changes are relatively rare these days. It would probably also make sense to add a Meson build invocation to the CI script.

ximion:
@s-ludwig Yes, I would like that - unfortunately, the only Meson version with D support is in Debian Stretch and Ubuntu Yakkety right now, and you can only get Ubuntu Trusty on the CI, unless you do tricks with Docker ;-)

s-ludwig:
Is downloading + building it from source an option?

ximion:
@s-ludwig Jup, would be very easy even, since you don't need to build it and would just have to download it (it's Python and allows being executed without installing).

PR #1576, opened 2016-09-27
Fix intermittent infinite loop error in readUntil

etcimon:
In very rare cases, this caused the application (on my fork) to be stuck in an infinite loop in server.d's parse HTTP headers, leastSize was used to make 0 byte reads rather than recognize the value as a connection failure.


etcimon:
hm actually it's `nread` being 0 that causes the problem

PR #1581, opened 2016-10-01
declare SMTPClientSettings parameter of sendMail as 'in'

tosttost:
as it is not modified.

### DIFF
--- a/source/vibe/mail/smtp.d
+++ b/source/vibe/mail/smtp.d
@@ -120,7 +120,7 @@ final class Mail {
 
 	Valid headers can be found at http://tools.ietf.org/html/rfc4021
 */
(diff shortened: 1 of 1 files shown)


tosttost:
Is it possible to get some feedback on this?
My motivation for this:
- store the settings in a global variable and reuse it -> with current API, it has to be mutable. Therefore it needs to be constructed in static this()
- once you want to send a mail on server startup, it runs in shared static ctor and therefore before any static ctor is executed and results in a segfault.

With this param declared const, one can store the settings in a immutable field and initialise it in shared static ctor.

s-ludwig:
Sorry, missed this. Should be fine!

Issue #1580, opened 2016-10-01
validateEmail incorrectly tests results of std.net.isemail.isEmail()


radcapricorn:
vibe.utils.validation.validateEmail tests the results of std.net.isemail.isEmail like this:
```
return isEmail(str) != EmailStatusCode.valid;
```
However, isEmail [returns EmailStatus](http://dlang.org/phobos/std_net_isemail.html#.isEmail), not EmailStatusCode. Incidentally, EmailStatus aliases self to a bool property valid(), which allows the code above to compile. But the logic is reversed, since EmailStatusCode.valid == 0, although isEmail(str) == false when the e-mail is **invalid**.
Paired with Phobos fixes between versions, this leads to further inconsistencies at run time.
For example, with dmd 2.071.1 (when there was [an issue](https://issues.dlang.org/show_bug.cgi?id=16036) in Phobos):
```
auto result = isEmail("a@b.c");
/*
result == EmailStatus
{
	valid: false
	localPart: a
	domainPart: b.c
	statusCode: error
}
*/
auto err = appender!string();
auto vibeResult = validateEmail(err, "a@b.c");
/*
vibeResult == true
*/
```
Even though isEmail() (incorrectly) returned an invalid status, vibe treats it as valid.
And with current dmd nightly:
```
auto result = isEmail("a@b.c");
/*
result == EmailStatus
{
	valid: true
	localPart: a
	domainPart: b.c
	statusCode: valid
}
*/
auto err = appender!string();
auto vibeResult = validateEmail(err, "a@b.c");
/*
vibeResult == false
err.data == "The email address is invalid."
*/
```
Vice-versa, isEmail() now correctly returns a valid status, but vibe treats it as invalid.

radcapricorn:
Note: Phobos issue mentioned above holds for Phobos included with dmd 2.071.2 as well. The proposed fix would effectively break validateEmail() for anyone not using nightly dmd/Phobos builds.

s-ludwig:
Fixed by #1582.

PR #1582, opened 2016-10-01
Fix validateEmail/isEmail inconsistency

radcapricorn:
Proposed fix for [issue 1580](https://github.com/rejectedsoftware/vibe.d/issues/1580).

### DIFF
--- a/source/vibe/utils/validation.d
+++ b/source/vibe/utils/validation.d
@@ -67,7 +67,13 @@ bool validateEmail(R)(ref R error_sink, string str, size_t max_length = 64)
 		return false;
 	}
 
-	if (() @trusted { return !isEmail(str); }()) {
+	if (() @trusted {
+				static if (__VERSION__ >= 2072) {
+					return !isEmail(str);
+				} else {
+					return isEmail(str) != ErrorStatusCode.valid;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Interesting, so do I understand it right that `isEmail` has actually never worked at all? In that case, the affected `if` statement should be wrapped inside `static if (__VERSION__ >= 2072)`, so that the validation function does at least something useful for the released compiler versions.

radcapricorn:
I'm not sure about "never", perhaps it's just not that widely used so it went unnoticed for a long time. I've added the version check.

s-ludwig:
Okay, with "never" I meant something like at least since 2.064 or so, or since vibe.d started to use it.

I've moved the `static if` to outside of the `if`\*, so that the whole `isEmail` check is skipped on older DMD versions (doesn't make sense anyway). The basic checks are still performed, so it isn't that bad. The 2.072.0-b1 build presumably will still fail, but I'll fix that after merging this.

\* Just used the new PR edit feature for the first time. Definitely a lot nicer/more efficient than requesting the PR author to change small bits here and there.

radcapricorn:
Thanks!

PR #1592, opened 2016-10-09
Adds a persist flag in FileDescriptorEvent.

japplegame:
This is the same as @oglu's PR #763, but with less overhead.

### DIFF
--- a/source/vibe/core/core.d
+++ b/source/vibe/core/core.d
@@ -826,10 +826,10 @@ Timer createTimer(void delegate() callback)
 		Returns a newly created FileDescriptorEvent associated with the given
 		file descriptor.
 */
-FileDescriptorEvent createFileDescriptorEvent(int file_descriptor, FileDescriptorEvent.Trigger event_mask)
+FileDescriptorEvent createFileDescriptorEvent(int file_descriptor, FileDescriptorEvent.Trigger event_mask, bool persist = true)
 {
 	auto drv = getEventDriver();
-	return drv.createFileDescriptorEvent(file_descriptor, event_mask);
+	return drv.createFileDescriptorEvent(file_descriptor, event_mask, persist);
 }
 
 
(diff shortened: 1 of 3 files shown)


s-ludwig:
Did you measure if the overhead is actually lower? I'd expect both to be approximately the same, but #763 to be slightly faster on some systems because disabling the event is done automatically by the low-level API used.

So what I'm mainly wondering in general is whether we actually need a switch here, or just a sane default behavior. If it turns out a switch is needed, it should take the form of an enum instead of a Boolean.

What I'm leaning towards for efficiency is to make edge-triggered the default for the long term. Edge-triggered works fine for the usual approach of "try to perform op -> if returns EWOULDBLOCK or similar, wait -> try again". Non-persistent is a little more forgiving when someone departs from that route, but it might also hide issues in that case.

In general, I'd like to keep API extensions to a minimum, because they will have to be supported in the long term (also for example in the new vibe-core module and the underlying eventcore abstraction). I could even imagine that a hard switch to ET could work fine (coupled with the possibility to set a `VibeOldFileDestriptorWait` version to give projects more time to adjust), but in that case I'd like to be sure that at least no open source project depends on the old behavior to be reasonably sure not to break too much code.

s-ludwig:
> in that case I'd like to be sure that at least no open source project depends on the old behavior to be reasonably sure not to break too much code.

Okay, scratch that. Six out of six open source projects that use `createFileDescriptorEvent` use "wait -> try to read (-> repeat if still have to wait)" instead of the above protocol, so this will obviously break more code than I had anticipated. So I'd propose to add an `enum EventWaitMode { persistent, edgeTriggered }`, with the long term goal to switch to edge triggered as the default after going through a proper deprecation path. That is, if nothing speaks against edge-triggered in your case, of course!

japplegame:
Еhanks for the detailed answer. I'll try to write new PR.

japplegame:
New PR #1596

PR #1593, opened 2016-10-09
Test Meson build on the CI

ximion:
Hi!
Follow-up to https://github.com/rejectedsoftware/vibe.d/pull/1574: I quickly made the CI test Meson builds as well.
The problem with this is that we can't continue to do rootless builds, since Travis' environment is super ancient and we need a Ninja version > 1.6. There is an open feature request at Travis for this, since Ninja is required to build Chromium as well, but nothing happened there as well.

This change in the environment apparently also breaks the tests, I am working on fixing this, but since I don't know the testsuite well it's a bit difficult.

Also, for some reason DMD fails with `Error: unrecognized file extension 30` on some files, I have no idea why yet.

So, this PR is not really for merging yet but rather a "how to probably do it" :-)
That DMD thing looks like a DMD bug or an issue in the Meson settings for DMD, which needs investigating.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: d
+dist: trusty
+sudo: true
 
 d:
   # order: latest DMD, oldest DMD, LDC/GDC, remaining DMD versions
@@ -28,10 +30,13 @@ matrix:
     - d: ldc-1.0.0
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
+before_script:
+ - sudo add-apt-repository -y ppa:ximion/tests
+ - sudo apt-get update
+ - sudo apt-get install -y ninja-build libevent-dev
+
 script: ./travis-ci.sh
 
 services:
   - mongodb
   - redis-server
-
-sudo: false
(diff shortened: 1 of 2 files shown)


s-ludwig:
Hi! The test failure at the bottom looks strange (failing to listen on port 11388), but maybe something is running on that port in the root environment? The port is set in three places in the [corresponding test](https://github.com/rejectedsoftware/vibe.d/blob/master/tests/vibe.http.server.1388/source/app.d#L12).

Generally, losing `sudo: false` would be unfortunate, because the build times and build throughput is already borderline at times - but of course it needs a passing build to see if the difference is actually that big in practice.

The unrecognized file extension error appears to be caused by passing "libvibe-utils.so.0.7.30" to DMD in the Meson based build. Apparently DMD can't cope with the version suffix and expects ".so" to come last. Not sure how to fix this aside from creating a symlink with an .so extension or fixing DMD.

ximion:
Superseded by https://github.com/rejectedsoftware/vibe.d/pull/1720, which would have to be merged into `master` as well (it only targets the stable branch at time)

PR #1599, opened 2016-10-28
seralizer: Fix wrong order of unittest/docu

Marenz:
The documentation for `isStringSerializable` / `isPolicySerializable` is generated from the unittests.
The unittests are however in the wrong order, so that the docu for `isStringSerializable` is found where 
one would expect to see `isPolicySerializable` and vice versa.

See an example of the wrong docu at (see example code there):

http://vibed.org/api/vibe.data.serialization/isStringSerializable
http://vibed.org/api/vibe.data.serialization/isPolicySerializable

### DIFF
--- a/source/vibe/data/serialization.d
+++ b/source/vibe/data/serialization.d
@@ -901,22 +901,16 @@ template isStringSerializable(T)
 unittest {
 	import std.conv;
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks! Wonder how *that* happened...

PR #1603, opened 2016-10-30
sort allow methods before comparing

MartinNowak:
- fixes tests with modified sorting function (https://github.com/dlang/phobos/pull/4816)


MartinNowak:
This fixes testing w/ 2.072.0 @s-ludwig

s-ludwig:
Thanks! I was just about to look into this. The forward-reference issue has already been fixed/worked around in memutils, so that everything should work fine now with 2.072.0.

MartinNowak:
We should push back on such breakages and fix dmd instead. The project tester on https://ci.dawg.eu tests the latest stable release version of projects in order to better find regressions in the compiler.

s-ludwig:
True of course, I just wanted to make sure that the new vibe.d release can be tagged safely for now.

PR #1609, opened 2016-11-02
Split up sub package into separate folders

s-ludwig:
This corrects import dependencies and prepares for being able to replace individual sub packages by external packages (e.g. vibe-core).


s-ludwig:
Moving forward with this to enable the next steps for the vibe-core transition. This will probably break all existing pull requests, but most are trivial to fix (I'll do that if necessary) or require larger changes anyway.

PR #1618, opened 2016-11-10
Make the code base compatible with vibe-core

s-ludwig:
This PR adds the possibility to use vibe-core as a drop-in replacement for vibe:core.


s-ludwig:
Integrating into mainline to get better real world coverage. Using vibe-core is opt-in using the "vibe-core" configuration and the rest are more or less trivial API/annotation changes.

PR #1622, opened 2016-11-16
1615 json exception messages

### LINKED ISSUE #1615: improve the error message in JSONException
In `/vibe/data/json.d` is the following code:

```D
	//
	// deserialization
	//
	void readDictionary(Traits)(scope void delegate(string) field_handler)
	{
		enforceJson(m_current.type == Json.Type.object, "Expected JSON object, got "~m_current.type.to!string);
		auto old = m_current;
		foreach (string key, value; m_current) {
			m_current = value;
			field_handler(key);
		}
		m_current = old;
	}
```

which puts out the message *Expected JSON object, got array* when m_current.type isn't an object.

That's fair enough and would be fine if the data I was deserializing only had a few fields. Unfortunately for me I have about 100 fields that I now have to check to see which one caused the problem. Please can you get the error message to also display the name of the field which caused the problem.

gedaiu:
fixed #1615

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1551,29 +1551,6 @@ unittest { // const and mutable json
 	assert(serializeToJson(k) == Json(2));
 }
 
-unittest {
-	import std.stdio;
-
-	bool hasException;
-
-	struct Child {
-		string b;
-	}
-	struct Item {
-		Child c;
-	}
-
-	try {
-		Item a = deserializeJson!Item("{ \"c\": 3}");
-	} catch(Exception e) {
-		assert(e.msg == "Expecting `c` to be object.");
-		hasException = true;
-	}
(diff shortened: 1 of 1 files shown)


SingingBush:
@gedaiu thanks

s-ludwig:
Can you wrap the additions within `version (VibeJsonFieldNames)`? The default code path should stay as efficient as possible and this adds quite some memory allocations, especially in the `JsonStringSeriailzer`.

gedaiu:
I will take a look these days...

gedaiu:
@s-ludwig ready for another look

WebFreak001:
I think it would be a better idea to add a `string[] path` in JSONException and modify the toString method which takes a sink which can also avoid some GC allocations

PR #1632, opened 2016-11-25
Implemented SCRAM-SHA1 authentication method for mongo

ngurrola:
The default is still MONGO-CR, even for server versions 3.0+ to maintain backwards compatibility. It can be switched by adding "authMechanism=SCRAM-SHA-1" to the connection query string.

### DIFF
--- a/mongodb/vibe/db/mongo/sasl.d
+++ b/mongodb/vibe/db/mongo/sasl.d
@@ -1,0 +1,153 @@
+/**
+	SASL authentication functions
+
+	Copyright: © 2012-2016 Nicolas Gurrola
+	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
+	Authors: Nicolas Gurrola
+*/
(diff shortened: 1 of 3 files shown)


mongodb/vibe/db/mongo/sasl.d
```diff
@@ -0,0 +1,153 @@
...
+import std.digest.sha;
+import std.exception;
+import std.format;
+import std.string;
+import std.utf;
+import vibe.crypto.cryptorand;
+
+private SHA1HashMixerRNG g_rng;
+
+struct ScramState
```
s-ludwig: I'd mark this as `package`, so that we don't have to worry about the API so much for now.

s-ludwig:
Really great to finally have this! I never had the time (and need) to implement it, even though it's a rather important feature.

ngurrola:
I added those changes as requested.

s-ludwig:
Looks good to merge. Thanks a lot!

PR #1634, opened 2016-11-26
fix #716 (Add partial content with range)

### LINKED ISSUE #716: Support range based HTTP queries for the HTTP file server
To support resuming of downloads and fix certain issues on some browsers, the `Accept-Ranges` and related headers should be supported by the HTTP file server module.

See also http://forum.dlang.org/post/vrshdjypnvxkurufqmut@forum.dlang.org

WebFreak001:
Mostly based on https://github.com/rejectedsoftware/vibe.d/issues/716#issuecomment-220606604

Changes to that comment: added `-[X]` format and sending partialContent status code. Also some cleaner code and checks against bad requests.

Not sure how to add an automatic test but I did manually test it locally using this very simple file server in chrome and firefox:

```d
import vibe.http.server;
import vibe.http.fileserver;

shared static this()
{
	settings.port = 10716;
	listenHTTP(settings, serveStaticFiles("data"));
}
```

My aim was only fixing media streams (audio & video + seeking), but pausing downloads and resuming also works now.

### DIFF
--- a/http/vibe/http/fileserver.d
+++ b/http/vibe/http/fileserver.d
@@ -13,7 +13,6 @@ import vibe.http.server;
 import vibe.inet.message;
 import vibe.inet.mimetypes;
 import vibe.inet.url;
-import vibe.stream.counting;
 
 import std.conv;
 import std.datetime;
@@ -403,7 +402,7 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse 
 
 	if (prange) {
 		fil.seek(rangeStart);
-		res.bodyWriter.write(new LimitedInputStream(fil, rangeEnd - rangeStart + 1));
+		res.bodyWriter.write(fil, rangeEnd - rangeStart + 1);
 		logTrace("partially sent file %d-%d, %s!", rangeStart, rangeEnd, res.headers["Content-Type"]);
 	} else {
 		if (pce && !encodedFilepath.length)
--- a/http/vibe/http/fileserver.d
+++ b/http/vibe/http/fileserver.d
@@ -18,6 +18,7 @@ import std.conv;
 import std.datetime;
 import std.digest.md;
 import std.string;
+import std.algorithm;
 
 
 /**
@@ -318,6 +319,8 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse 
 
 	if (prange) {
 		auto range = (*prange).chompPrefix("bytes=");
+		if (range.canFind(','))
+			throw new HTTPStatusException(HTTPStatus.notImplemented);
 		auto s = range.split("-");
 		// https://tools.ietf.org/html/rfc7233
 		// Range can be in form "-\d", "\d-" or "\d-\d"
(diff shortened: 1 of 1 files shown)


WebFreak001:
Using this on my websites now and everything seems to work as expected and users don't complain about video playback anymore.

http/vibe/http/fileserver.d
```diff
@@ -311,7 +312,41 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse
...
+	res.headers.addField("Accept-Ranges", "bytes");
+	ulong rangeStart = 0;
+	ulong rangeEnd = 0;
+	auto prange = "Range" in req.headers;
+
+	if (prange) {
+		auto range = (*prange).chompPrefix("bytes=");
+		auto s = range.split("-");
+		// https://tools.ietf.org/html/rfc7233
+		// Range can be in form "-\d", "\d-" or "\d-\d"
```
s-ludwig: The range can also have "/\d+" or "/\*" as an additional suffix, and multiple ranges can be separated by comma. We should probably check that `s.length == 2` and that `s[1].indexOfAny("/*") < 0`. Actually supporting those two features seems less important, though.

s-ludwig: "\" - not sure if that's just my browser, but that's supposed to be a backslash.

s-ludwig:
Instead of `res.bodyWriter.write(new LimitedInputStream(fil, rangeEnd - rangeStart + 1));`, `write` accepts a length parameter: `res.bodyWriter.write(fil, rangeEnd - rangeStart + 1);`. The `writeRawBody` optimization (enables the use of `sendfile`) could also be used for the range, but I'm afraid that this will uncover a bug in the current implementation (I think it doesn't check `File.tell` before calling `sendfile`).

The functionality itself looks good AFAICS.

http/vibe/http/fileserver.d
s-ludwig: It needs to be sent using a multi-part response. But I'd just send a 400 Bad Request back instead, as I don't think that any typical client will ever request multiple ranges.

s-ludwig: ...or rather 501 Not Implemented

s-ludwig:
Okay, looks good to merge. Thanks for tackling this!

I'll look into adding a test after the merge.

http/vibe/http/fileserver.d
```diff
@@ -311,7 +312,43 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse
...
+		auto range = (*prange).chompPrefix("bytes=");
+		if (range.canFind(','))
+			throw new HTTPStatusException(HTTPStatus.notImplemented);
+		auto s = range.split("-");
+		// https://tools.ietf.org/html/rfc7233
+		// Range can be in form "-\d", "\d-" or "\d-\d"
+		if (s[0].length) {
+			rangeStart = s[0].to!ulong;
+			rangeEnd = s[1].length ? s[1].to!ulong : dirent.size;
+		} else if (s[1].length) {
```
yazd: You need to check `length` of `s` before you access `s[1]`.
The client can send malformed headers.

WebFreak001: Uh well this PR is merged already and I recreated the repo so I can't edit this anymore, I will make a separate PR.

PR #1635, opened 2016-11-27
Fix #1628

### LINKED ISSUE #1628: URLRouter not escaping params properly
```d
router.get("/tag/:tag", &myFunc);
```

Will match `/tag/foo` and `/tag/A%2BB` (A.B) but it will not match `/tag/A%2FB` (A/B).

I see why it wouldn't match `/foo/A/B` but why doesn't it match when I escape the `/`?

WebFreak001:
Fixes requests on a urlmatcher with `/` inside the parameters

### DIFF
--- a/http/vibe/http/router.d
+++ b/http/vibe/http/router.d
@@ -12,6 +12,8 @@ module vibe.http.router;
 public import vibe.http.server;
 
 import vibe.core.log;
+import vibe.inet.url;
+import vibe.textfilter.urlencode;
 
 import std.functional;
 
@@ -253,7 +255,7 @@ final class URLRouter : HTTPServerRequestHandler {
 			return p.relativeToWeb(Path(req.path)).toString();
 		}
 
-		auto path = req.path;
+		auto path = URL.parse(req.requestURL).pathString; // get the encoded raw url string to allow escaped / inside variables
 		if (path.length < m_prefix.length || path[0 .. m_prefix.length] != m_prefix) return;
 		path = path[m_prefix.length .. $];
 
@@ -263,7 +265,7 @@ final class URLRouter : HTTPServerRequestHandler {
 				if (r.method != method) return false;
 
 				logDebugV("route match: %s -> %s %s %s", req.path, r.method, r.pattern, values);
-				foreach (i, v; values) req.params[m_routes.getTerminalVarNames(ridx)[i]] = v;
+				foreach (i, v; values) req.params[m_routes.getTerminalVarNames(ridx)[i]] = urlDecode(v); // decode because using raw url
 				if (m_computeBasePath) req.params["routerRootDir"] = calcBasePath();
 				r.cb(req, res);
 				return res.headerWritten;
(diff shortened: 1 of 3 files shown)


s-ludwig:
This is the right approach, the implementation just currently has the drawback of adding GC allocations in the core system. Probably a proper solution will require a redesign of how `req.params` are handled, and thus is a target for the revamped HTTP package.

But here I would like to at least avoid making the common case (unencoded ASCII characters) worse in this regard. I'll improve `urlDecode` to return a slice of the input in case of no encoding sequences and then pull this.

s-ludwig:
Thinking about this some more, doesn't this break the specific case that I mentioned in the other PR? So `router.get("/foo bar/")` wouldn't match "/foo%20bar/".

It seems like the only clean solution is to compare each path segment individually, of both, the match string and the incoming URI. That would need to happen either in decoded form, or in a normalized encoded form (i.e. always  upper case hex numbers and having a strict set of characters that is encoded). The latter would be the better approach, as the router could then continue to operate on a character basis.

To make this efficient, match strings can be converted to the normalized form once at startup and the incoming URI can be transformed using a special `@nogc` forward range. It will just presumably be a little difficult to still keep the slicing behavior for the `params` array, but I can look into that.

What do you think? 

----
This is how I imagine things should work:
```D
bool matches(string pattern, string uri) { ... }
assert(matches("/foo bar/", "/foo%20bar/"));   // normalized pattern: "/foo%20bar/"
assert(matches("/foo%20bar/", "/foo%20bar/")); // normalized pattern: "/foo%20bar/"
assert(matches("/foo/bar/", "/foo/bar/"));     // normalized pattern: "/foo/bar/"
assert(!matches("/foo/bar/", "/foo%2fbar/"));
assert(matches("/foo%2fbar/", "/foo%2fbar/")); // normalized pattern: "/foo%2Fbar/"
assert(matches("/foo%2Fbar/", "/foo%2fbar/")); // normalized pattern: "/foo%2Fbar/"
assert(matches("/foo%2fbar/", "/foo%2Fbar/"));
assert(!matches("/foo%2fbar/", "/foo/bar/"));
assert(matches("/:foo/", "/foo%2Fbar/"));
assert(!matches("/:foo/", "/foo/bar/"));
```

WebFreak001:
Hm about the `/foo bar/` thing I would say that isn't a valid router pattern because it isn't a valid URL segment. One could avoid this by escaping all invalid characters except `/` like chrome does when the user inputs some URL.

That matches function would be a good, clean and easy to implement solution but you would need to make sure that it is used everywhere before any escapes and used properly, which might be pretty hard to accomplish. It also needs to reference the req.params variable so it can fill the variables, or at least that would be an efficient way of handling things.

I could make said `matches` function but I think implementing it into the existing code will be a bit difficult

s-ludwig:
> Hm about the /foo bar/ thing I would say that isn't a valid router pattern because it isn't a valid URL segment. One could avoid this by escaping all invalid characters except / like chrome does when the user inputs some URL.

That can be argued, but it would still leave the possibility for different case of the hex digits, so that path normalization is needed anyway. But more importantly, this was supported all the time, so there is no chance we can simply change/break it. At the very least, a proper deprecation path would be necessary.

About the `matches` function, it was really just meant as some kind of wrapper that tests what the actual router does. I'll add something like that as a unit test with the currently passing cases to go forward from there.

s-ludwig:
Added on master: 9a820c11399131953a1c21acc55dc9f7cda7a886

PR #1642, opened 2016-12-16
Respect explicitly set Content-Type in sendFile

rexso:
Currently when using **sendFile**, the *Content-Type* header is automatically set based on file extension. While this default behaviour is fine, the problem is that the header is overwritten if already explicitly set.

In my case, files are not stored using any extension, so all files are treated as *application/octet-stream*. I have the correct media type stored elsewhere, and it would be nice if I could manually specify this, without having the header overwritten internally.


s-ludwig:
Thanks, definitely makes sense! In the current version, you could also use the `preWriteCallback` to explicitly override the content type header.

PR #1645, opened 2016-12-26
Propagate XHR error to on_error in JS RestInterface

Timoses:
Allow using the on_error function also when the XHR request fails completely (e.g. when the host is not available).
This is useful when needing feedback on whether or not the XHR request was successful or not.

If there's anything I can improve, please let me know!

PR #1646, opened 2016-12-26
Fix and further error handling capability

Timoses:
Sorry again for previous mishap (https://github.com/rejectedsoftware/vibe.d/pull/1645).

This one also allows to always receive an error when the server responds with an error (in case could not update, etc...).

This time it's tested..

### DIFF
--- a/web/vibe/web/internal/rest/jsclient.d
+++ b/web/vibe/web/internal/rest/jsclient.d
@@ -139,6 +139,9 @@ class JSRestClientSettings
 			fout.put("};\n");
 		}
 
+		// error handling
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks, that was easy to miss. I should really start to use the JS generator in practice and ideally there should be an automated test, too., but I need to look into which headless browser engine is best to use for Travis CI.

s-ludwig:
The tests that currently check for the existence of "onload =" need to be adjusted (should probably check for existence of "on_result(" instead.

Timoses:
> The tests that currently check for the existence of "onload =" need to be adjusted (should probably check for existence of "on_result(" instead.

Oh yes, I didn't think of checking.

PR #1649, opened 2016-12-31
Fix #1648 - ignore authorization in RestInterfaceClient

### LINKED ISSUE #1648: RestInterfaceClient with @requiresAuth service
When trying new vibe.web.auth facility it works ok on the server side, but I've a problem with the client side where it demands:

```D
../../../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/web/auth.d(238,6): Error: static assert  "IFooService must have an authenticate(...) method that takes HTTPServerRequest/HTTPServerResponse parameters and returns an authentication information object."
../../../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/web/auth.d(252,51):        instantiated from here
```

How is this supposed to work? I tried to add the authenticate method to the interface itself, but then there is a problem with registering the REST service because it tries to register route for it.
Also tried the @noRoute attribute, but it does not work with REST.

I'm currently setting the authentication header through REST clients `requestFilter` but new auth is somehow stepping in the way.

Is it possible for RestInterfaceClient to somehow ignore the @requiresAuth on the interface?

Thx

tchaloupka:
Not sure if this is the acceptable way to handle REST API with @requiresAuth on the client side, but it seems to solve the problem for me - server side still requires authorization and rest client just ignores this attribute.

### DIFF
--- a/web/vibe/web/internal/rest/common.d
+++ b/web/vibe/web/internal/rest/common.d
@@ -250,7 +250,9 @@ import std.meta : anySatisfy, Filter;
 
 		StaticRoute[routeCount] ret;
 
-		alias AUTHTP = AuthInfo!TImpl;
+		static if (is(TImpl == class)) {
+			alias AUTHTP = AuthInfo!TImpl;
+		}
 
 		foreach (fi, func; RouteFunctions) {
 			StaticRoute route;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks, I missed the case where the authorization attributes are placed in the interface instead of in the class. But for documentation purposes it would actually be quite nice to place them there. In the long run it would also be nice to extend `@requiresAuth` to also work in a meaningful way for the client, but for now just allowing this case seems like a good idea.

s-ludwig:
I commited this now with the mentioned change regarding the if statement and with an additional test.

PR #1659, opened 2017-01-19
Added a missing toRedis() -conversion for RedisHash.setIfNotExist

Soletek:
RedisHash!T did not compile when type T used a custom string serialization.


s-ludwig:
Thanks!

PR #1662, opened 2017-01-23
Support for named tuple de/serialization as associative array

DBankov:
This is an initial proposal for named tuple de/serialization support as associative arrays. Also part of this change is the deserialization support for the existing tuple serialization as standard arrays.


DBankov:
After looking at the contribution recommendations it seems that it may be better to split this PR in two. One with the Tuple deserialization support (from array) and another with the named tuple to/from associative array support.

Please let me know if the above would be better from reviewers' point of view.

s-ludwig:
Looks like this conflicted with the just merged tuple serialization PR. The change set otherwise looks ready to pull.

DBankov:
asapI'm sorry for the delay. I'll rebase ASAP.

s-ludwig:
Thanks! Merging.

DBankov:
Thanks a lot for merging this PR! :)

PR #1663, opened 2017-01-30
Support deserialization of AA whose key type is string-based enum

### LINKED ISSUE #1660: AA cannot be deserialized if its key type is string-based enum
The following code throws ConvException at runtime but it should not.
I confirmed this issue on Linux 64bit systems with dmd master.

```d
import vibe.data.json;

void main()
{
    enum Foo: string
    {
        Bar = "bar",
        Buzz = "buzz"
    }

    struct S {
        int[Foo] f;
    }

    auto json = Json.emptyObject;
    json["f"] = [Foo.Bar: Json(2000)];
    auto s = json.deserialize!(JsonSerializer, S);
    assert(s.f == [Foo.Bar: 2000]);
}
```

Expected: This code can be compiled and run successfully.

Actual: This code can be compiled successfully but it throws the following exception at runtime:
```
std.conv.ConvException@/home/tom-tan/.denv/versions/dmd-2.072.1/linux/bin64/../../src/phobos/std/conv.d(2475): Foo does not have a member named 'bar'
```

tom-tan:
It fixes #1660.
I also added a unittest for serializing/deserializing AA whose key type is string-based enum.

I leave the case of AA with `@byName` because AA does not support this attribute yet (in implementation).

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1557,6 +1557,24 @@ unittest { // const and mutable json
 	assert(serializeToJson(k) == Json(2));
 }
 
+unittest { // issue #1660 - deserialize AA whose key type is string-based enum
+	enum Foo: string
+	{
+		Bar = "bar",
+		Buzz = "buzz"
+	}
+
+	struct S {
+		int[Foo] f;
+	}
+
+	const s = S([Foo.Bar: 2000]);
+	assert(serializeToJson(s)["f"] == Json([Foo.Bar: Json(2000)]));
+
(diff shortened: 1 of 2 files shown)


tom-tan:
rebased.

s-ludwig:
Thanks, looks good to merge!

I wonder if this should be extended to types that cast implicitly to `string` and can be converted back using `T(str)`. For enums it also may be a good idea to provide some kind of annotation to explicitly request a certain behavior. But for now the existing change looks like a step in the right direction.

PR #1670, opened 2017-02-03
vibe.http.session: Allow to remove a field from a session

wilzbach:
Adds the capability to remove a field from a session.
It's named `remove` after the D's [HashMap `remove`](https://dlang.org/spec/hash-map.html#removing_keys)


s-ludwig:
Thanks, makes sense. I guess this is also small enough to still get into 0.8.0.

PR #1671, opened 2017-02-04
add safe to dynamic array deserializing delegate

tchaloupka:
Just a little fix I came about switching to 8.0.0


s-ludwig:
Is there an easy way to produce a compile error *without* the fix? It should usually infer the `@safe`ty and it would be interesting to know at which point that breaks down.

PR #1674, opened 2017-02-05
Fix #1673 - safety inference in tree structure of REST API

### LINKED ISSUE #1673: vibe-0.8.0-beta.2 - tree structure in REST API
If I have a tree structure defined in REST API newly annotated with @safe it fails to register such an API.

Try this:

```D
@safe unittest {
	static struct Foo { Foo[] foos; }
	serialize!TestSerializer(Foo());
}
```
It ends up with
```
data/vibe/data/serialization.d(1588,26): Error: @safe function 'vibe.data.serialization.__unittestL1586_178' cannot call @system function 'vibe.data.serialization.serialize!(TestSerializer, Foo).serialize'
```

The problem seems to be with safety inference here: https://github.com/rejectedsoftware/vibe.d/blob/master/data/vibe/data/serialization.d#L340

Not sure what to do about it.

tchaloupka:
I solved it by annotating serializeValue method and fixed all unittests so it compiles (with dmd-2.073.0).

There are some more pragma deprecation messages that can be fixed, but I wanted this to be minimal and to see if this is even acceptable.

### DIFF
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -349,7 +349,7 @@ private template serializeValueImpl(Serializer, alias Policy) {
 	static assert(Serializer.isSupportedValueType!string, "All serializers must support string values.");
 	static assert(Serializer.isSupportedValueType!(typeof(null)), "All serializers must support null values.");
 
-	private void serializeValue(T, ATTRIBUTES...)(ref Serializer ser, T value)
+	private void serializeValue(T, ATTRIBUTES...)(ref Serializer ser, T value) @safe
 	{
 		import std.typecons : BitFlags, Nullable, Tuple, Typedef, TypedefType, tuple;
 
@@ -1679,4 +1679,9 @@ unittest {
 	assert(ser == "V(i)(42)", ser);
 	auto deser = deserialize!(TestSerializer, T)(ser);
 	assert(deser == 42);
+}
+
+@safe unittest {
+	static struct Foo { Foo[] foos; }
+	serialize!TestSerializer(Foo());
 }
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -221,7 +221,7 @@ unittest {
 		import std.conv;
 		import std.array;
 
-		static string toRepresentation(T value) {
+		static string toRepresentation(T value) @safe {
 			return to!string(value.x) ~ "x" ~ to!string(value.y);
 		}
 


s-ludwig:
I'm kind of torn on this one and #1671, because on one hand the reasons for both are very valid (in case of the error messages IMO DMD really should generally output the most specific place where `@safe`/`@nogc`/`nothrow` inference failed). On the other hand, forcefully adding `@safe`/`@trusted` breaks inference completely, so now either all user defined serialization methods need to be annotated safe, or the library opens a hidden `@trusted` safety hole.

I will have a look if I can find a higher level spot where the REST code can be fixed.

tchaloupka:
Problem here is worse in that even the struct is safe the inference doesn't work because of cycle of the tree structure which is infered as system as specified in https://dlang.org/spec/function.html#function-attribute-inference

s-ludwig:
Interesting, good to know. Do you possibly know if the decision to infer recursive functions as `@system` was discussed somewhere? It seems a bit arbitrary at first sight.

tchaloupka:
I don't remember nor found any discussion about this. Just somewhat related bug report: https://issues.dlang.org/show_bug.cgi?id=16528

tchaloupka:
Any news about this?
Currently (0.8.0-beta.3) it is impossible to make REST API with tree structure

```D
struct Foo
{
    Foo[] foos;
}
```

tchaloupka:
I've rebased this so now it's evens simpler.
This is stays to be a problem when tree structures are used in REST API.

tchaloupka:
I've resolved the conflict, retested it and somehow @s-ludwig probably fixed this by other change because now it compiles fine without any safety attribute change. (tried with dmd-2.074.0 and dmd-2.073.2).

So this was reduced only to the unittest for the tree structure.. :)

tchaloupka:
Rebased again. Thought that it now works without safe addition, but forgot to unittest with `--combined` switch..

data/vibe/data/serialization.d
```diff
@@ -1474,14 +1474,14 @@ unittest // Testing corner case: Variadic template constructors and methods
 @safe unittest // Immutable data deserialization
 {
 	import vibe.data.json;
-
```
wilzbach: Unrelated changes due to trailing whitespace - this is the past!
-> You will like https://github.com/rejectedsoftware/vibe.d/pull/1800 :)

s-ludwig:
There is one case that I would like to avoid, namely that passing a `@system`  output range for serialization fails to compile. I'll pull this to master now and will then see if it's possible to use a string mixin to fake the job of the compiler here. If that works out, I'll squeeze this into 0.8.0, because it is indeed a regression.

PR #1675, opened 2017-02-06
Add support for retrieving the websocket close code/reason

b1naryth1ef:
This PR adds support for retrieving a websocket connections close code and reason. Fairly straightforward, but vibe.d's source is all new territory so please advise if something is glaringly wrong or horrible.

The TL;DR implementation;

1. If we receive a close frame with at least enough bytes for a `short`, we read the short in as the close code
2. If the close frame contains more than enough bytes for a `short`, we read the short in as the close code and read the remaining bytes in as a character array for the close reason.

### DIFF
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -386,8 +386,9 @@ final class WebSocket {
 		Returns the close code sent by the remote end.
 
 		Note if the connection was never opened, is still alive, or was closed
-		locally this value will be 0. If the connection was not closed cleanly by
-		the remote end, this value will be 1006.
+		locally this value will be 0. If no close code was given by the remote
+		end in the close frame, the value will be 1005. If the connection was
+		not closed cleanly by the remote end, this value will be 1006.
 	*/
 	@property short closeCode() { return m_closeCode; }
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks, looks good. The only thing that I noticed is that the WebSocket spec defines some specific values if no explicit close status was sent, so after close, the status should always be `!= 0`:

> If this Close control frame contains no status code, _The WebSocket
> Connection Close Code_ is considered to be 1005.  If _The WebSocket
> Connection is Closed_ and no Close control frame was received by the
> endpoint (such as could occur if the underlying transport connection
> is lost), _The WebSocket Connection Close Code_ is considered to be 1006.

(https://tools.ietf.org/html/rfc6455#section-7.1.5)

b1naryth1ef:
@s-ludwig should match the RFC better now.

s-ludwig:
Not sure about the reason for the Travis-CI failures, this happens since a while, but it's not an issue in vibe.d itself, at least not directly. I'll just restart the test and will wait until it finishes before merging.

s-ludwig:
Okay, failures have been fixed externally.

Issue #1422, opened 2016-02-12
REST: Serializing the entire body into one parameter


wilzbach:
Hey I am just getting started with vibe.d, however I am struggling with this issue.

Let's say I have a more complex `FooObject` that my API expects. Is there an easy way to teach the REST API to accept a request body that is the entire object like `{"a": "foo", "b": "bar"}` for the example below?
In other words I would like to serialize the body into one attribute.

I know that [best practices](http://jsonapi.org/format/#crud), but this is possible in most REST frameworks I have worked with and the still the default setting for most JS REST libs.

```
struct FooObject{
    string  a;
    string  b;
}
interface Example4API
{
    int createFoo(FooObject whole_body);
}
class Example4 : Example4API
{
    override:
            int createFoo(FooObject whole_body){
                    return 1;
            }

}
```

GayDragon:
Workaround (not my work, obviously):
https://gist.github.com/yannick/ffa037b2338f95f65e4e

wilzbach:
duplicate of https://github.com/rejectedsoftware/vibe.d/issues/1549

PR #1676, opened 2017-02-06
Fix #1549 - Support bodyParam attribute without form parameter name too map the whole body to a single parameter

### LINKED ISSUE #1549: Support bodyParam attribute without form parameter name to map the whole body to a single parameter
Should be supported for the web and REST interfaces:

``` D
struct S { int foo; string bar; }
@bodyParam("s") void postFoo(S s) {}
```

For the web interface generator, this means that the endpoint "POST /foo" accepts two form parameters, "foo" and "bar". The REST interface generator instead expects a JSON object of the form `{"foo": ..., "bar": ...}`.

wilzbach:
As discussed in #1549 and  #1422:

```d
@bodyParam("obj")
string postFoo(FooType obj);
```

Moreover if
- no `@bodyParam` annotation is provided
- the method is automatically recognized as `POST`
- there's only _one_ `@bodyParam` parameter

```d
string postFoo(FooType obj);
```

then the entire Json body is serialized into this parameter. This might be a breaking change, but imho it's a sane default setting as most APIs want to serialize the entire request body and vibe.d hasn't reached stable yet.

### DIFF
--- a/examples/rest/source/app.d
+++ b/examples/rest/source/app.d
@@ -353,8 +353,7 @@ unittest
  * However, it makes no sense to have 'ref' or 'out' parameters on
  * body or query parameter, so those are treated as error at compile time.
  *
- * To serialize the entire Json body into one parameter, simply name the
- * parameter "foo"
+ * If no Json fieldname is passed to @bodyParam, the entire Json body is deserialized into the respective field.
  */
 @rootPathFromName
 interface Example6API
--- a/examples/rest/source/app.d
+++ b/examples/rest/source/app.d
@@ -352,6 +352,9 @@ unittest
  * This is to be consistent with the way D 'out' and 'ref' works.
  * However, it makes no sense to have 'ref' or 'out' parameters on
  * body or query parameter, so those are treated as error at compile time.
+ *
+ * To serialize the entire Json body into one parameter, simply name the
+ * parameter "foo"
  */
 @rootPathFromName
 interface Example6API
@@ -373,6 +376,13 @@ interface Example6API
 	// currently serializing passed data as Json and pass them through the body.
 	@bodyParam("myFoo", "parameter")
 	string postConcat(FooType myFoo);
+
+	// If no field name is passed to @bodyParam the entire json object is
+	// serialized into the parameter.
+	// Moreover if only one bodyParameter is present, this is the default
+	// behavior.
+	@bodyParam("obj")
+	string postConcatBody(FooType obj);
 
 	struct FooType {
 		int a;
@@ -413,6 +423,11 @@ override:
 	{
 		import std.conv : to;
 		return to!string(myFoo.a)~myFoo.s~to!string(myFoo.d);
+	}
+
+	string postConcatBody(FooType obj)
+	{
+		return postConcat(obj);
 	}
 }
 
@@ -558,6 +573,11 @@ shared static this()
 			auto api = new RestInterfaceClient!Example6API("http://127.0.0.1:8080");
 			auto answer = api.postAnswer("IDK");
 			assert(answer == "False");
+
+			Example6API.FooType fType = {a: 1, s: "str", d: 3.14};
+			auto expected = "1str3.14";
+			assert(api.postConcat(fType) == expected);
+			assert(api.postConcatBody(fType) == expected);
 		}
 
 		// Example 7 -- Custom JSON response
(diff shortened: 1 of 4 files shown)


zunkree:
Any news?

web/vibe/web/common.d
```diff
@@ -463,11 +463,14 @@ package struct WebParamAttribute {
 	string field;
 }
 
+package(vibe.web) enum bodyParamWholeName = "bodyParamWhole";
```
s-ludwig: What about using an empty string here and asserting in the two-argument `@bodyParam` overload that the field name is always non-empty? That would eliminate any possibility for name clashes. I'd also tend to encapsulate the "whole body" logic into `WebParamAttribute` (e.g. `WebParamAttribute.isAllInOne` or similar, which performs the name comparison) - thinking about it, it may also make sense to extend this to query string parameters later on.

examples/rest/source/app.d
```diff
@@ -352,6 +352,9 @@ unittest
  * This is to be consistent with the way D 'out' and 'ref' works.
  * However, it makes no sense to have 'ref' or 'out' parameters on
  * body or query parameter, so those are treated as error at compile time.
+ *
+ * To serialize the entire Json body into one parameter, simply name the
+ * parameter "foo"
```
s-ludwig: The "foo" was a surprise here, was that left over from an earlier approach?

s-ludwig:
> there's only one @bodyParam parameter
```string postFoo(FooType obj);```
then the entire Json body is serialized into this parameter. This might be a breaking change, but imho it's a sane default setting as most APIs want to serialize the entire request body and vibe.d hasn't reached stable yet.

This may go a bit too far. The library may not be declared stable, but is already used in production all over the place, so silent breaking changes like these are usually a no-go. Generally, I think that keeping things orthogonal here may be a good idea. For example, in the web frontend the equivalent rule would be very surprising and generally undesirable (form with only a single input field).

**Comment by s-ludwig** [comment]
Apart from the open questions, the implementation looks good and ready to merge!

**Comment by s-ludwig** [comment]
Hm, I can't really adjust to the new review feature. I guess I'll just stick to normal comments.

**Comment by wilzbach** [comment]
>  I'd also tend to encapsulate the "whole body" logic into WebParamAttribute (e.g. WebParamAttribute.isAllInOne or similar, which performs the name comparison) - thinking about it, it may also make sense to extend this to query string parameters later on.

I am not sure how I would be able to access the `WebParamAttribute` object from the `jsonMethodHandler`

> This may go a bit too far. The library may not be declared stable, but is already used in production all over the place, so silent breaking changes like these are usually a no-go. Generally, I think that keeping things orthogonal here may be a good idea. For example, in the web frontend the equivalent rule would be very surprising and generally undesirable (form with only a single input field).

I get your point and I removed the default rule to serialize the entire body into an parameter if only parameter is provided, s.t. this isn't blocking this PR.
However, imho in the context of REST APIs it's _very_ common to accept a custom object - the best example being the [GH API](https://developer.github.com/v3/) itself. Or in Vibe.d "words" a usual API looks like this:

```d
auto postA(FooType1 obj);
auto postB(FooType2 obj);
auto postC(FooType3 obj);
....
```

Maybe we can use a special name like it's currently done for `id`?

s-ludwig:
> I am not sure how I would be able to access the `WebParamAttribute` object from the `jsonMethodHandler`

Right, I didn't think of that. But looking at it, the cleanest way to pass that information seems to be to create a new `ParameterType` for full body parameters. Then `RestInterface!T` would access the `WebParamAttribute` and the logic could stay close together.

> I get your point and I removed the default rule to serialize the entire body into an parameter if only parameter is provided, s.t. this isn't blocking this PR.
However, imho in the context of REST APIs it's very common to accept a custom object - the best example being the GH API itself. Or in Vibe.d "words" a usual API looks like this:
> ```
> auto postA(FooType1 obj);
> auto postB(FooType2 obj);
> auto postC(FooType3 obj);
> ....
> ```
> Maybe we can use a special name like it's currently done for id?

I was thinking about using a default name, too, but the "id" feature is rather questionable today when the `Collection` functionality is taken into account (each "id" needs to have a unique name there, the auth module is in a similar boat). But I can definitely see how the difference that this would make cannot be neglected. I'll have to think about this some more before being able to give a meaningful answer.

s-ludwig:
Suggestions implemented in #1723.

PR #1677, opened 2017-02-08
Lazy initialization of req.json

wilzbach:
I assume that there's probably a reason why the request parameters aren't lazily initialized, but I think the "pay once you use" is quite nice. Hence this PR is an open question whether the request parameters can be initialized on their first use. An additional benefit of using this pattern is that the returned `Json` reference is `const`.
Of course, the same pattern could be used to "lazify" other request parameters as well.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -443,7 +443,7 @@ enum HTTPServerOption {
 	parseQueryString          = 1<<1 | parseURL,
 	/// Fills the `.form` field in the request
 	parseFormBody             = 1<<2,
-	/// Fills the `.json` field in the request
+	/// Fills the `.json` field in the request (Deprecated, lazily parsed)
 	parseJsonBody             = 1<<3,
 	/// Fills the `.files` field of the request for "multipart/mixed" requests
 	parseMultiPartBody        = 1<<4, // todo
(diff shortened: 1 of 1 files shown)


http/vibe/http/server.d
```diff
@@ -765,12 +765,21 @@ final class HTTPServerRequest : HTTPRequest {
...
+			A JSON request must have the Content-Type "application/json" or "application/vnd.api+json".
 		*/
-		Json json;
+		@property ref const(Json) json() {
+			if (_json.isNull) {
+				if (icmp2(contentType, "application/json") == 0 || icmp2(contentType, "application/vnd.api+json") == 0 ) {
+					auto bodyStr = () @trusted { return cast(string) bodyReader.readAll(); } ();
+					if (!bodyStr.empty) _json = parseJson(bodyStr);
+				} else {
+					_json = Json.undefined;
```
wilzbach: This is done to emulate the current behavior when no json content type is present.

s-ludwig:
I'm actually planning to make the whole request parsing as lazy as possible for the pending rewrite of the HTTP module (in the form of a new `vibe-http` package), so this fits right in. Since `.json` should be used in a read-only manner anyway, this change should be no problem. The only thing missing is a `@safe` annotation.

The merge for this will be deferred until the 0.8.0 release is out to avoid last-minute regressions.

s-ludwig:
This just needs a rebase, and it would be good to have the `parseJsonBody' option documented as "deprecated". It's unfortunate that enum values can't actually be deprecated, the cases for this really start to pile up.

wilzbach:
> That means that #1677 can be merged now, too.
> This just needs a rebase, and it would be good to have the `parseJsonBody' option documented as "deprecated".

Cool. Done.

s-ludwig:
Meh... looks like Travis has the same issues as yesterday. I'll take the risk and just pull as is ;)

PR #1678, opened 2017-02-09
Add web authentication example

wilzbach:
Vibe.d's authorization framework is amazing, but it's a quite hard to find examples for it (except for [its documentation](http://vibed.org/api/vibe.web.auth/)).

This is based on the original web service example, but instead of `@before` it uses the `vibe.web.auth` framework for access and role management. I added two dummy pages, s.t. the access roles can be checked.

I deliberately avoided using `SessionVar` as this would somewhat make the injection of the current `authUser` redundant in this example.

### DIFF
--- a/examples/web-auth/dub.json
+++ b/examples/web-auth/dub.json
@@ -1,0 +1,8 @@
+{
+	"name": "web-auth-example",
+	"description": "Simple service using the web framework.",
+	"dependencies": {
+		"vibe-d:web": {"path": "../../"}
+	},
+	"versions": ["VibeDefaultMain"]
+}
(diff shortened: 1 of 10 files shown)


s-ludwig:
This looks really nice. Thanks a lot for putting this together! I also still have to write a section in the [general documentation](http://vibed.org/docs#web). Travis errors are unrelated, so this looks good to merge.

s-ludwig:
BTW, is the `hasAuth` enum supposed to improve compile time, or just to save typing (which is of course a very valid reason)? I'm just asking because I always thought that template results are cached anyway for a certain set of template parameters.

PR #1687, opened 2017-02-19
Add checkBasicAuth function.

### LINKED ISSUE #1449: performBasicAuth: plz separate "Authorization" header read code and other stuff
Two performBasicAuth functions contain code duplication and also without separate header read function it isn't possible to check auth data without different response cases.

(Faced out with it because I am writing server what uses auth info as parameter, but response isn't depends from availability of auth info.)

s-ludwig:
Separates the authorization test from authorization enforcement.

Fixes #1449.

### DIFF
--- a/http/vibe/http/auth/basic_auth.d
+++ b/http/vibe/http/auth/basic_auth.d
@@ -24,28 +24,12 @@ HTTPServerRequestDelegateS performBasicAuth(string realm, PasswordVerifyCallback
 {
 	void handleRequest(scope HTTPServerRequest req, scope HTTPServerResponse res)
 	@safe {
-		auto pauth = "Authorization" in req.headers;
-
-		if( pauth && (*pauth).startsWith("Basic ") ){
-			string user_pw = () @trusted { return cast(string)Base64.decode((*pauth)[6 .. $]); } ();
-
-			auto idx = user_pw.indexOf(":");
-			enforceBadRequest(idx >= 0, "Invalid auth string format!");
-			string user = user_pw[0 .. idx];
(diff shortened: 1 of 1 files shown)


s-ludwig:
@denizzzka: Is `checkBasicAuth` sufficient for your use case, or is more control necessary?

http/vibe/http/auth/basic_auth.d
```diff
@@ -71,28 +55,54 @@ HTTPServerRequestDelegateS performBasicAuth(string realm, bool delegate(string,
...
+		pwcheck = A delegate queried for validating user/password pairs
+
+	Returns: Returns `true` $(I iff) a valid Basic Auth header is present
+		and the credentials were verified successfully by the validation
+		callback.
+
+	Throws: Throws a `HTTPStatusExeption` with `HTTPStatusCode.badRequest`
+		if the "Authorization" header is malformed.
+*/
+bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pwcheck)
```
denizzzka: Can ``req`` argument be a ``const`` (or ``in``) type?

denizzzka:
@s-ludwig checkBasicAuth exactly what is needed. Thanks!

(I am changed my code with checkBasicAuth, but can't run my integration tests because something changed in vibe.d: #1688.)

s-ludwig:
Okay, I think it should be safe to still merge this into 0.8.0.

Regarding const/in, I don't want to shut the door for more lazy evaluation and reference counting, which is why the `HTTPServerRe*` values are always passed in plain form.

PR #1689, opened 2017-02-21
Update .travis.yml

John-Colvin:
ran in to some problems myself with dmd 2.073.1, perhaps it'll show up here also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.073.0
+  - dmd-2.073.1
   - dmd-2.070.2
+  - ldc-1.1.0
   - ldc-1.0.0
-  - ldc-1.1.0-beta6
   - dmd-2.072.2
   - dmd-2.071.2
   - dmd-beta


wilzbach:
@John-Colvin Vibe.d is one of the projects that are run by the Project Tester. This means that on every commit in DMD/Druntime/Phobos all these projects get run. However this setup is still in testing mode.

For more details:
https://ci.dawg.eu/job/projects/

John-Colvin:
@wilzbach, that's good to know., 

Because Travis is so popular it seems important that vibe.d is tested in that environment. Also useful to test ldc. Also useful to check changes to vibe.d don't break compatibility with older compilers.

s-ludwig:
I updated to 2.073.2 now on master and 0.7.0 and the tests all succeed. BTW, do the 2.073.1 issues still show up on 2.073.2? I think I had some issues, too, but I can't remember what those were and 2.073.2 appears to work flawlessly.

wilzbach:
This seems to have been superseded by another PR. At least both, new compilers are in:
https://github.com/rejectedsoftware/vibe.d/blob/master/.travis.yml

@John-Colvin @s-ludwig can we close this?
Though it might be a good idea to use the `dmd`, `ldc` tags to avoid needing to continuously update the travis file, see e.g. https://github.com/rejectedsoftware/vibe.d/pull/1772

John-Colvin:
Replaced with: https://github.com/rejectedsoftware/vibe.d/pull/1814

PR #1696, opened 2017-02-27
vibe.web.web: add convenience methods for status and header

wilzbach:
These methods are pretty trivial, but convenient in practice.

### DIFF
--- a/web/vibe/web/web.d
+++ b/web/vibe/web/web.d
@@ -379,6 +379,28 @@ void redirect(string url)
 	ctx.res.redirect(fullurl);
 }
 
+///
+unittest {
+	import vibe.data.json : Json;
+
+	class WebService {
+		// POST /item
+		void postItem() {
+			redirect("/item/1");
+		}
+	}
+
+	void run()
+	{
+		auto router = new URLRouter;
+		router.registerWebInterface(new WebService);
+
+		auto settings = new HTTPServerSettings;
+		settings.port = 8080;
+		listenHTTP(settings, router);
+	}
+}
+
 /**
 	Sets a response header.
 
@@ -394,6 +416,29 @@ void header(string name, string value)
 	assert(s_requestContext.req !is null, "header() used outside of a web interface request!");
 	alias ctx = s_requestContext;
 	ctx.res.headers[name] = value;
+}
+
+///
+unittest {
+	import vibe.data.json : Json;
+
+	class WebService {
+		// POST /item
+		Json postItem() {
+			header("X-RateLimit-Remaining", "59");
+			return Json(["id": Json(100)]);
+		}
+	}
+
+	void run()
+	{
+		auto router = new URLRouter;
+		router.registerWebInterface(new WebService);
+
+		auto settings = new HTTPServerSettings;
+		settings.port = 8080;
+		listenHTTP(settings, router);
+	}
 }
 
 /**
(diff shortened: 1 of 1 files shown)


web/vibe/web/web.d
```diff
@@ -378,6 +379,43 @@ void redirect(string url)
...
+/**
+	Sets the response status code.
+
+	Params:
+		statusCode = the HTTPStatus code to send to the client
+
+	Note that this may only be called from a function/method
+	registered using registerWebInterface.
+*/
+@property void status(int statusCode)
```
s-ludwig: I think property syntax may be a little surprising here (not that omitting `@property` would change anything, but I'd not endorse the `status = x;` syntax).

s-ludwig:
Makes sense. I'd also add a short usage example to each declaration, e.g.:
```
///
unittest {
    class WebService {
        Json postCreateItem() {
            // ...
            status(HTTPStatus.created);
            return Json(["id": 100]);
        }
    }
}
```

wilzbach:
> Makes sense. I'd also add a short usage example to each declaration, e.g.:


Alright while I was at it, I also added an example for the already existing `terminateSession`  and `redirect`.

(sorry for the rebase spam - I discovered a couple of nits).

web/vibe/web/web.d
wilzbach: Yeah I agree, but `201.status` is just awesome ;-)

s-ludwig: I'm not *quite* sure I'd agree there 😄 ...but that would actually be UFCS, not `@property`.

wilzbach:
> that omitting @property would change anything

Done. Btw have you considered doing sth. like `TaskLocal!RequestContext s_requestContext` for the new, upcoming `http` module? Not needing to specify `req` and `res` in the parameters, is very handy - though I am not sure how much overhead `TaskLocal` would bring.

s-ludwig:
I was thinking more about a way to possibly make the request/response handling more safe w.r.t. order of operations and possibly more efficient at the same time (by avoiding allocations for example). Unfortunately, D doesn't have anything particularly useful for this in its type system, though, so I'm not sure how far that will get.

For the convenience part, IMO, having the web module should be sufficient, as I'd expect all usual applications to be written in terms of `vibe.web.*` nowadays. So if anywhere, I'd expose it there. But then again, the goal of that module is to hide the bare request/response as well as possible by providing more robust high-level logic, so the question is if this would send the right message.

Performance-wise it shouldn't be critical unless it's queried numerous times per request. It's not exactly free, but definitely dwarfed by everything that happens in the request/response processing logic.

web/vibe/web/web.d
```diff
@@ -396,6 +418,29 @@ void header(string name, string value)
...
 
+///
+unittest {
+	import vibe.data.json : Json;
+
+	class WebService {
+		// POST /item
+		Json postItem() {
+			header("X-RateLimit-Remaining", 59);
+			return Json(["id": 100]);
```
s-ludwig:
```
web/vibe/web/web.d(428,10): Error: function vibe.web.web.header (string name, string value) is not callable using argument types (string, int)
web/vibe/web/web.d(429,15): Error: none of the overloads of '__ctor' are callable using argument types (int[string]), candidates
```

web/vibe/web/web.d
```diff
@@ -417,6 +462,29 @@ body
...
 
+///
+unittest {
+	import vibe.data.json : Json;
+
+	class WebService {
+		// POST /item
+		Json postItem() {
+			status(HTTPStatus.created);
+			return Json(["id": 100]);
```
s-ludwig: Also needs -> `Json(100)`

wilzbach:
> Also needs

Damn, sorry. I just ran `dub test` locally and forgot that I have to run `dub test :web` - is there an easy way that we could make `dub test` run test on all sub packages by default?

s-ludwig:
> is there an easy way that we could make dub test run test on all sub packages by default?

We should definitely do that. Of course using "--combined" also works, but ideally `dub test` would be all that one has to run. I'd even like to implement a way to integrate high-level test scripts.

PR #1697, opened 2017-02-27
[RFC] vibe.web.web: Support JSON more natively

wilzbach:
A couple of improvement ideas to help working with Json:

- allow `Json` as function parameter -> call with `req.json`
- by default an unknown return type is serialized to Json (no need to yield a compile-time error)
- allows to return strings -> `req.writeBody` (for those simple APIs where "OK" is needed)
- allow auto-deserialization of Json parameters (this is only attempted if there's a json response and then only for the first parameter)


s-ludwig:
The direction definitely makes sense. What I had in mind for quite a while now was to let the REST and web interface generators share more code - maybe in the form of using `RestInterface!T` also for web interfaces (with a few adjustments). This means that specifying a whole-body JSON parameter in web interfaces would work the same as for RESTful interfaces: `@bodyParam("p") void fun(Json p) {}`.

One reason for going that route (apart from strengthening the least-surprise factor) is that I recently had a lot of places where having `@queryParam` and `@bodyParam` in the web interface would have been really useful.

wilzbach:
> The direction definitely makes sense.

As this stalled a bit in the queue, I have split this PR into three pieces which should make it a lot easier to evaluate them in small pieces (I think the split-offs are non-controversial, only this one might be).

> This means that specifying a whole-body JSON parameter in web interfaces would work the same as for RESTful interfaces: @bodyParam("p") void fun(Json p) {}.

I agree that this wouldn't be bad, but Json is so commonly nowadays that I would prefer `void fun(Json p){}` (https://github.com/rejectedsoftware/vibe.d/issues/1853) or even auto-deserialization from Json: `void fun(MyStruct myStruct)` (this PR)

wilzbach:
> This means that specifying a whole-body JSON parameter in web interfaces would work the same as for RESTful interfaces: @bodyParam("p") void fun(Json p) {}.

FYI: this might soon be `void fun(@bodyParam Json p){}` - the respective language change has already been approved and a working PR is in the queue (https://github.com/dlang/dmd/pull/7576). It just needs a few final tweaks, so if you want to I could with-hold the param changes and make this PR, s.t. it's just about allowing to return a arbitrary type in vibe.web.web which in turn will be serialized to Json.

PR #1706, opened 2017-03-07
Mark vibe.data.json.parseJsonString as trusted

tom-tan:
It takes address of local variable `line` but it cannot be used outside this scope.


tom-tan:
Thanks!

PR #1712, opened 2017-03-15
Simple fix to REST bool query parameters

tchaloupka:
Just a simple fix to REST handler when bool query parameter is used. Not all clients generates expected "true" so it should be case insensitive.
Also sending 0 or 1 is pretty common.


s-ludwig:
Going to trust that Travis would pass if it weren't for the MongoDB compile error. The change looks good.

Issue #1715, opened 2017-03-22
Appveyor tests failing for x86 configuration on Windows.


DBankov:
Appveyor tests fail to build for x86 configuration on Windows because of the following type conversion error:
> mongodb\vibe\db\mongo\connection.d(381,25): Error: cannot implicitly convert expression (cast(ulong)msglen - (this.m_bytesRead - bytes_read)) of type ulong to uint

and most likely the fix is to use the same cast as at:
> https://github.com/rejectedsoftware/vibe.d/blob/master/mongodb/vibe/db/mongo/connection.d#L370

I'll try to prepare a PR with a fix shortly.

DBankov:
Appveyor builds are green again so closing this issue.

PR #1716, opened 2017-03-22
Fix for #1715

DBankov:
This is a fix for https://github.com/rejectedsoftware/vibe.d/issues/1715.
The change is to use the same cast as at: https://github.com/rejectedsoftware/vibe.d/blob/master/mongodb/vibe/db/mongo/connection.d#L370

### DIFF
--- a/mongodb/vibe/db/mongo/connection.d
+++ b/mongodb/vibe/db/mongo/connection.d
@@ -378,7 +378,7 @@ final class MongoConnection {
 
 		on_msg(cursor, flags, start, numret);
 		static if (hasIndirections!T || is(T == Bson))
-			auto buf = new ubyte[msglen - (m_bytesRead - bytes_read)];
+			auto buf = new ubyte[msglen - cast(size_t)(m_bytesRead - bytes_read)];
 		foreach (i; 0 .. cast(size_t)numret) {
 			// TODO: directly deserialize from the wire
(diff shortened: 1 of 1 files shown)


DBankov:
Unfortunately the fix  revealed another issue which is triggered by the vibe-core target:
> ..\..\Users\appveyor\AppData\Roaming\dub\packages\eventcore-0.8.8\eventcore\source\eventcore\drivers\winapi\sockets.d(10,7): Error: class eventcore.drivers.winapi.sockets.WinAPIEventDriverSockets interface function 'void setTargetAddress(Handle!("datagramSocket", Handle!("socket", Handle!("fd", uint, 4294967295u), Handle(4294967295u)), Handle(Handle(4294967295u))) socket, scope Address target_address) nothrow @safe' is not implemented

DBankov:
So it seems the new issue is fixed with https://github.com/vibe-d/eventcore/commit/https://github.com/vibe-d/eventcore/commit/420c361b5f8b2ded5de9e6ab166cf27ab1dada97 but we are not yet consuming the eventcore sources. I'll try to figure out what can be done.

s-ludwig:
Okay, thanks for preparing the fix! I'll merge this and will also have a look at the remaining errors on Travis. I've also tagged eventcore 0.8.9 with the WinAPI fix present. Let's see what happens...

PR #1720, opened 2017-03-24
Fix & test Meson builds

ximion:
This PR fixes the meson builds and adds a few quirk fixes as well. It also allows testing Meson in Travis (LDC only for now).

The following issues exist and have been partially resolved:
 * Meson doesn't allow absolute paths in `include_directories` anymore if they point to the source dir. This breaks a previous hack added to download 3rd-party dependencies. I added another hack to work around this for now, but I will raise this at Meson upstream now to get this resolved somehow.
 * Building Vibe with DMD+Meson apparently doesn't work since DMD doesn't recognize the versioned library parameter. I am not sure of this is a Meson or DMD issue, at time I think it's Meson not translating a flag to the DMD-specific syntax. Will need some investigation.

The other change is installing the whole source directory to `include/d` now, since previously some paths weren't correctly created, breaking software depending on Vibe.
This patch is also already applied in Debian.

Due to the issues with DMD, the Travis test builds for Meson are only enabled for LDC for now.
Since libasync doesn't have a Meson build file, Meson build-config is also libevent-only, but that has been this was before.

Please note that this PR targets the 0.7.x branch, not master!

Thanks for developing Vibe.d!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,18 @@
 language: d
 sudo: false
+dist: trusty
 
 services:
   - mongodb
   - redis-server
+
+addons:
+  apt:
+    packages:
+    - libevent-dev
+    - libssl-dev
+    - pkg-config
+    - zlib1g-dev
 
 d:
   # order: latest DMD, oldest DMD, LDC/GDC, remaining DMD versions
@@ -13,7 +22,8 @@ d:
   - dmd-2.073.2
   - dmd-2.068.2
   - ldc-1.0.0
-  - ldc-1.1.0
+  - ldc-1.1.1
+  - ldc-1.2.0-beta1
   - dmd-2.072.2
   - dmd-2.071.2
   - dmd-2.070.2
@@ -29,9 +39,22 @@ matrix:
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
   allow_failures:
     - d: dmd-beta
-    - d: ldc-1.1.0
+    - d: ldc-1.2.0-beta1
+      env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
+    - d: ldc-1.1.1
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
     - d: dmd-2.068.2
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
+
+before_install:
+  - pip3 install meson
+
+install:
+  - mkdir .ntmp
+  - curl -L https://github.com/ninja-build/ninja/releases/download/v1.7.2/ninja-linux.zip -o .ntmp/ninja-linux.zip
+  - unzip .ntmp/ninja-linux.zip -d .ntmp
+
+before_script:
+  - export PATH=$PATH:$PWD/.ntmp
 
 script: ./travis-ci.sh
--- a/meson.build
+++ b/meson.build
@@ -4,7 +4,7 @@ source_root = meson.source_root()
 build_root = meson.build_root()
 
 project_version      = '0.7.30'
-project_version_name = '0.7.30~beta.1'
+project_version_name = '0.7.30'
 project_soversion    = '0'
 
 pkgc = import('pkgconfig')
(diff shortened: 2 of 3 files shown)


s-ludwig:
Can you do a quick rebase? The base master commit had a compile error that I noticed too late.

ximion:
Done :-)

ximion:
`/usr/bin/ld: /home/travis/dlang/ldc-1.0.0/bin/../lib/libphobos2-ldc.a(concurrency.o): relocation R_X86_64_32 against `_D36TypeInfo_xE3std11concurrency7MsgType6__initZ' can not be used when making a shared object; recompile with -fPIC` - meh, looks like we are running into that issue again with LDC 1.0.0, and the other LDC-based tests are marked as "expected to fail".

ximion:
I could disable the actual build and only have the Meson configure phase in the CI - that way, the Meson file would at least be partially be tested for validity.
The real solution would of course be to fix LDC to actually build the tests correctly in it's 1.1.1 version (I don't think 1.0.0 binaries will receive an update where it's Phobos/DRuntime are built with PIC enabled).

s-ludwig:
Okay, good news. 0.7.x passes the tests with LDC 1.1.0 again.

ximion:
Okay, so I will rebase when I'm home.
The Meson build failure with the older 1.0 LDC is unfortunately a bug in the LDC binary distribution used by Travis, so there is not much I can do about it.

ximion:
Looks like the D runtime/phobos isn't compiled with PIC on the 1.1.0 LDC binaries either. Weird.

ximion:
I now altered the Travis configuration to *not* build with LDC 1.0 but instead build with LDC 1.1.1 and the beta version of LDC 1.2. Those LDC versions don't have the bug displayed before.
Now, only the LDC 1.1.1 libasync version seems to fail with:
```
Running ./vibe-d-core-test-libasync 
Task terminated with unhandled exception: Range violation
Task terminated with unhandled exception: Range violation
Task terminated with unhandled exception: Range violation
core.exception.RangeError@/home/travis/dlang/ldc-1.1.1/bin/../import/std/container/array.d(528): Range violation
----------------
Task terminated with unhandled exception: Only InterruptException supposed to be thrown!

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

The rest builds & tests fine :-)

s-ludwig:
Nice, I didn't realize that there was a 1.1.1 patch release for LDC (thought you just made a typo ;-)). Libasync doesn't properly support LDC, so all LDC+libasync can just be allowed failures (leaving them there so it's possible to notice when libasync starts to work).

LDC 1.0.0 should still stay in travis.yml, though, if the effort to selectively exclude the Meson test is low, that would be the best approach, but otherwise I'm fine with changing it into an allowed failure, too.

ximion:
LDC 1.1.1 contains precisely the patch which fixes the Meson issue Travis was running into previously - I just thought before that Travis would auto-update to patch releases, apparently that's not the case (the linking error also happens with dub and other build systems, so having it fixed was quite important).

Excluding 1.0 from running Meson just required a bit of sed magic (not sure if it will work with other D compilers though), so these builds succeed now.
Should I leave the 1.2 Beta build in there?

In any case, thanks for the support and responsiveness! :-)

ximion:
Looks like my relative-path hack in Meson wasn't necessary, as `include_directories()` acts on both the source and build directories.
Meson upstream clarified this in their Wiki and documentation, and it works like a charm (and makes the Meson file for Vibe much nicer again).

s-ludwig:
Okay, looks good. With this in Travis-CI, I can now also take care of not letting this be broken in the 0.7.x branch. I'm still not sure what the best approach will be for 0.8.x and up. There will still be deprecations and some breakage until all sub packages are in their own repository (each with a 1.x.x version).

.travis.yml
```diff
@@ -29,9 +39,22 @@ matrix:
...
-    - d: ldc-1.1.0
+    - d: ldc-1.2.0-beta1
+      env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
+    - d: ldc-1.1.1
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
     - d: dmd-2.068.2
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
+before_install:
+  - pip3 install meson
```
MartinNowak: Any version specification required, e.g. `>0.40.0`?

ximion: Well, technically >0.40, but pip will always get the latest version, so this is not an issue here.

Issue #1562, opened 2016-09-15
Double quotes in multipart request


JanVana:
Hello, we have a server implemented using Vibe. We are sending it multi-part requests from .Net, but Vibe can't read any of the uploaded files. Comparing it with the same request created by Firefox, I can see it is missing double quotes around the name and the filename parameters. Manually adding double quotes to the .Net request fixes the issue.

Would it be possible to fix this in Vibe, so that we can send requests from .Net without any hacks?

Incorrectly handled request: [BadRequest.txt](https://github.com/rejectedsoftware/vibe.d/files/474570/BadRequest.txt)
Correctly handled request: [FixedRequest.txt](https://github.com/rejectedsoftware/vibe.d/files/474585/FixedRequest.txt)

s-ludwig:
Fixed by #1725.

PR #1725, opened 2017-03-27
Fix parseMultipartForm to accept unquoted strings

tchaloupka:
Content-Disposition fields can be passed as an unquoted string in which case the current parser fails.

https://tools.ietf.org/html/rfc6266#section-4.1

So I adjusted the parser to depend on semicolon as a separator or to use the rest of the input if the semicolon is not found.

For example .Net can send requests like this:

```
POST https://192.168.0.78:8082/api/v1/foos HTTP/1.1
User-Agent: bar
Accept-Encoding: gzip, deflate
Accept-Language: en-US
Content-Type: multipart/form-data; boundary="7518e894-589d-44be-9a04-9f45c6124429"
Host: 192.168.0.78:8082
Content-Length: 55839
Expect: 100-continue
Connection: Keep-Alive

--7518e894-589d-44be-9a04-9f45c6124429
Content-Type: text/plain; charset=utf-8
Content-Disposition: form-data; name=name

Some test
--7518e894-589d-44be-9a04-9f45c6124429
Content-Type: text/plain; charset=utf-8
Content-Disposition: form-data; name=desc

Some test description
--7518e894-589d-44be-9a04-9f45c6124429
Content-Disposition: form-data; name=file; filename=CadEx_schema.png; filename*=utf-8''CadEx_schema.png

...file content...
--7518e894-589d-44be-9a04-9f45c6124429--
```

### DIFF
--- a/inet/vibe/inet/webform.d
+++ b/inet/vibe/inet/webform.d
@@ -204,6 +204,33 @@ unittest { // issue #1220 - wrong handling of Content-Length
 	assert(files["files"].filename == "file1.txt");
 }
 
+unittest { // use of unquoted strings in Content-Disposition
+	import vibe.stream.memory;
+
+	auto content_type = "multipart/form-data; boundary=\"AaB03x\"";
+
+	auto input = createMemoryStream(cast(ubyte[])(
+			"--AaB03x\r\n" ~
+			"Content-Disposition: form-data; name=submitname\r\n" ~
+			"\r\n" ~
+			"Larry\r\n" ~
+			"--AaB03x\r\n" ~
+			"Content-Disposition: form-data; name=files; filename=file1.txt\r\n" ~
+			"Content-Type: text/plain\r\n" ~
+			"Content-Length: 29\r\n" ~
+			"\r\n" ~
+			"... contents of file1.txt ...\r\n" ~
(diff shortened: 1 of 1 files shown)


tchaloupka:
This fixes #1562

s-ludwig:
To be fully correct, shouldn't there be two branches, one that searches for `"` in case the value begins with `"` and one that searches for `;` if it doesn't?

Otherwise LGTM.

tchaloupka:
I've made the branch but also it is actually needed to skip escaped quotes inside so I changed that too.

inet/vibe/inet/webform.d
```diff
@@ -220,24 +247,56 @@ struct FilePart {
...
+	//find end of quoted string
+	auto indexOfQuote(string str) {
+		ptrdiff_t res;
+
+		foreach (ch; str) {
+			if (ch == '"' && (res == 0 || str[res-1] != '\\')) return res;
+			res++;
+		}
+
+		return -1;
```
s-ludwig: Probably would be the best idea to throw an exception here. Otherwise this would result in an out-of-bounds error in `parseValue`, which terminates the application.

s-ludwig:
Thanks! I indeed missed the escaping, but this looks good now.

Issue #1724, opened 2017-03-26
WebSocket crashes when a connection gets closed


d-sobczyk:
Recently I encountered a problem with WebSockets. Our server application crashes with the following callstack:

```
[EB19FFA5:00000000 dbg] Socket event on fd 10: 33 (21178C8 vs 21178C8)
[EB19FFA5:EB1A17A5 dbg] TCP close request true open
[EB19FFA5:00000000 ERR] Task terminated with unhandled exception: Trying to acquire a TCP connection that is currently owned.
[EB19FFA5:00000000 dbg] Full error: core.exception.AssertError@../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/core/drivers/libevent2_tcp.d(145): Trying to acquire a TCP connection that is currently owned.
[EB19FFA5:00000000 dbg] ----------------
[EB19FFA5:00000000 dbg] ??:? _d_assert_msg [0x965f92]
[EB19FFA5:00000000 dbg] ../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/core/drivers/libevent2_tcp.d:145 void vibe.core.drivers.libevent2_tcp.Libevent2TCPConnection.acquire() [0x8c9931]
[EB19FFA5:00000000 dbg] ../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/core/drivers/libevent2_tcp.d:188 void vibe.core.drivers.libevent2_tcp.Libevent2TCPConnection.close() [0x8c9dd9]
[EB19FFA5:00000000 dbg] ../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/stream/wrapper.d:111 void vibe.stream.wrapper.ConnectionProxyStream.close() [0x90b78a]
[EB19FFA5:00000000 dbg] ../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/http/websockets.d:547 void vibe.http.websockets.WebSocket.startReader() [0x875668]
[EB19FFA5:00000000 dbg] ../../../.dub/packages/vibe-d-0.7.30/vibe-d/source/vibe/core/core.d:592 void vibe.core.core.makeTaskFuncInfo!(void delegate()).makeTaskFuncInfo(ref void delegate()).callDelegate(vibe.core.core.TaskFuncInfo*) [0x799c24]
[...]

PR #1727, opened 2017-03-28
Fix TCP close behavior in case of cuncurrent read+write operations

### LINKED ISSUE #1726: TCP concurrent read/write not properly terminated if peer closes connection
Most likely the cause for #1724. This happens for the libevent and libasync drivers, not for vibe-core.

s-ludwig:
Fixes #1726 and possibly #1724.

### DIFF
--- a/core/vibe/core/drivers/libasync.d
+++ b/core/vibe/core/drivers/libasync.d
@@ -1182,6 +1182,20 @@ final class LibasyncTCPConnection : TCPConnection/*, Buffered*/ {
 	override void close()
 	{
 		logTrace("Close TCP enter");
+
+		// resume any reader, so that the read operation can be ended with a failure
+		Task reader = m_settings.reader.task;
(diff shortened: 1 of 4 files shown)


core/vibe/core/drivers/libasync.d
```diff
@@ -1315,6 +1329,7 @@ final class LibasyncTCPConnection : TCPConnection/*, Buffered*/ {
 	}
 
 	void acquireReader() {
+		assert(!m_settings.writer.isWaiting, "Acquiring reader that is already in use.");
```
ZombineDev: I have almost no knowledge of the code, but shouldn't this be `m_settings.reader.isWaiting`?

s-ludwig: I was about to refactor those methods into the `Waiter` struct... and then went the copy/past route ;)

s-ludwig:
Okay, fixed and passing. Merging in.

PR #1730, opened 2017-04-03
Made DefaultFilters public

WebFreak001:
this makes it possible for the user to define their own custom render functions with the new diet renderer (namely you can create a render function using compileHTMLDietString now)

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -248,7 +248,7 @@ version (Have_diet_ng)
 	import diet.traits;
 
 	@dietTraits
-	private struct DefaultFilters {
+	struct DefaultFilters {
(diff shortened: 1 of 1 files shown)


WebFreak001:
@s-ludwig bump

http/vibe/http/server.d
```diff
@@ -248,7 +248,7 @@ version (Have_diet_ng)
 	import diet.traits;
 
 	@dietTraits
-	private struct DefaultFilters {
+	struct DefaultFilters {
```
s-ludwig: This should be renamed to `DefaultDietFilters`, so that it makes sense in the public context.

WebFreak001:
I lost my branch on this, gonna make a new PR

PR #1737, opened 2017-04-08
Output null instead of undefined with Json.undefined serialization

### LINKED ISSUE #1735: Json.undefined serializes to invalid Json
This is related to: #1442

```D
auto json = Json();
writeln(json.serializeToPrettyJson);
```

Json should allways serialize to a valid Json I think. So probably to `null` in this case?

tchaloupka:
Fixes #1735

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1949,6 +1949,12 @@ struct JsonStringSerializer(R, bool pretty = false)
 unittest
 {
 	assert(serializeToJsonString(double.nan) == "null");
+	assert(serializeToJsonString(Json()) == "null");
+	assert(serializeToJsonString(Json(["bar":Json("baz"),"foo":Json()])) == `{"bar":"baz"}`);
+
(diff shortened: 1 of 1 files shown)


data/vibe/data/json.d
```diff
@@ -2010,7 +2016,6 @@ void writeJsonString(R, bool pretty = false)(ref R dst, in Json json, size_t lev
 			dst.put('{');
 			bool first = true;
 			foreach (string k, ref const Json e; json.byKeyValue) {
-				if( e.type == Json.Type.undefined ) continue;
```
tchaloupka: I've removed this because it will not serialize the attribute at all and I think, that this should be handled by `@optional` attribute.

thaven: This is incorrect because `Json.undefined` literally means the key should not be present.

s-ludwig: Seconded. This was done to be consistent with the ECMA script spec, from which the `undefined` value originates.

Issue #1743, opened 2017-04-14
vibe.core.net.1726 fails intermittently when compiled with dmd-2.072.2 and the libevent/libasync configurations


DBankov:
See https://travis-ci.org/rejectedsoftware/vibe.d/jobs/221892384 and https://travis-ci.org/rejectedsoftware/vibe.d/jobs/221892386 for failures examples. While I'm unable to reproduce these on my machines it seems to me that increasing the overall timeout to e.g. 2 seconds will help to avoid these.  Actually 2 seconds seems a lot to me for this test but since it is obvious that such case is hit I'd rather leave a potential performance optimization as a separate task.

DBankov:
Seems this is not an issue anymore. Closing.

PR #1744, opened 2017-04-14
Fix for #1743. Increase the overal timeout for vibe.core.net.1726 to 2 s.

DBankov:


DBankov:
It seems this is not an issue anymore (which is great) so closing the PR as well.

PR #1746, opened 2017-04-23
Allow building with Meson and DMD

ximion:
Older Meson versions had a bug which prevented library builds when DMD was used, which is why DMD+Meson builds were excluded in Vibe.d so far.
I resolved the issue in Meson, so we can now test with DMD as well, yay!

(I also updated the library versions for the current in development version)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ matrix:
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
 before_install:
-  - pip3 install meson
+  - pip3 install meson>=0.40
 
 install:
   - mkdir .ntmp
--- a/meson.build
+++ b/meson.build
@@ -3,8 +3,8 @@ project('Vibe.d', 'd')
 source_root = meson.source_root()
 build_root = meson.build_root()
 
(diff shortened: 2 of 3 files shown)


s-ludwig:
Nice! And good that you also caught the version number change along the way. I'll have to automate this somehow.

ximion:
There is one other change we might need to make, and that is installing Vibe into a private subdirectory in `/usr/include/d/`. The currently solution is neat because you don't need to configure much to use Vibe, but `/usr/include/d` being in the default include path messes with the real Vibe build sometimes, as the D compiler prefers the old installed bits in `/usr/include/d` over the new changed ones in the Vibe project folder.

I'll submit a patch for that later, maybe.

For the version numbers, they are used to properly version the shared libraries. After the discussion about this on the D mailinglist, I wonder how useful these shared libs will be longterm - at the moment they save a great deal of build time and are nice to use, but when the D ABI changes we will run into problems.

Usually in other projects the global version number is only set in the Meson file and then inserted into the source code by the build process, either by defining a macro or by writing a bit of source/header to contain the version number.
Maybe we can find a way where you would only need to set the version number once and have it used by Dub and Meson.

PR #1754, opened 2017-05-04
Handle "before" annotations on REST interface classes.

### LINKED ISSUE #1753: "before" annotations on class member functions not handled by the REST server interface
If an `@before` annotation is put on the class method as opposed to the declaration in the interface that the class derives from, the attribute is ignored.

See also https://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/43066/

s-ludwig:
Fixes #1753.

### DIFF
--- a/utils/vibe/internal/meta/funcattr.d
+++ b/utils/vibe/internal/meta/funcattr.d
@@ -12,6 +12,7 @@
 module vibe.internal.meta.funcattr;
 
 import std.traits : isInstanceOf, ReturnType;
+import vibe.internal.meta.traits : RecursiveFunctionAttributes;
 
 /// example
 unittest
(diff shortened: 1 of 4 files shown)


Geod24:
Shouldn't `@before` on interface be discouraged ? The problem is that it describes server-only behavior so it can easily make the interface unusable for client generation (e.g. if the `@before` function have dependency on the DB, it drags it into the client as well).

s-ludwig:
Definitely should be discouraged in their current form. The problem is that only those on the interface seem to have worked for a while now and I didn't notice when that became the case (don't use them anywhere anymore, AFAIR). So I'd let them go through a full deprecation cycle (doc -> warn -> error) to make sure they only appear on the class (and maybe any base class).

Issue #1758, opened 2017-05-12
openssl 1.1.0 binding issue


burner:
Openssl made the jump to 1.1.0 they removed some symbols. Which makes it impossible to use on arch linux without some non sustainable hacks.

`CRYPTO_num_locks'
`CRYPTO_set_id_callback'
`CRYPTO_set_locking_callback'
`SSL_get_ex_new_index'
`SSL_library_init'
`SSL_load_error_strings'
`SSLv23_client_method'
`SSLv23_server_method'

deimos is out of date and without git version tags.
I looked into creating bindings with dstep, but thats gone be a real pain.
IMHO it might be easier to create the definitions we need from openssl ourself and drop the deimos binding. Comments?

wilzbach:
> Openssl made the jump to 1.1.0 they removed some symbols. Which makes it impossible to use on arch linux without some non sustainable hacks.

There are two simple "hacks" to solve the problem temporarily:
- `versions "Have_Botan"`
- `versions "VibeNoSSL"` (use Nginx etc. as reverse proxy)

burner:
requestHTTP might need https as well. I'm not sure how to use nginx for that.
At some point we need update openssl or drop it. And dropping it is not really an option IMO.

s-ludwig:
Are there any known examples of how to properly achieve the same effect that those functions have with 1.1.x? To be honest that's the main roadblock - documentation is awful, code is awful and it's more or less impossible to get the initialization sequence right without an example.

Updating the Deimos headers indeed doesn't sound like fun. The huge mess that the OpenSSL code base is presumably makes that quite a PITA. BTW, I had to [clone the Deimos repository](https://github.com/s-ludwig/openssl) anyway to be able to make version tags a reality. There have been no responses in the main repository.

burner:
yeah openssl init looks like a big mess.
void SSL_load_error_strings() {} does not do anything anymore for example

I might have some time on Sunday to see if I can get dstep+vim to get a new version up and running

CyberShadow:
> There have been no responses in the main repository.

Just ping someone (with the necessary privileges), sheesh :)

CyberShadow:
FWIW, I would certainly be thankful for an update to the Deimos bindings, since I use those in `ae`! Right now I have to link DFeed against openssl 1.0.0 explicitly.

MartinNowak:
Duplicate of #1748

Issue #1748, opened 2017-04-26
Linking error with OpenSSL 1.1


o3o:
I've last vibe.d:
```
$ cd vibe.d/
$ git describe
v0.8.0-beta.5-39-gd39a3601
```
After updating my ArchLinux:
```
$ cd ./examples/app_skeleton
$ dub 

Linking...
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:419: error: undefined reference to 'SSLv23_client_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:420: error: undefined reference to 'SSLv23_client_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:424: error: undefined reference to 'SSLv23_client_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:425: error: undefined reference to 'SSLv23_client_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:432: error: undefined reference to 'SSLv23_server_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:433: error: undefined reference to 'SSLv23_server_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:435: error: undefined reference to 'SSLv23_server_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:436: error: undefined reference to 'SSLv23_server_method'
../../../.dub/packages/vibe-d-0.8.0-beta.5/vibe-d/stream/vibe/stream/openssl.d:674: error: undefined reference to 'get_rfc3526_prime_2048'
../../../.dub/packages/openssl-1.1.5_1.0.1g/openssl/deimos/openssl/safestack.d:140: error: undefined reference to 'sk_num'
../../../.dub/packages/openssl-1.1.5_1.0.1g/openssl/deimos/openssl/safestack.d:142: error: undefined reference to 'sk_value'
[...]

PR #1759, opened 2017-05-12
got vibe to compile against openssl 1.1.0

burner:
just a crude hack, but I think nothing is fundamentally wrong with it.

https://github.com/rejectedsoftware/vibe.d/issues/1758

### DIFF
--- a/stream/vibe/stream/openssl.d
+++ b/stream/vibe/stream/openssl.d
@@ -46,6 +46,77 @@ version(VibeForceALPN) enum alpn_forced = true;
 else enum alpn_forced = false;
 enum haveALPN = OPENSSL_VERSION_NUMBER >= 0x10200000 || alpn_forced;
 
+// openssl 1.1.0 hack
+version (Have_openssl) {
+extern(C) const(SSL_METHOD)* TLS_client_method();
+
+const(SSL_METHOD)* SSLv23_client_method() {
+	return TLS_client_method();
+}
+
+extern(C) const(SSL_METHOD)* TLS_server_method();
+
+const(SSL_METHOD)* SSLv23_server_method() {
+	return TLS_server_method();
+}
+
+// this does nothing in > openssl 1.1.0
+void SSL_load_error_strings() {}
+
+extern(C)  int OPENSSL_init_ssl(ulong opts, const void* settings);
+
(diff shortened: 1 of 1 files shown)


burner:
is there any way to see the openssl version

s-ludwig:
There is `OpenSSL_version_num()` - hopefully available on all involved versions.

burner:
I meant at link time. The tests seem to fail to link.

burner:
libressl looks like an alternative, has anyone had a look at that one yet?
they wrote a new tls library (tls.h) https://github.com/daniloegea/libressl-tls-api-examples shows an example

wilzbach:
> I meant at link time. The tests seem to fail to link.

@burner: if you want to build a hack on top of a hack, you could use a `preBuildsCommands` and utilize the output of:

```
readelf -d /usr/lib/libssl.so | grep SONAME | sed -E 's/.*libssl[.]so[.](.*)].*/\1/'
```

e.g. by storing it in a gitignored D file

burner:
@wilzbach thanks I will try that

s-ludwig:
I've pulled this now and embedded it into a `version (VibeUseOpenSSL11)` until a better option supported by DUB is available.

s-ludwig:
Maybe this can be solved in a proper way by extending the libs field to support version specifications or similar. The actual version could also be reported through a mechanism analog to https://github.com/dlang/dub/issues/6.

BTW, @burner thanks a lot for having done the fiddling to get this working!

burner:
@s-ludwig nice thanks

MartinNowak:
> Maybe this can be solved in a proper way by extending the libs field to support version specifications or similar. The actual version could also be reported through a mechanism analog to dlang/dub#6.
> 
> BTW, @burner thanks a lot for having done the fiddling to get this working!

One supported way would be to let people choose different openssl dub versions, e.g. `1.1.5+1.0.2j` or `2.0.1+1.1.0f`. Unfortunately this is not automatic.
The libs field already does support versioned libraries, see the workaround in https://github.com/rejectedsoftware/vibe.d/issues/1748#issuecomment-297647832. Most ld versions do support this syntax by now.

wilzbach:
Why not list the openssl library dynamically?
Dlang-requests does this: https://github.com/ikod/dlang-requests/issues/45#issuecomment-317208703

PR #1760, opened 2017-05-12
Use a custom multi-region allocator for the HTTP server.

s-ludwig:
This works around a memory leak issue in `std.experimental.allocator.building_blocks.allocator_list.AllocatorList` ([17393](https://issues.dlang.org/show_bug.cgi?id=17393)). There is still a very high amount of virtual memory use, but reserved memory appears to be back to normal.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -1770,9 +1770,13 @@ private void handleHTTPConnection(TCPConnection connection, HTTPListenInfo liste
 		bool keep_alive;
 
 		() @trusted {
-			import std.algorithm.comparison : max;
-			auto request_allocator_s = AllocatorList!((n) => Region!GCAllocator(max(n, 1024)), NullAllocator).init;
-			scope request_allocator = request_allocator_s.allocatorObject;
+			import vibe.internal.utilallocator: RegionListAllocator;
+
+			version (VibeManualMemoryManagement)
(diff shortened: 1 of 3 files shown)


http/vibe/http/server.d
```diff
@@ -1733,7 +1733,7 @@ private void listenHTTPPlain(HTTPServerSettings settings, HTTPServerContext cont
 
 
 private void handleHTTPConnection(TCPConnection connection, HTTPListenInfo listen_info)
-@safe {
+@trusted {
```
wilzbach: I know that there are many `@trusted` necessary for Vibe.d to work, but imho it just feels like we lost the battle if we need to put `@trusted` at nearly every function :S

thaven: I agree to that. Even more off-topic: `@trusted` functions are actually a D language misdesign. Would have been better to have unsafe blocks like C#. Fortunately, we can emulate those in D with `() @trusted { ... } ();`.
Okay, to get on topic: Further down this diff, such an `@trusted`-block was already there, and all changes are inside that block, so I don't really understand the reason of `@trusted` here.

s-ludwig: This one was actually left over from a change where the allocator was instantiated outside of the request handler loop (will remove). But as much as the need for `@trusted` is annoying in the general case, here it actually serves its purpose, as custom allocators are involved (`deallocateAll()` in particular). Would of course be nice to be able to implement `@safe` wrappers with DIP1000...

PR #1767, opened 2017-05-22
Mark BotanTLSContext constructor as trusted

Yoplitein:
Fixes
```
../vibe.d/stream/vibe/stream/tls.d(102,11): Error: @safe function 'vibe.stream.tls.createTLSContext.createBotanContext' cannot call @system constructor 'vibe.stream.botan.BotanTLSContext.this'
```


skoppe:
Had to do it as well. LGTM

PR #1769, opened 2017-05-26
Have versions block prevent importing openssl when using botan

skoppe:
I had to move around the version block in openssl.d and add a few version blocks in tls.d to prevent pulling in openssl when using botan.

It all started on ArchLinux yesterday where I found out the deimos bindings for openssl are incompatible with the latest openssl version on arch.

Then I started trying to compile vibe.d without openssl. This turned out to be quite hard since no matter what I did, dub kept settings the 'Have_openssl' version, which in turn made vibe.d import the deimos bindings. (Is this correct behaviour in dub? settings the version for optional dependencies? I think it should not since I am using the botan dependency in my own dub.selections.json).

By adding the version blocks in this PR, no openssl code is imported when using botan.

### DIFF
--- a/stream/vibe/stream/openssl.d
+++ b/stream/vibe/stream/openssl.d
@@ -6,7 +6,11 @@
 	Authors: Sönke Ludwig
 */
 module vibe.stream.openssl;
-version(Have_openssl):
+version(Have_botan)
+{
+} else version(VibeNoSSL)
+{
+} else version(Have_openssl):
 import vibe.core.log;
 import vibe.core.net;
 import vibe.core.stream;
--- a/stream/vibe/stream/tls.d
+++ b/stream/vibe/stream/tls.d
@@ -30,8 +30,8 @@ import core.sync.mutex;
 import core.thread;
 
 version (VibeNoSSL) {}
-else version(Have_openssl) version = OpenSSL;
 else version(Have_botan) version = Botan;
+else version(Have_openssl) version = OpenSSL;
 
 
 /// A simple TLS client


wilzbach:
Had to do this as well. Thanks for trying to get this upstream!

thaven:
Sounds more like a dub-level problem to me. Openssl needs to be removed from dub.selections.json manually to build without it. Maybe it would be better to have separate subconfigurations for openssl and botan, so that when building the botan configuration, openssl dependency is not listed at all (and vice versa). That should prevent both openssl and botan being selected.

skoppe:
Agreed. It is a dub issue. Dub should not set the Have_openssl version when the dependency isn't used. Will raise an issue on dub repo.

skoppe:
BTW, what is the issue with the CI? Is it flaky?

wilzbach:
> Sounds more like a dub-level problem to me. Openssl needs to be removed from dub.selections.json manually to build without it. Maybe it would be better to have separate subconfigurations for openssl and botan, so that when building the botan configuration, openssl dependency is not listed at all (and vice versa). That should prevent both openssl and botan being selected.

Agreed, but this fix works now, is tiny and easily revertable if the dub problem gets fixed. However, as I doubt this will happen soon, I suggest to go with the "workaround" for the time being ;-)
After all, it's just `6+,2-`

skoppe:
Also see https://github.com/dlang/dub/pull/1148

arpie42:
Well, I'm totally confused.  I have spent hours trying to build a vibe.d example http server on ArchLinux using this workaround but openssl kept being pulled in during linking.  I finally ran dub --vverbose and trawled through the output only to find this :

```
Found dependency botan 1.12.8
            Collecting dependencies for botan
              Found dependency botan:base 1.12.8
              Collecting dependencies for botan:base
                Found dependency botan-math 1.0.3
                Collecting dependencies for botan-math
                Version selection for dependency openssl (openssl) of botan:base is missing.
```

WTF?  Is openssl now a required dependency of botan?  I'm trying to force a build using botan 1.12.0 instead of 1.12.8 but that's causing issues of its own... I'll keep trying.

arpie42:
Okay, I can't get any earlier versions of botan to compile so giving up on that as it is not a sensible solution anyway.
It looks like somewhere between version 1.12.0 and 1.12.8 an alternative configuration was added to the dub.json in the botan package.  The new configuration is called 'full_openssl' and it had a required dependency of openssl.  Even if this configuration is not being used (the default configuration is 'full'), it's presence in the dub.json file is forcing openssl to be added to the dub.selections.json file, even after I have manually removed it.  If I edit botan's dub.json file within the dub package and manually remove the dependency specification, I can successfully compile and link and run a vibe.d server (woohoo!).  But doing so will certainly break the intended functionality of the optional configuration 'full_openssl'.  Not a problem for me... but I have no idea what a correct solution might be.
I am a near-total newbie in the world of D, and a complete beginner in the world of DUB so if I should be reporting this as a bug somewhere more relevant, please point me in the right direction.

skoppe:
@arpie42 I have got a vibe.d application using botan to compile correctly on ArchLinux using this PR.

The dub behaviour you are seeing is correct. Dub does dependency resolution before pruning configurations. As a result dependencies from unused (sub)configurations do get pulled. However they should not be included after the correct set of configs is determined. Unfortunately, due to a bug in dub, optional default dependencies always get included whenever any other package has a dependency on that same optional package, even if that is in a config you aren't using.

in dlang/dub#1148 I tried to solve it, but it has problems. Sonke has suggested another solution but I haven't had time to dig into it.

wilzbach:
Ping @s-ludwig - can we go with this fix? Is anything block this PR?

arpie42:
@wilzbach As far as I can see this PR does nothing useful.  The problem is openssl being pulled in as a dependency of botan, as described in [dlang/dub#1148](https://github.com/dlang/dub/pull/1148) (I'm sure you've already read that, too).

In other words, I can compile vibe.d without this PR by manually removing the openssl dependency from botan's dub.json file.  Not a great solution, but it works.  (As already mentioned in this thread, it is also necessary to manually remove openssl from the dob.selections.json file).  I'm using vibe-d-master and botan-1.12.9 but I'm pretty sure I had the same results when I tried with earlier versions.

If I install this PR, it appears to make no difference here.  It is possible that I have missed some nuance, though, and would be happy to be corrected.

Would a good general solution to this sort of issue would be to add a feature to dub to allow forced overrides of dependencies (somewhat akin to removing a dependency from dub.selections but more explicit and more permanent).  This is clearly not the 'right' answer to this particular problem but it would provide a clear and effective work-around while waiting for a better solution!

arpie42:
@wilzbach My apologies - I have tried again this morning and it seems I didn't apply this PR correctly.  It does indeed do what it claims.
@s-ludwig +1 for this to be pulled asap, please?

arpie42:
@skoppe Can you explain why swapping the version blocks in tls.d works here?  AFAICS the behaivour should be identical whatever the ordering of the two lines that you have swapped.  What am I missing?

skoppe:
The issue is that when you want vibe.d to use botan (by specifing it in your projects' dependencies), both `Have_botan` and `Have_openssl` are set (due to dub bug). In its original position this means that tls.d will always set `version = OpenSSL;`.

By changing the ordering only `version = Botan;` is going to be set.

arpie42:
Okay, got it.  Thanks for taking the time to answer my silly questions!

Do you have a reference to the 'dub bug' that you mention?  As far as I understand it, your PR dlang/dub#1148 is dealing with a slightly different take on this same issue (focussing on the linker issue, rather than the versions() flags issue)?  And, is it really a bug?  It looks to me like dub is doing exactly what it should.  What is actually needed is a dub enhancement/feature to go beyond having simple optional dependencies, to enable explicit handling of alternative dependencies (ie. a way to specify that openssl and botan are mutually exclusive so that Have_openssl and Have_Botan are never both set).  This is made more complex by the fact that Botan itself may depend on OpenSSL... so Have_openssl needs to be set when compiling Botan but not when compiling everything else.  If that can be achieved, the linker will sort itself out even if openssl is listed as a dependency, won't it?

Regardless of the underlying issue, this PR does seem to provide a good workaround.

skoppe:
The bug I mentioned is indeed dlang/dub#1148. The issue there is not with the linker, it is with dub setting both Have_botan and Have_openssl. This happens because botan has a configuration with openssl as a dependency, which messes up dub's internal state and it starts compiling all dependencies of your project with Have_openssl.

The solution is as Sönke said. I just haven't had time to prepare an PR as of yet.

arpie42:
Thank you again for taking the time to help me to understand what is going on here.

It seems that there are two independent issues at play:
-  One is dependency related - how does dub determine which dependencies should be placed in dub.selections.json (and therefore which Have_blah flags should be set).  This is your dlang/dub#1148.
- The other issue is that of how to use the version() flags to compile the relevant code in vibe-d.

For example, let's not forget that there is a "VibeNoSSL" version flag available.  If set, then openssl is still pulled in by dub as it is a default optional dependency (despite not being required).  The linker presumably is intelligent enough to not include openssl in the final executable as no references are present in the compiled code (is this correct?).  It would be impossible for dub to determine that openssl is not a dependency, so we have to manually remove it from the dub.selections.json file if we want to skip compiling the openssl library.  So far, so good.

Instead of using version(Have_botan) and version(Have_openssl) in the vibe-d code, which are themselves dependent on what is present in dub.selections.json, would it not be clearer to introduce a version(VibeBotanSSL) to complement version(VibeNoSSL)?  There would be an implicit version(VibeOpenSSL) flag too but as this is defined as the default optional library, it would not actually need to be set.  The desired SSL library could then be explicitly specified by setting versions: "VibeBotanSSL" in the application dub.sdl/json file.  This, all 

Would it be useful if I demonstrate what I mean in a PR or do you think this is a non-starter?  Sorry if I seem to be flogging this issue to death but it has been a great opportunity for me to get to grips with how dub works.

s-ludwig:
Okay, I finally gave this a thought and would propose an alternative approach. The `version(VibeOpenSSL)` approach goes in the right direction, but still having a link-time dependency could of course still result in issues, generally. The current approach has the drawback of disallowing the admittedly uncommon but valid use case to compile in support for multiple TLS providers.

The workaround would simply drag out the openssl.d and botan.d files into individual sub packages. The dependency to openssl/botan would be non-optional in there, but the dependency to those new sub packages would be optional/optional-default. That sidesteps the issue and also allows to get rid of the `version (Have_xxx)` within the TLS implementations. The drawback, if botan is also a separate sub package, is that people using it will have to adjust their package recipe to depend on that new sub package (changing botan to a sub package could be deferred, though).

For the proper fix, there is an initial idea in dlang/dub#1148, but although that could fix this particular issue, it's not powerful enough to be a general solution. I'll have to think about that some more.

s-ludwig:
I would just word it a bit differently, not the wrong versions are set, but the optional openssl dependency is regarded as being chosen no matter what the user preference is, because the dependency in botan overrides that choice, even if a different configuration is being built. That in turn of course results in the versions being defined accordingly.

Now by shifting the "optional" attribute from openssl to a separate intermediate package that in turn references openssl, botan can't affect the "chosen" state (now of this new package) anymore. The result will be that openssl will not be included in the build, unless either the new intermediate package appears explicitly in dub.selections.json, or the botan openssl configuration is chosen. That in turn will result in `Have_openssl` to stay undefined.

Because that intermediate package (let's call it `vibe-d:openssl`) references openssl as non-optional, the `version (Have_openssl)` in openssl.d *can* then be dropped, because it can be assumed to always be set.

skoppe:
Ok, I understand correctly.

s-ludwig:
Question is if we should still squeeze this into 0.8.0, or rather release a quick 0.8.1 afterwards. It's kind of a breaking change, so having it in 0.8.0 would generally be cleaner. I'll open a PR to see if any unforeseen issues come up.

s-ludwig:
Turns out that this isn't possible because of cyclic dependencies between tls.d and openssl.d/botan.d. That means that there seemingly is no way to fix this without taking away the possibility to use both TLS implementations in a single build. If that is true, a simple and clean approach would be to use configurations to select between openssl/botan/notls.

s-ludwig:
PR: #1823 Let's see if this breaks any platforms...

skoppe:
I see. And to break the cyclic dependency you would have to add yet another subpackage with only the interfaces and struct that openssl/botan depend on to connect them with vibe.

#1823 is a better option. If it works :)

arpie42:
#1823 seems an excellent solution - it makes it clear that NoSSL, Botan and OpenSSL are alternatives, not just independant optional flags.  Thank you for taking the time to do this.

But... I am having trouble getting dub to use any configuration apart from the default openssl.  My dub.sdl file looks like this:
```
name "rpvibed"
description "Arpie's Vibe.d web server"
authors "arpie"
copyright "Copyright © 2017, arpie"
license "proprietary"
lflags "-fuse-ld=gold"
dependency "vibe-d" version="~>0.7.0"
subconfiguration "vibe-d:stream" "botan"
```
Dub is ignoring that subconfiguration line.  What am I doing wrong?

s-ludwig:
> Dub is ignoring that subconfiguration line. What am I doing wrong?

This is something that really needs to be fixed. `subConfiguration` can only operate on a dependency that has been specified in the same scope and not on indirect dependencies. Taking into account the new tls sub package, it has to look like this:
```
name "rpvibed"
description "Arpie's Vibe.d web server"
authors "arpie"
copyright "Copyright © 2017, arpie"
license "proprietary"
lflags "-fuse-ld=gold"
dependency "vibe-d" version="~>0.7.0"
dependency "vibe-d:tls" version="~>0.7.0"
subConfiguration "vibe-d:tls" "botan"
```

So the first thing DUB needs to do here is to issue an error whenever a package gets referenced that isn't directly referenced as a `dependency`. The next thing would be to support configuration selection from within the `dependency` directive: `dependency "vibe-d:tls" config="botan"`. This would basically just be syntax sugar, but it supports a very common case.

PR #1772, opened 2017-05-29
Test botan on Travis

wilzbach:
Analog to https://github.com/dlang/dub-registry/pull/214 and https://github.com/dlang/dub/pull/1138

The main idea is to use daily crons and avoid being broken in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,18 +28,21 @@ d:
   - dmd-2.073.2
   - dmd-2.072.2
   - dmd-2.071.2
-  - dmd-beta
 
 env:
-  - VIBED_DRIVER=libevent BUILD_EXAMPLE=1 RUN_TEST=1
-  - VIBED_DRIVER=vibe-core BUILD_EXAMPLE=1 RUN_TEST=1
-  - VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
+  - VIBED_DRIVER=libevent VIBED_SSL=openssl BUILD_EXAMPLE=1 RUN_TEST=1
+  - VIBED_DRIVER=vibe-core VIBED_SSL=openssl BUILD_EXAMPLE=1 RUN_TEST=1
+  - VIBED_DRIVER=libasync VIBED_SSL=openssl BUILD_EXAMPLE=0 RUN_TEST=0
 
 matrix:
   include:
     - d: dmd
       # https://issues.dlang.org/show_bug.cgi?id=13742
       env: DFLAGS="-cov -version=VibedSetCoverageMerge" DUB_BUILD_MODE="singleFile"
+    - d: dmd
+    - env: VIBED_SSL=botan BUILD_EXAMPLE=0 RUN_TEST=1
+    - d: ldc
+    - env: VIBED_SSL=botan BUILD_EXAMPLE=0 RUN_TEST=1
   exclude:
     - d: ldc-1.2.0
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
@@ -50,7 +53,6 @@ matrix:
   allow_failures:
     - d: dmd-beta
     - d: ldc-1.3.0
-      env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
 script: ./travis-ci.sh
 after_success:
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -9,6 +9,12 @@ DUB_ARGS=${DUB_ARGS:-}
 # Check for trailing whitespace"
 grep -nrI --include=*.d '\s$'  && (echo "Trailing whitespace found"; exit 1)
 
+if [ "${VIBED_SSL:-openssl}" == "botan" ]; then
+    grep -qF "Have_botan" stream/dub.sdl || echo 'versions "Have_botan"' >> stream/dub.sdl
+    # Make botan mandatory (i.e. non-optional)
+    sed -E 's/(dependency "botan".*) optional=true/\1/' -i stream/dub.sdl
+fi
+
 # test for successful release build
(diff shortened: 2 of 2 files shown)


travis-ci.sh
```diff
@@ -2,6 +2,12 @@
 
 set -e -x -o pipefail
 
+if [ "${VIBED_SSL:-openssl}" == "botan" ]; then
+    grep -qF "Have_botan" stream/dub.sdl || echo 'versions "Have_botan"' >> stream/dub.sdl
+    # Make botan mandatory (i.e. non-optional)
+    sed -E 's/(dependency "botan".*) optional=true/\1/' -i stream/dub.sdl
+fi
```
ZombineDev: Wow, that's quite an ugly hack. I guess you had no other choice..

.travis.yml
```diff
@@ -19,29 +19,28 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.074.0-rc1
+  - dmd-2.074.0
   - dmd-2.070.2
-  - ldc-1.1.1
```
ZombineDev: Can you keep LDC 1.1.1?

wilzbach:
Well I think [auto-cancellation of outdated commits](https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation) should be enabled for this repo:

image

Moreover as it now takes a very long time per job, I guess we have to be very careful with how many builds on OSX we enable..

Last, but not least, my main point was to show that non-hardcoded CI tags + daily rebuilds are a good way to catch regressions early on as currently Vibe.d doesn't build with `dmd-nightly`, e.g.

```
stream/vibe/stream/memory.d(19,80): Error: cannot implicitly convert expression processAllocator() of type shared(ISharedAllocator) to std.experimental.allocator.IAllocator
stream/vibe/stream/memory.d(48,42): Error: cannot implicitly convert expression processAllocator() of type shared(ISharedAllocator) to std.experimental.allocator.IAllocator
utils/vibe/utils/array.d(715,35): Error: function vibe.utils.array.ArraySet!(Task).ArraySet.AW (IAllocator a) is not callable using argument types (shared(ISharedAllocator))
core/vibe/core/drivers/libevent2.d(656,23): Error: template instance vibe.utils.hashmap.HashMap!(Thread, ThreadSlot, DefaultHashMapTraits!(Thread)) error instantiating
utils/vibe/utils/array.d(715,35): Error: function vibe.utils.array.ArraySet!ulong.ArraySet.AW (IAllocator a) is not callable using argument types (shared(ISharedAllocator))
dmd failed with exit code 1.
```

The first error originates from dlang/phobos#5355
If anyone wants to look at it, on the journey I already discovered a [nasty ICE](https://issues.dlang.org/show_bug.cgi?id=17451).

But there's more fun as LDC >= 1.0.0 segfaults with `libasync`:

```
CoreTaskFiber was terminated unexpectedly: Range violation
```

.travis.yml
wilzbach: The [build matrix](https://travis-ci.org/rejectedsoftware/vibe.d/builds/237027475) is already extremely long. Do you have any specific reason to keep ldc-1.1.1?

travis-ci.sh
wilzbach: Yup, at least I couldn't think of one and this one is not too bad: it's only two lines long and even idempotent.

If wanted I could also add an else clause that idempotently resets this change, but imho (1) this script is only used for the CI anyways and (2) if used locally `git checkout` already works well.

MartinNowak: The Have_botan part shouldn't be necessary if botan is a dependency.
Usually you have to add it to your dub.selections.json to select an optional dependency, so `dub upgrade` and some `jq` command might work, not much better than your sed line though. We definitely need a dub command for that stuff.
Would also make sense to change openssl's `default=true` to `default=false`.
But weirdly enough botan itself lists openssl as dependency, so you can't really get rid of it.
https://github.com/etcimon/botan/blob/32f01e942dc44b8fbbc5eb7fbd285619c57a42af/dub.json#L127

wilzbach: > But weirdly enough botan itself lists openssl as dependency, so you can't really get rid of it.

But only for the `full_openssl` configuration, not by default.

wilzbach:
Removed `dmd-nightly`, `dmd-beta` and `ldc-beta` form the list to avoid breaking Vibe.d by accident. And as Vibe.d is active enough, I removed `dmd` and `ldc` as well ...

s-ludwig:
Good news is that the Botan hack can also be replaced by `--override-config vibe-d:tls/botan` now due to #1823.

One thing that still bothers me is that the list of tests is now even longer. What about making this an `includes:` configuration that is just tested with "dmd" and maybe one with "ldc"?

wilzbach:
> Good news is that the Botan hack can also be replaced by --override-config vibe-d:tls/botan now due to #1823.

Hmm - how do we trick DUB in ignoring this warning then:

> Package 'vibe-d:tls', marked for configuration override, is not present in dependency graph.

Maybe a workaround isn't that bad after all?

> One thing that still bothers me is that the list of tests is now even longer. What about making this an includes: configuration that is just tested with "dmd" and maybe one with "ldc"?

Makes sense.

s-ludwig:
> Package 'vibe-d:tls', marked for configuration override, is not present in dependency graph.

For the unit tests it would just be a matter of specifying it for only those sub packages that use `:tls` directly or indirectly. For the high level tests... it would make sense to run only those tests that use TLS anyway, so we'd probably need a marker file there.

> Maybe a workaround isn't that bad after all?

The other problem is that it doesn't work at all anymore, OpenSSL will still be used as the TLS provider. It could of course be adjusted to simply remove the "openssl" configuration of "tls/dub.sdl".

Geod24:
Closing as Travis is gone.

PR #1774, opened 2017-05-29
Fix Vibe.d nightly (phobos/5355: ISharedAllocator)

wilzbach:
Since 2.074.0 I could observe at least three regressions:
- [ ] `processAllocator` is now `shared` and an `ISharedAllocator` (https://github.com/dlang/phobos/pull/5355)
- [ ] `resolveInternalPointer` accepts now only`const` (https://github.com/dlang/phobos/pull/5321)
- [x] For running `dub test` and its resulting linking errors (https://github.com/dlang/phobos/pull/5432 has been merged).

This is a dirty workaround against the ISharedAllocator changes and allows vibe.d to be built again.

### DIFF
--- a/core/vibe/internal/allocator.d
+++ b/core/vibe/internal/allocator.d
@@ -1,9 +1,24 @@
 module vibe.internal.allocator;
 
-public import std.experimental.allocator;
+public import std.experimental.allocator : allocatorObject, CAllocatorImpl, dispose,
+	   expandArray, IAllocator, make, makeArray, shrinkArray, theAllocator;
+
 public import std.experimental.allocator.building_blocks.allocator_list;
 public import std.experimental.allocator.building_blocks.null_allocator;
 public import std.experimental.allocator.building_blocks.region;
 public import std.experimental.allocator.building_blocks.stats_collector;
 public import std.experimental.allocator.gc_allocator;
 public import std.experimental.allocator.mallocator;
+
+__gshared IAllocator _processAllocator;
+
+shared static this()
+{
+    import std.experimental.allocator.gc_allocator : GCAllocator;
+    _processAllocator = allocatorObject(GCAllocator.instance);
+}
+
+@property IAllocator processAllocator()
+{
+    return _processAllocator;
+}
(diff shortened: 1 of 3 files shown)


ZombineDev:
Thanks for doing this! If it's not too much work, can you separate the two workarounds in separate commits and a a link to the respective PRs that introduced the breakage?

wilzbach:
> Thanks for doing this! If it's not too much work, can you separate the two workarounds in separate commits and a a link to the respective PRs that introduced the breakage?

Not at all.
I also reworded the the intro of this PR to reflect the current situation.

SerialVelocity:
@wilzbach @ZombineDev There seems to be another issue now:
```
../../../.dub/packages/vibe-d-0.8.0-beta.7/vibe-d/core/vibe/core/sync.d(143,2): Error: @safe constructor 'vibe.core.sync.LocalTaskSemaphore.this' cannot call @system destructor 'vibe.core.sync.LocalTaskSemaphore.~this
../../../.dub/packages/vibe-d-0.8.0-beta.7/vibe-d/http/vibe/http/server.d(832,2): Error: @safe constructor 'vibe.http.server.HTTPServerRequest.this' cannot call @system destructor 'vibe.http.server.HTTPServerRequest.~this'
```

wilzbach:
> @wilzbach @ZombineDev There seems to be another issue now:

Yeah - isn't that fun? ;-)
Also in Vibe.d 0.7.31-rc.2 [there is/was also an ICE](https://issues.dlang.org/show_bug.cgi?id=17489) - luckily it's just a regression from https://github.com/dlang/phobos/pull/5427 now.

MartinNowak:
The Allocator part would be needed for 2.075.0 to work.

wilzbach:
I just realized that `disposeMultidimensionalArray` and `makeMultidimensionalArray` has been added in 2.074 (https://github.com/dlang/phobos/pull/4982), so it's better to avoid importing them (this lead to the CI build failures).

s-ludwig:
@wilzbach Thanks a lot! I'll merge and then rename the vibe.d version of `processAllocator` to avoid name clashes/confusion.

MartinNowak:
Using the thread-safe `theAllocator` might be an option as well.

s-ludwig:
@MartinNowak: Unfortunately fails (it was actually originally used, but I had to switch to `processAllocator` because of this): https://github.com/dlang/phobos/pull/5519

PR #1777, opened 2017-06-09
Re-up TTL when opening an existing session

schveiguy:
When reading the description of expirationTime, it says "The duration without access after which a session expires." To me, this means if you continue to "access" the session, it resets the duration. However, in the implementation, it only sets the TTL once, when the session is created. This means that accessing it does not re-up the timeout.

This fixes that, so now every time you open the session, it also resets the timeout. This is what I would expect session expiration to mean.


s-ludwig:
Definitely the right behavior!

PR #1778, opened 2017-06-09
Remove costly default key

schveiguy:
Currently when using redis for sessions, the default key is the session id, and the default value is the session id. The session id's length is 86 characters long. This means that for every session, redis will require 86 * 3 characters minimum of memory, instead of just 86 characters. The key name and data isn't important at all. It's just a minimal default key so the session can be created (redis does not allow empty hashmaps).

This replaces the very long name and value with a small one. While it may not make a huge difference (comes out to about 16.5MB difference if you have 100k sessions), it seems unnecessarily wasteful. In my case, at least, this "default" key is more than 2x larger than all the session data I am storing.

### DIFF
--- a/redis/vibe/db/redis/sessionstore.d
+++ b/redis/vibe/db/redis/sessionstore.d
@@ -37,7 +37,7 @@ final class RedisSessionStore : SessionStore {
 	Session create()
 	{
 		auto s = createSessionInstance();
-		m_db.hmset(s.id, s.id, s.id); // set place holder to avoid create empty hash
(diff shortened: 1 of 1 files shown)


schveiguy:
I also changed hmset to hset, as you are only setting one key.

s-ludwig:
Thanks! Incidentally I stumbled over the same line a few days ago, but didn't take the time to change it. Good to have this taken care of.

schveiguy:
Thanks!

PR #1781, opened 2017-06-17
test that versions strings are equal

MartinNowak:
- also add small helper script to update and tag a new version

### DIFF
--- a/scripts/tag_version.sh
+++ b/scripts/tag_version.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -ueo pipefail
+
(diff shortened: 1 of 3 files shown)


MartinNowak:
BTW, you might want to turn on Travis-CI's auto-cancellation.
[The Travis CI Blog: Get faster results when testing your latest commits with Auto Cancellation](https://blog.travis-ci.com/2017-03-22-introducing-auto-cancellation)

s-ludwig:
Thanks! That will come in very handy. Also good that this will cause pre-release versions to get the proper version string.

PR #1782, opened 2017-06-17
update latest dmd to 2.074.1

MartinNowak:


MartinNowak:
What about just using `dmd` instead?

wilzbach:
> What about just using dmd instead?

Have you seen this?

https://github.com/rejectedsoftware/vibe.d/pull/1772

PR #1800, opened 2017-06-29
Remove all trailing whitespace

wilzbach:
I don't feel strongly about this, my vim just complained about it & suggesting to fix it would have taken longer than the PR itself.

This adds a check to the Travis script, s.t. no new whitespace can be checked in.

### DIFF
--- a/core/vibe/core/concurrency.d
+++ b/core/vibe/core/concurrency.d
@@ -1198,7 +1198,7 @@ unittest {
 
 unittest {
 	import vibe.core.core : sleep;
-	
+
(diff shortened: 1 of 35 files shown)


ZombineDev:
Needs (probably manual) rebase, otherwise LGTM

wilzbach:
> Needs (probably manual) rebase, otherwise LGTM.

Thanks, I can simply drop the commit & rerun the trailing whitespace removal command. For reference, it's really simple:

```
sed 's/[ \t]*$//' -i **/*.d
```

PR #1801, opened 2017-06-29
More lazy loading in HTTPServerRequest

wilzbach:
Similarly to https://github.com/rejectedsoftware/vibe.d/pull/1677

Some values are actually a bit problematic thanks to D's type concept (mutable methods!), for example consider the method `fullURL` method:

```d
@property URL fullURL()
const @safe {
     URL url;
...
	url.username = this.username; //  Error: mutable method vibe.http.server.HTTPServerRequest.username is not callable using a const object
	url.password = this.password; //  Error: mutable method vibe.http.server.HTTPServerRequest.password is not callable using a const object
        return url;
}
```

I initially had the idea to hack the type system with something like `assumeConst`:

```d
import std.traits;
private auto assumeConst(T)(T t) @trusted
    if (isFunctionPointer!T || isDelegate!T)
{
    enum attrs = functionAttributes!T | FunctionAttribute.const_;
    return cast(SetFunctionAttributes!(T, functionLinkage!T, attrs)) t;
}
```

But. of course, without success

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -441,13 +441,13 @@ enum HTTPServerOption {
 	parseURL                  = 1<<0,
 	/// Fills the `.query` field in the request
 	parseQueryString          = 1<<1 | parseURL,
-	/// Fills the `.form` field in the request
+	/// Fills the `.form` field in the request (deprecated)
 	parseFormBody             = 1<<2,
 	/// Fills the `.json` field in the request (Deprecated, lazily parsed)
 	parseJsonBody             = 1<<3,
-	/// Fills the `.files` field of the request for "multipart/mixed" requests
+	/// Fills the `.files` field of the request for "multipart/mixed" requests (deprecated)
 	parseMultiPartBody        = 1<<4, // todo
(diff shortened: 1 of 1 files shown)


http/vibe/http/server.d
```diff
@@ -795,20 +803,33 @@ final class HTTPServerRequest : HTTPRequest {
...
+			return _form.get;
+		}
 
-			Remarks:
-				This field is only set if HTTPServerOption.parseFormBody is set
-				and if the Content-Type is "multipart/form-data".
+		private Nullable!FormFields _form;
+
+		private void parseFormAndFiles() @safe {
+			parseFormData(_form, _files, contentType, bodyReader, MaxHTTPHeaderLineLength);
```
s-ludwig: If I'm not overlooking something, I think this needs to set `_form` to `.init` first, or it would try to access the null field.

wilzbach:
> If I'm not overlooking something, I think this needs to set _form to .init first, or it would try to access the null field.

Excellent catch! Fixed.

PR #1804, opened 2017-06-29
Allow to run CodeCov on Vibe.d

wilzbach:
This allows coverage from simple test suite as a first step.

IIRC CodeCov will send the notifications automatically, but it's better to enable them as web hook and GitHub integration (better stability).

Currently this will fail, because trailing whitespace managed to sneak in & then the test suite gets prematurely stopped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,10 @@ env:
   - VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
 matrix:
+  include:
+    - d: dmd
+      # https://issues.dlang.org/show_bug.cgi?id=13742
+      env: DUB_ARGS="--build=unittest-cov --build-mode=singleFile"
   exclude:
     - d: ldc-1.0.0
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -34,6 +34,10 @@ env:
   - VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
 matrix:
+  include:
+    - d: dmd
+      # https://issues.dlang.org/show_bug.cgi?id=13742
+      env: DUB_ARGS=--build=unittest-cov --build-mode=singleFile
```
s-ludwig: Not very familar with YAML, but form the error message it seems like this may need some form of quotes.

wilzbach:
> Not very familar with YAML, but form the error message it seems like this may need some form of quotes.

Yep, thanks - though apparently I did it correctly for https://github.com/vibe-d/vibe-core/pull/19

wilzbach:
Wohoo it's working: https://codecov.io/gh/rejectedsoftware/vibe.d/tree/ed302d3543feb7db727e5ee8699039d95b221fd5

PR #1806, opened 2017-06-30
libevent udp ip multicast implementation

skoppe:
Adds necessary functions on UdpConnection to control basic IP multicast.

I am working on getting a PR into druntime to get rid of the ugly conditional imports.

### DIFF
--- a/core/vibe/core/drivers/libevent2.d
+++ b/core/vibe/core/drivers/libevent2.d
@@ -62,6 +62,16 @@ version(Windows)
 	alias EWOULDBLOCK = WSAEWOULDBLOCK;
 }
 
+version(OSX)
+{
+	import std.c.osx.socket : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
+} else version(Posix)
+{
+    import std.c.linux.socket : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
+} else version(Windows)
+{
+	// IP_ADD_MEMBERSHIP and IP_MULTICAST_LOOP are included in winsock(2) import above
+}
 
 final class Libevent2Driver : EventDriver {
 @safe:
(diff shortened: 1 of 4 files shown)


core/vibe/core/net.d
```diff
@@ -434,6 +434,24 @@ interface UDPConnection {
 	ubyte[] recv(ubyte[] buf = null, NetworkAddress* peer_address = null);
 	/// ditto
 	ubyte[] recv(Duration timeout, ubyte[] buf = null, NetworkAddress* peer_address = null);
+
+	version (VibeLibeventDriver) {
```
s-ludwig: Instead of making the interface specific to libevent, the other drivers should just have simple stubs with `assert(false, "TODO!");` inside. It would also be good to update the `vibe-core` package accordingly, but I can do that, too.

core/vibe/core/drivers/libevent2.d
```diff
@@ -1086,6 +1103,34 @@ final class Libevent2UDPConnection : UDPConnection {
 		}
 	}
 
+	void addMembership(string group)
```
s-ludwig: At least here, the parameter should be of type `ref NetworkAddress`, so that the name resolution happens outside, and the driver function code can focus on the core functionality. But for convenience, `UDPConnection` could have a second `final` overload that accepts a string instead.

core/vibe/core/net.d
skoppe: Thought about it as well. Will adjust.

core/vibe/core/drivers/libevent2.d
skoppe: Makes sense. Do you want me to squash and force update? Or create additional commit?

s-ludwig: An additional commit is okay (I'm happy as long as the history stays reasonably readable).

skoppe:
Still haven't had time to finish this. It is on my radar though.

PR #1808, opened 2017-07-02
Improve coverage testing: merge coverage files from all runs + run examples

wilzbach:
Follow-up to https://github.com/rejectedsoftware/vibe.d/pull/1804

> BTW, are the .lst files actually updated incrementally, or do they get overwritten for each compiler invocation?

By default, they get overwritten except one sets [`coverSetMerge`](https://dlang.org/phobos/core_runtime.html#.dmd_coverSetMerge).
Sadly there's an option to specify this at compile-time, so one needs to resort to more or less ugly workarounds - even the setup at DMD does this](https://github.com/dlang/dmd/blob/2c02eb49f738611689ab89353e0ae64d2e826ffe/src/ddmd/mars.d#L1733). In this case we can't use `D_Coverage` as, of course, libraries want to be able to set their own settings.

FWIW the LDC people want to [add a custom frontend flag](https://github.com/ldc-developers/ldc/pull/1672/files) for this.

If I have done this correctly, the coverage should increase drastically (that's why I have split it up into two PRs).


s-ludwig:
Hm, looks like `--parallel` isn't accepted by "dub test". I'll open a PR there to add it.

wilzbach:
So it seems that this worked and the total coverage got increased by 11.4%:

image

### Before

image

### After 

image

I am not sure why CodeCov didn't send a status notification. I guess it needs to be activated as integration to send a status?
-> https://github.com/apps/codecov

FYI:
- CodeCov does combine the coverage output from multiple runs, but I just enabled one build for now to save time 
- in case you didn't know, their browser extension is pretty amazing: https://github.com/codecov/browser-extension (it shows the coverage info directly on the GH diff).

s-ludwig:
Ah, that starts to look like a better starting point to add more tests! I've authorized CodeCov for vibe.d and the vibe-d organization now. Will have a look at the extension, I didn't really use CC actively because the actual coverage is hidden behind another step.

wilzbach:
> Ah, that starts to look like a better starting point to add more tests! I've authorized CodeCov for vibe.d and the vibe-d organization now.

Thanks a lot!

Btw if you are okay with this approach / workaround to set the merge option in druntime, I will submit a similar PR to vibe-core ;-)

s-ludwig:
Yeah, this sounds reasonable. I should also revise the DUB solution to see if that actually works as intended, I don't think there is a call to `dmd_setMerge` anywhere.

PR #1809, opened 2017-07-02
Add CodeCov badge to README

wilzbach:
Didn't think about this in the initial PR, sorry.
At least it's a good opportunity to test whether CodeCov now sends status updates ;-)

Preview: 

image

Of course, once [Travis build 3613](https://travis-ci.org/rejectedsoftware/vibe.d/builds/249414450) (the build for #1808 on master) has passed, the coverage badge will update itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Visit the website at <http://vibed.org/> for more information.
 [![DUB Package](https://img.shields.io/dub/v/vibe-d.svg)](https://code.dlang.org/packages/vibe-d)
 [![Posix Build Status](https://travis-ci.org/rejectedsoftware/vibe.d.svg?branch=master)](https://travis-ci.org/rejectedsoftware/vibe.d)
 [![Windows Build Status](https://ci.appveyor.com/api/projects/status/cp2kxg70h54pga9d/branch/master?svg=true)](https://ci.appveyor.com/project/s-ludwig/vibe-d/branch/master)
-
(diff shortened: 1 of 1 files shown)


s-ludwig:
Not sure if this actually already a good idea from a publicity POV. It kind of leaves the false impression that this is rather badly tested. I'd say it makes sense to work towards >50% coverage before actively advertising it.

But, BTW, shouldn't there be an additional CC check for this PR, like it happens in the DUB repository? I double-checked and this repository *is* enabled in the CC settings.

wilzbach:
> Not sure if this actually already a good idea from a publicity POV. It kind of leaves the false impression that this is rather badly tested. I'd say it makes sense to work towards >50% coverage before actively advertising it.

Fair point. Feel free to close this PR after we have verified that CodeCov sends the checks then ;-)

> But, BTW, shouldn't there be an additional CC check for this PR, like it happens in the DUB repository? I double-checked and this repository is enabled in the CC settings.

CodeCov waits until Travis completes - after all, the respective build which generates the listing files (and uploads them to CodeCov) hasn't run yet:

image

Once the CIs have finished it should (hopefully) post status updates.

s-ludwig:
Ah okay, I thought that it would at least display a pending check.

Geod24:
Closing as it's unlikely to be merged without a significant effort to make the coverage %age go higher (it's currently 40ish percent).

PR #1811, opened 2017-07-02
Remove deprecated stdc import

wilzbach:
Probably not seen because no developers actively compiles Vibe.d on Windows and looked at the deprecation warnings.

See also: https://github.com/dlang/phobos/pull/5532

I am doing this blindly and relying on AppVeyor for testing.

### DIFF
--- a/core/vibe/core/drivers/libevent2_tcp.d
+++ b/core/vibe/core/drivers/libevent2_tcp.d
@@ -45,11 +45,11 @@ private {
 			alias INADDR_ANY = core.sys.windows.winsock2.INADDR_ANY;
 			alias IN6ADDR_ANY = core.sys.windows.winsock2.IN6ADDR_ANY;
 		} else {
-			import std.c.windows.winsock;
+			import core.sys.windows.winsock2;
 			// make some neccessary parts of the socket interface public
-			alias in6_addr = std.c.windows.winsock.in6_addr;
-			alias INADDR_ANY = std.c.windows.winsock.INADDR_ANY;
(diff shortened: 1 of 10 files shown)


core/vibe/core/drivers/libevent2_tcp.d
```diff
@@ -45,11 +45,11 @@ private {
...
 		} else {
-			import std.c.windows.winsock;
+			import core.sys.windows.winsock2;
 			// make some neccessary parts of the socket interface public
-			alias in6_addr = std.c.windows.winsock.in6_addr;
-			alias INADDR_ANY = std.c.windows.winsock.INADDR_ANY;
-			alias IN6ADDR_ANY = std.c.windows.winsock.IN6ADDR_ANY;
+			alias in6_addr = core.sys.windows.winsock2.in6_addr;
+			alias INADDR_ANY = core.sys.windows.winsock2.INADDR_ANY;
+			alias IN6ADDR_ANY = core.sys.windows.winsock2.IN6ADDR_ANY;
```
ZombineDev: A selective import would be equivalent to writing those aliases by hand, btw.

wilzbach: Yep, but I did these replacements automatically and I didn't want to have anything that could potentially block them.

s-ludwig:
I just noticed this now - the reason why those deprecations came up is not because I didn't notice/fix them before, but because of https://issues.dlang.org/show_bug.cgi?id=17614

PR #1810, opened 2017-07-02
Add convenience constructor to HTTPServerSettings.

s-ludwig:
Allows to specify the bind interface and port as a single constructor argument

PR #1812, opened 2017-07-03
Set the coverageDestPath for all files to the root folder

wilzbach:
Follow-up to https://github.com/rejectedsoftware/vibe.d/pull/1810#issuecomment-312599635

Not sure whether this helps to improve things or if CodeCov already merges the lst files automagically.

### DIFF
--- a/core/vibe/core/core.d
+++ b/core/vibe/core/core.d
@@ -2062,6 +2062,9 @@ unittest {
 // a long version
 version(VibedSetCoverageMerge)
 shared static this() {
-	import core.runtime : dmd_coverSetMerge;
+	import core.runtime : dmd_coverDestPath, dmd_coverSetMerge;
+	import std.path : buildPath, dirName;
 	dmd_coverSetMerge(true);
+	static immutable coverageDestDir = __FILE_FULL_PATH__.dirName.buildPath("..", "..", "..");
+	dmd_coverDestPath(coverageDestDir);
 }


s-ludwig:
At least for "dub test" it could be really useful to let DUB get an optional (merge) destination path. It could then inject a `shared static this` in the compiled code that does the necessary druntime fiddling. I guess that could save quite some manual work in this area. Of course that would potentially still leave the separate tests, if CC indeed doesn't merge those results automatically...

wilzbach:
> I guess that could save quite some manual work in this area

Or we simply could push for a `-cov-dest` and `-cov-merge` CLI flags at DMD - these hypothetical flags would save us most of the troubles here, but I am not sure whether this will happen in the near future if neither of us isn't pushing it.

s-ludwig:
True, that would even be better.

s-ludwig:
Although for some reason the CodeCov status isn't shown here, this actually made the coverage drop by 2 percent points for some reason. This is definitely too hard to get right currently.

s-ludwig:
Source. https://codecov.io/gh/rejectedsoftware/vibe.d/pulls
for some reason clicking on the PR there results in 404, although CodeCov is authorized for this repo.

wilzbach:
> Although for some reason the CodeCov status isn't shown here, this actually made the coverage drop by 2 percent points for some reason. This is definitely too hard to get right currently.

Yep, what do you propose? Adding the runtime fiddling by default in DUB's `unittest-cov` mode?

> for some reason clicking on the PR there results in 404, although CodeCov is authorized for this repo.

I have already written the CodeCov support about this. At least this seems to be a problem on their side.

wilzbach:
FYI there's https://github.com/dlang/druntime/pull/1916 now

PR #1813, opened 2017-07-03
Added unix domain socket support to http.proxy

georgi-d:
- changed the HTTPReverseProxySettings class to hold a destination URL instead
of (host, port)
- added reverseProxyRequest overload with URL
- updated the proxy handler logic
- made getFilteredHost() public

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -1118,7 +1118,7 @@ final class HTTPClientResponse : HTTPResponse {
 }
 
 /** Returns clean host string. In case of unix socket it performs urlDecode on host. */
-private auto getFilteredHost(URL url)
+package auto getFilteredHost(URL url)
 {
 	version(UnixSocket)
 	{
--- a/http/vibe/http/proxy.d
+++ b/http/vibe/http/proxy.d
@@ -40,9 +40,12 @@ void listenHTTPReverseProxy(HTTPServerSettings settings, HTTPReverseProxySetting
 /// ditto
 void listenHTTPReverseProxy(HTTPServerSettings settings, string destination_host, ushort destination_port)
 {
+	URL url;
+	url.schema = "http";
+	url.host = destination_host;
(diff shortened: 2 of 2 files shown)


http/vibe/http/client.d
```diff
@@ -1118,7 +1118,7 @@ final class HTTPClientResponse : HTTPResponse {
 }
 
 /** Returns clean host string. In case of unix socket it performs urlDecode on host. */
-private auto getFilteredHost(URL url)
+auto getFilteredHost(URL url)
```
s-ludwig: I'd leave this as `package` for now, because the place here isn't really fitting for a public API.

http/vibe/http/proxy.d
```diff
@@ -196,20 +196,29 @@ HTTPServerRequestDelegateS reverseProxyRequest(HTTPReverseProxySettings settings
...
 /**
 	Provides advanced configuration facilities for reverse proxy servers.
 */
 final class HTTPReverseProxySettings {
-	/// The destination host to forward requests to
-	string destinationHost;
-	/// The destination port to forward requests to
-	ushort destinationPort;
+	/// The destination URL to forward requests to
+	URL destination;
```
s-ludwig: This should get some backwards-compatibility properties:
```
/// Scheduled for deprecation - use `destination.host` instead.
@property ushort destinationPort() const { return destination.port; }
/// ditto
@property void destinationPort(ushort port) { destination.port = port; }

/// Scheduled for deprecation - use `destination.host` instead.
@property string destinationHost() const { return destination.host; }
/// ditto
@property void destinationHost(string host) { destination.host = host; }
```

To make that fully work, there would also need to be a default value of `URL("http", "")` for `destination`.

georgi-d: Thanks for the review. I will address the comments.

georgi-d:
I have addressed the comments.

PR #1814, opened 2017-07-04
Update .travis.yml

John-Colvin:
The scheduling isn't perfect but it's ok, I reckon there'll be an overhang of about 1/2 of one ldc run atm but it'll change anyway as new versions are added.


John-Colvin:
Woops, didn't see https://github.com/rejectedsoftware/vibe.d/pull/1772 before, closing for now

PR #1816, opened 2017-07-04
Allow listenHTTP to accept a convenience setting string

wilzbach:
Follow-up to https://github.com/rejectedsoftware/vibe.d/pull/1810

Now that `HTTPServerSettings` accept a string, `listenHTTP` can do so as well :tada: 

Instead of adding five new overloads, I opted for templates.

### DIFF
--- a/tests/vibe.http.server.listenHTTP/source/app.d
+++ b/tests/vibe.http.server.listenHTTP/source/app.d
@@ -1,0 +1,22 @@
+import vibe.core.core;
+import vibe.core.core;
+import vibe.core.log;
+import vibe.http.client;
+import vibe.http.server;
+import vibe.stream.operations : readAllUTF8;
+
+shared static this()
+{
+	listenHTTP(":11721", (scope req, scope res) {
+		res.writeBody("Hello world.");
+	});
+
+	runTask({
+		scope (exit) exitEventLoop();
+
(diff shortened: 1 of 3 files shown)


tests/vibe.http.server.listenHTTP/source/app.d
```diff
@@ -0,0 +1,22 @@
+import vibe.core.core;
+import vibe.core.core;
+import vibe.core.log;
+import vibe.http.client;
+import vibe.http.server;
+import vibe.stream.operations : readAllUTF8;
+
+shared static this()
+{
+	listenHTTP(":11721", (scope req, scope res) {
```
wilzbach: I copied the hard-coded port from other tests - how about adding something like [`getFreePort`](https://github.com/dlang-bots/dlang-bot/blob/master/test/utils.d#L26) to Vibe.d?

s-ludwig: Or supporting `":0"` and providing a way to retrieve the actual listen port.

wilzbach:
>  s-ludwig added the auto-merge label 3 hours ago

@s-ludwig are you sure that you configured the webhook correctly here? I can't find any events in the log.
The config should look roughly like this:

image

s-ludwig:
Yay, finally worked!

PR #1821, opened 2017-07-06
Initialize parseURL and parseQueryString lazily

wilzbach:
This makes the last two options of HTTPServerSettings lazy. As mentioned [before](https://github.com/rejectedsoftware/vibe.d/pull/1801#issue-239615384), lazy intitialization costs us `const`ness of the methods.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -2113,13 +2113,15 @@ private bool handleRequest(InterfaceProxy!Stream http_stream, TCPConnection tcp_
 			}
 		}
 
+        // eagerly parse the URL as its lightweight and defacto @nogc
+		auto url = URL.parse(req.requestURL);
+		req.queryString = url.queryString;
+		req.username = url.username;
+		req.password = url.password;
+
 		// URL parsing if desired
 		if (settings.options & HTTPServerOption.parseURL) {
-			auto url = URL.parse(req.requestURL);
 			req.path = urlDecode(url.pathString);
-			req.queryString = url.queryString;
-			req.username = url.username;
-			req.password = url.password;
 		}
 
 		// lookup the session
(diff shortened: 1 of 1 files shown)


http/vibe/http/server.d
```diff
@@ -800,29 +800,51 @@ final class HTTPServerRequest : HTTPRequest {
...
+			_username = url.username;
+			_password = url.password;
+			urlParsed = true;
+		}
 
-			Remarks: This field is only set if HTTPServerOption.parseURL is set.
+		/** The _path part of the URL.
 		*/
-		string path;
+		ref string path() @safe {
```
wilzbach: There's a bit of code that expects these variables to be mutable (hence `ref`), e.g.

```d
http/vibe/http/auth/basic_auth.d(100,4): Error: function vibe.http.server.HTTPServerRequest.username () is not callable using argument types (string)
http/vibe/http/auth/digest_auth.d(119,4): Error: function vibe.http.server.HTTPServerRequest.username () is not callable using argument types (string)
http/vibe/http/server.d(391,2): Error: function vibe.http.server.HTTPServerRequest.path () is not callable using argument types (string)
http/vibe/http/server.d(392,2): Error: function vibe.http.server.HTTPServerRequest.queryString () is not callable using argument types (string)
http/vibe/http/server.d(393,2): Error: function vibe.http.server.HTTPServerRequest.username () is not callable using argument types (string)
http/vibe/http/server.d(394,2): Error: function vibe.http.server.HTTPServerRequest.password () is not callable using argument types (string)
```

s-ludwig:
I'd say lets postpone the URL/path part for now, as the best solution may still shift w.r.t. the path/URL implementation. The query string makes sense in any case, though.

What I think makes sense for `parseURL` even now, is to always parse the URL, eagerly (it's de-facto `@nogc` and very lightweight), but to keep `req.path` tied to `parseURL` for now, because that actually may allocate memory.

But `.path` is ugly anyway. With vibe-core, this could be an `InetPath`, which doesn't allocate upon construction, and would also behave correctly w.r.t. URL encoding. Currently there are issues with encoded slashes within path segments (in the router and arguably also in the file server). So a transition path to `InetPath` should be the goal (not sure if `alias toString this; ` in `InetPath` is a good idea).

PR #1824, opened 2017-07-06
Json.opIndex: Document behavior if key is not found

Marenz:
Documented that it returns Type.undefined objects if the key wasn't found.
Also added an example to the unittest.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -326,6 +326,7 @@ struct Json {
 	/**
 		Allows direct indexing of object typed JSON values using a string as
 		the key.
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks! I just made some small whitespace fixups. Will be merged once the tests finish.

wilzbach:
> Will be merged once the tests finish.

Unfortunately this doesn't work well with spurious failures. This one is from the registry not being accessible (we really need to improve this, will reply to Martin's mail later):

```
Root package vibe-d:core references unknown package libasync
```

PR #1825, opened 2017-07-06
Fix vibe.stream.operations.readUntilGeneric to use IOMode.all

skoppe:
Recently IOMode argument was added to read function, but readUntilGeneric was not updated.

### DIFF
--- a/stream/vibe/stream/operations.d
+++ b/stream/vibe/stream/operations.d
@@ -488,7 +488,7 @@ private @safe void readUntilGeneric(R, InputStream)(InputStream stream, ref R ds
 		auto str = stream.peek(); // try to get some data for free
 		if( str.length == 0 ){ // if not, read as much as possible without reading past the end
 			nread = min(least_size, end_marker.length-nmatched, buf.length, max_read);
-			stream.read(buf[0 .. nread]);
+			stream.read(buf[0 .. nread], IOMode.all);
 			str = buf[0 .. nread];
 			bytes_read += nread;
 		} else if( str.length > max_read ){


s-ludwig:
The fix should probably be in the stream class instead, and maybe there is a bug in `isInputStream`, because there is an overload of `read` that has only one argument and forwards to the two argument overload with `IOMode.all`:
- vibe-d:core: https://github.com/rejectedsoftware/vibe.d/blob/4d952f7abdf72098a4e1919cfd18b8a6ec9213f7/core/vibe/core/stream.d#L123
- vibe-core: https://github.com/vibe-d/vibe-core/blob/cb169b850231891df90111ff1469a96e385080f4/source/vibe/core/stream.d#L160

In which context did the error occur?

skoppe:
```D
auto stdin = new StdinStream();
auto line = stdin.readLine(size_t.max, "\x0a");
```

skoppe:
readUntilSmall also does a read(..., IOMode.all). In fact, there are about 8 places where this pattern occurs.

isInputStream's definition is so simple, it is hard to have a bug there.

When I add the following to the StdFileStream class however, it does work:

```D
void read(scope ubyte[] dst) @safe
{
	read(dst, IOMode.all);
}
```

Which is plain silly since the interface ALSO has this defined.

s-ludwig:
I've added aliases now to make the overload of the interface available. I don't think this C++ inherited behavior makes any sense here, except for being consistent with the the case of a base class instead of a base interface.

skoppe:
Of course, that was it.

PR #1827, opened 2017-07-07
[RFC] Make Nullable @optional by default

wilzbach:
The following pattern is sadly happens in [real life](https://github.com/dlang-bots/dlang-bot/pull/130):

```d
struct Foo {
  @optional Nullable!bool flag;
}
```

Which imho is redundant as `Nullable` already implies that the parameter is `@optional`.

### DIFF
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -760,7 +760,8 @@ private template deserializeValueImpl(Serializer, alias Policy) {
 				});
 			}
 			foreach (i, mname; SerializableFields!(T, Policy))
-				static if (!hasPolicyAttribute!(OptionalAttribute, Policy, TypeTuple!(__traits(getMember, T, mname))[0]))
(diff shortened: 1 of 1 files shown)


s-ludwig:
Unfortunately we have to distinguish between a nullable and an optional field. I really wish Phobos had a simple optional type that behaves as a zero/one element range. That would naturally fit the role you have in mind here and would also be very useful in general.

wilzbach:
> I really wish Phobos had a simple optional type that behaves as a zero/one element range. 

There was a PR:

https://github.com/dlang/phobos/pull/3915


And a bit related https://github.com/dlang/phobos/pull/5154

wilzbach:
@s-ludwig if you plan to move sth. like https://github.com/dlang/phobos/pull/3915 to vibe.d (I am happy to help), then let's keep this open, otherwise I guess closing is the only option for now :cry:

s-ludwig:
Okay, I've opened #1849 to keep track of this. I'd like to see if a Phobos solution is realistic first. If not, a custom solution might be the only way...

PR #1835, opened 2017-07-11
Remove vibe-d:diet.

s-ludwig:
Should ideally have been removed together with 0.8.0 already. There is no reason anymore to rely on the old code rather than diet-ng.

See also #1834.

PR #1837, opened 2017-07-13
Remove explicit thread destruction code in the libasync driver.

s-ludwig:
See the discussion in #1835. Libasync now uses the default thread pool of `std.parallelism`, which is shut down automatically.


wilzbach:
Wow that was a quick resolution - could you maybe release a new tag, s.t. these changes are propagated to the Project-Tester? Thanks!

s-ludwig:
Let's wait with this until #1838 is merged, too.

s-ludwig:
0.8.1-alpha.1 is now tagged and also available on code.dlang.org.

wilzbach:
> 0.8.1-alpha.1 is now tagged and also available on code.dlang.org.

Thanks a lot. I will try to keep an eye on the Project-Tester to check whether this solved our problems.

PR #1841, opened 2017-07-13
Initialize variables in shared static this()

wilzbach:
If `connectMongo` with SASL (e.g. SCRAM) is called in a static shared constructor, it will nicely segfault.

### DIFF
--- a/mongodb/vibe/db/mongo/sasl.d
+++ b/mongodb/vibe/db/mongo/sasl.d
@@ -20,7 +20,12 @@ import vibe.crypto.cryptorand;
 
 @safe:
 
-private SHA1HashMixerRNG g_rng;
+private SHA1HashMixerRNG g_rng()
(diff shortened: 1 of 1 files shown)


s-ludwig:
Should be both, in `shared static this` for the main thread, and in `static this` for other threads. Or, and that's probably the better solution, it could just be created lazily. It currently uses a file descriptor per thread and risks a startup failure even for applications that don't use the SASL authentication.

wilzbach:
> Or, and that's probably the better solution, it could just be created lazily. 

Ok. Should I maybe even move this lazy initialization into `cryptorand.d` as e.g. `sha1HashMixer`?

s-ludwig:
That would generally make sense, but I'm not yet sure where the crypto package should be heading, so keeping it private in this module is the safer option for now.

PR #1842, opened 2017-07-15
Support full Redis IANA URI scheme

wilzbach:
For reference: many PaaS providers supply the Redis password directly in the URL and it took me quite a bit to figure out that the Redis implementation simply doesn't look at the `password` part of the Redis URI.

Regarding this PR, I am not sure how add test for this that work on all machines. Happy about ideas.

I copied the interesting bits from the [spec](https://www.iana.org/assignments/uri-schemes/prov/redis) for future readers.

BTW what do you think about adding a convenient UFCS-friendly way to build URLs?
I would love to write: `"https://google.com".url`, but then again `url` is a very commonly used variable.

### DIFF
--- a/redis/vibe/db/redis/redis.d
+++ b/redis/vibe/db/redis/redis.d
@@ -40,12 +40,53 @@ RedisClient connectRedis(string host, ushort port = RedisClient.defaultPort)
 	Returns a Redis database connecction instance corresponding to the given URL.
 
 	The URL must be of the format "redis://server[:port]/dbnum".
+
+	See_also: $(LINK2 https://www.iana.org/assignments/uri-schemes/prov/redis, Redis URI scheme)
 */
 RedisDatabase connectRedisDB(URL url)
 {
 	auto cli = connectRedis(url.host, url.port != 0 ? url.port : RedisClient.defaultPort);
-	// TODO: support password
-	return cli.getDatabase(url.localURI[1 .. $].to!long);
+
+	if (!url.queryString.empty)
+	{
+		import vibe.inet.webform : FormFields, parseURLEncodedForm;
+		auto query = FormFields.init;
+		parseURLEncodedForm(url.queryString, query);
+		foreach (param, val; query.byKeyValue)
+		{
+			switch (param)
+			{
+				/**
+				The password to use for the Redis AUTH command comes from either the
+				password portion of the "userinfo" URI field or the value from the
+				key-value pair from the "query" URI field with the key "password".
+				If both the password portion of the "userinfo" URI field and a
+				"query" URI field key-value pair with the key "password" are present,
+				the semantics for what password to use for authentication are not
+				well-defined.  Such situations therefore ought to be avoided.
+				*/
+				case "password":
+					if (!url.password.empty)
+						cli.auth(val);
+					break;
+				default:
+					throw new Exception(`Redis config parameter "` ~ param ~ `" isn't supported`);
+			}
+		}
+	}
+
+	/*
+	Redis' current optional authentication mechanism does not employ a
+	username, but this might change in the future
+	*/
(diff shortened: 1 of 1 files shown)


redis/vibe/db/redis/redis.d
```diff
@@ -33,19 +33,100 @@ import std.utf;
...
+
+	long databaseIndex;
+	if (url.localURI.length >= 2)
+		databaseIndex = url.pathString[1 .. $].to!long;
+
+	return cli.getDatabase(databaseIndex);
+}
+
+/// ditto
+RedisDatabase connectRedisDB(string host_or_url)
```
wilzbach: This is inspired by the way mongo handles things, hence the weird underscore parameter name.
I would prefer sht. like `hostOrURL` though ;-)

redis/vibe/db/redis/redis.d
```diff
@@ -33,19 +33,100 @@ import std.utf;
...
+						cli.auth(val);
+					break;
+				/**
+				The semantics of "query" URI field key-value pairs other than those
+				previously mentioned are implementation-defined.
+				However, it's a sensible default to allow them to be used as convenient
+				way to set configuration parameters.
+				*/
+				default:
+					cli.setConfig(param, val);
```
wilzbach: This might be controversial and I don't have a use case for it atm, I just thought it might be a nice feature, but I am happy to remove this bit.

s-ludwig:
> `"https://google.com".url`

I've got to say that I don't like these kinds of mini optimizations a lot. `URL("https://google.com")` is just a single character more to type and you can even do `"https://google.com".URL` if you want. Keeping the API slim and readable is IMO in most cases more important than saving a tiny bit of typing.

redis/vibe/db/redis/redis.d
```diff
@@ -33,19 +33,100 @@ import std.utf;
...
+}
+
+/// ditto
+RedisDatabase connectRedisDB(string host_or_url)
+{
+	/* If this looks like a URL try to parse it that way. */
+	if(host_or_url.startsWith("redis://")){
+		return connectRedisDB(URL(host_or_url));
+	} else {
+		return connectRedis(host_or_url).getDatabase(0);
```
s-ludwig: Not sure if it's a good idea to imply database 0 here. I'd normally say that a choice like this should be explicit.

redis/vibe/db/redis/redis.d
s-ludwig: Is there precedence for other implementations for this? It does bring the risk of clashing with other official options besides "password" that might be added in the future.

wilzbach:
> Is there precedence for other implementations for this? It does bring the risk of clashing with other official options besides "password" that might be added in the future.

No as mentioned it was just an idea -> replaced it with throwing an Exception about it being unsupported.

redis/vibe/db/redis/redis.d
wilzbach: As mentioned this is inspired by the way the Mongo adapter works and as `connectRedisDB`, returns a `RedisDatabase`, it has to be returned here as well.

wilzbach:
Okay I moved the controversial convenience overloads to a new PR (https://github.com/rejectedsoftware/vibe.d/pull/1852), s.t. nothing should be blocking this PR.

wilzbach:
@s-ludwig this still has the "Needs input" label, but AFAICT this PR has all comments addressed. Anything I am missing?

wilzbach:
Note that this would fix https://github.com/vibe-d/vibe.d/issues/1896

wilzbach:
Ping @s-ludwig I believe this got lost in the queue and nothing is blocking it.

s-ludwig:
True! Merging.

PR #1843, opened 2017-07-15
Fix #1785 - Use SCRAM-SHA-1 as default for MongoDB

### LINKED ISSUE #1785: Use SCRAM-SHA-1 as default for MongoDB
> MongoDB no longer defaults to MONGODB-CR and instead uses SCRAM-SHA-1 as the default authentication mechanism.

https://docs.mongodb.com/manual/core/security-mongodb-cr/#authentication-mongodb-cr
Released in March 2015
https://docs.mongodb.com/manual/release-notes/3.0

Either SCRAM-SHA-1 should be the default option or the initialization should detect the MongoDB version and thus set the auth mechanism automatically if Mongo > 3

wilzbach:
Added the fallback just to be sure to avoid breakage.

### DIFF
--- a/mongodb/vibe/db/mongo/connection.d
+++ b/mongodb/vibe/db/mongo/connection.d
@@ -184,10 +184,20 @@ final class MongoConnection {
 		m_bytesRead = 0;
 		if(m_settings.digest != string.init)
 		{
-			if (m_settings.authMechanism == MongoAuthMechanism.scramSHA1)
-				scramAuthenticate();
-			else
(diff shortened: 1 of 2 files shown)


wilzbach:
image

Hmm should we disable CodeCov status notifications until we get the coverage reporting fixed?
Uncovered bits will stay be highlighted via the browser extension ..

Lines in question would be here: https://github.com/rejectedsoftware/vibe.d/blob/master/.codecov.yml#L8 (setting them from `true` to `false`)

PR #1852, opened 2017-07-18
Provide a convenience string overload for connectRedisDB

wilzbach:
Split-off from https://github.com/rejectedsoftware/vibe.d/pull/1842


redis/vibe/db/redis/redis.d
```diff
@@ -33,6 +33,7 @@ import std.utf;
 */
 RedisClient connectRedis(string host, ushort port = RedisClient.defaultPort)
 {
+	assert(!host.startsWith("redis://", "Host must be a valid IP or domain, but not an URI."));
```
wilzbach: We could also think about parsing the `host` URL if the user accidentally called the wrong function, but imho it's better to use an error message here, because if the user has a full redis URI, he should have called `connectRedisDB` which returns a `RedisDatabase` and not only a `RedisClient`

wilzbach:
Added the error message to the `assert`. Anything else blocking this?

PR #1853, opened 2017-07-18
vibe.web.web: Allow Json as a auto-injected parameter type

wilzbach:
Split-off from https://github.com/rejectedsoftware/vibe.d/pull/1697

Now that `req.json` is lazy, this shouldn't be blocked anymore :)


s-ludwig:
Instead of giving `Json` such a special role, I'd like to go the same route as the REST interface. Having a parameter consume the full query/body would then work with an explicit `@queryParam`/`@bodyParam` attribute:
```
@bodyParam("bar", "b-ar") // searches for form field "b-ar"
void getFoo(string bar) {}

@bodyParam("bar") // the whole body will be stored in bar
void getFoo2(string bar) {}
```
That would also mean that `@headerParam` becomes available, which may be handy in some situations.

When that works, `Json` just has to be supported in some way in `readFormParamRec`.

This is of course a bit more to type and the question is if frequency wins over clarity of intent here. If it should turn out that it does, then we should at least restrict this to parameters named `_body`/`_query` or similar.

wilzbach:
> I'd like to go the same route as the REST interface. Having a parameter consume the full query/body would then work with an explicit @queryParam/@bodyParam attribute

As you might have guessed, I am not a huge fan of `@bodyParam`, because I don't have any need to handle old-school forms (it's 2017 after all). All my request PUT/POST requests originate from JS (or other) and thus I receive a rather large payload. Thus I hugely prefer to document the `InputType` explicitly:

```d
struct MyInputObject {
  string p1, p2, p3;
  @optional Nullable!int i;
}
void foo(MyInputObject input)
```

over picking out some attributes manually:

```d
@bodyParam("p1")
@bodyParam("p2")
@bodyParam("p3")
@bodyParam("i")
void (string p1, string p2, string p3, Nullable!int i = Nullable!int.init)
```

In practice there are much more attributes, of course.
Regarding `@queryParam`, while I agree that it's nice to have them documented as part of the function header, in practice they are almost always optional for me, which usually leads to nice code like this:

```d
void foo() {
if (auto tok = "adminToken" in req.query) {
...
```

Whereas with `@queyParam` this would be rather redundant:

```
@queryParam("adminToken")
void foo(Nullable!string adminToken = Nullable!string.init) {
if (!adminToken.isNull) {
...
```

> That would also mean that @headerParam becomes available, which may be handy in some situations.

Btw currently `headers(<key>, <value>)` only allows setting the response header, but `headers(<key>)` could be added.

> This is of course a bit more to type and the question is if frequency wins over clarity of intent here.

Well, I maintain a small fork of `vibe.web.web` ([renamed to hb-web](https://github.com/teamhackback/hb-web) to avoid conflicts) which contains https://github.com/rejectedsoftware/vibe.d/pull/1697
(It's intended as an experiment and I try to port successful/useful additions like this one to upstream).

> then we should at least restrict this to parameters named _body/_query or similar.

Would be fine for me.

wilzbach:
> Btw we could also - similarly to header, status. render, redirect, terminateSession etc. add a json helper function that can be used to access and write Json payloads.

-> https://github.com/vibe-d/vibe.d/pull/1945

wilzbach:
Rebased to trigger the CI systems.

> then we should at least restrict this to parameters named _body/_query or similar.

This is the case since some time. @s-ludwig anything else blocking this?
BTW (I repeat the question from above): is there any chance json could be part of the whitelist as well?

PR #1854, opened 2017-07-18
vibe.web.web: Allow to return strings

wilzbach:
Split-off from #1697

### DIFF
--- a/web/vibe/web/web.d
+++ b/web/vibe/web/web.d
@@ -1032,7 +1032,7 @@ private void handleRequest(string M, alias overload, C, ERROR...)(HTTPServerRequ
 			} else static if (is(RET : const(char)[])) {
 				res.writeBody(ret);
 			} else {
-				static assert(is(RET == void), M~": Only `InputStream`, `const(ubyte[])`, `Json`, const(char)[]` and `void` are supported as return types for route methods.");
+				static assert(is(RET == void), M~": Only `InputStream`, `const(ubyte[])`, `Json`, `const(char)[]` and `void` are supported as return types for route methods.");
 			}
 		}
 	} catch (Exception ex) {
--- a/web/vibe/web/web.d
+++ b/web/vibe/web/web.d
@@ -138,6 +138,15 @@ import std.encoding : sanitize;
 		case it will be used as an additional prefix to the one in
 		`WebInterfaceSettings.urlPrefix`.
 
+	Supported return types:
+		$(UL
+			$(LI $(D vibe.data.json.Json))
+			$(LI $(D const(char)[]))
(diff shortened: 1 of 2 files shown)


web/vibe/web/web.d
```diff
@@ -964,6 +964,8 @@ private void handleRequest(string M, alias overload, C, ERROR...)(HTTPServerRequ
 				} else {
 					res.writeBody(ret);
 				}
+			} else static if (is(RET : string)) {
+				res.writeBody(ret);
 			} else {
 				static assert(is(RET == void), M~": Only InputStream, Json and void are supported as return types for route methods.");
```
WebFreak001: Don't forget to include string in the error message

wilzbach:
> Don't forget to include string in the error message

Thanks. Added!

wilzbach:
@s-ludwig anything blocking this? Returning a `string` happens often enough, s.t. imho it justifies the convenience and also this is just an added bonus as a couple of return types (`InputStream` and `Json` are already accepted).

wilzbach:
Ping @s-ludwig - I think this is a trivial, non-controversial addition and helps a lot of people as returning `string` is a frequent use case.

WebFreak001:
@s-ludwig merge pls, this is really useful and further changes can be done afterwards

web/vibe/web/web.d
```diff
@@ -1020,8 +1029,10 @@ private void handleRequest(string M, alias overload, C, ERROR...)(HTTPServerRequ
 				} else {
 					res.writeBody(ret);
 				}
+			} else static if (is(RET : string)) {
```
s-ludwig: `const(char)[]` is sufficient here.

web/vibe/web/web.d
```diff
@@ -1020,8 +1029,10 @@ private void handleRequest(string M, alias overload, C, ERROR...)(HTTPServerRequ
...
+				res.writeBody(ret);
 			} else {
-				static assert(is(RET == void), M~": Only InputStream, Json and void are supported as return types for route methods.");
+				static assert(is(RET == void), M~": Only InputStream, Json, string and void are supported as return types for route methods.");
```
s-ludwig: Would be nice to add `const(ubyte)[]` here, too (and change to `const(char)[]` if adjusted).

s-ludwig:
Couldn't find a compelling argument against this, especially considering that `ubyte[]` and `char[]` are already supported by `writeBody` in a similar way. Would just be nice to loosen this up to `const(char)[]` instead of `string`.

wilzbach:
> Would just be nice to loosen this up to const(char)[] instead of string.

Good idea. Done.

PR #1856, opened 2017-07-18
Use abort() for fatal exit and avoid catching Throwable unless really necessary.

s-ludwig:
This avoids "dwarfeh(224) fatal error" that happens when exit() is called while an Error in in flight.

### DIFF
--- a/core/vibe/core/core.d
+++ b/core/vibe/core/core.d
@@ -1314,14 +1314,10 @@ private class CoreTask : TaskFiber {
 
 				s_availableFibers.put(this);
 			}
-		} catch(UncaughtException th) {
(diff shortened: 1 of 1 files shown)


wilzbach:
> Server returned nothing (no headers, no data) on handle 24EC680
> The command "./travis-ci.sh" exited with 2.

DUB failure again :/
I really hope that https://github.com/dlang/dub/pull/1190, https://github.com/dlang/dub/pull/1198 and https://github.com/dlang/dub-registry/pull/231 improve the status quo.

@s-ludwig btw I would be happy to restart Travis when I happen to browse through.

PR #1861, opened 2017-07-19
fix check for redefining symbol SO_REUSEPORT

rainers:
SO_REUSEPORT is defined in core.sys.posix.sys.socket by https://github.com/dlang/druntime/pull/1879, producing an ambiguity error.

Without an import, what symbol does the existing version check against?

### DIFF
--- a/core/vibe/core/drivers/utils.d
+++ b/core/vibe/core/drivers/utils.d
@@ -44,6 +44,7 @@ version (Windows) {
 } else alias SystemSocketException = ErrnoException;
 
 version (linux) {
+	import core.sys.posix.sys.socket;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Hm, looks like that was factored out into a separate module at some point, without also adding the matching imports.

schveiguy:
FYI, this still is broken in the 0.7.x branch. Can we get this there too? I can do the PR if you do the release to code.dlang.org. I'm having the issue now that the symbol was permanently added to 2.079 (I just tried to update my compiler).

I really should update to 0.8, but I'm not sure how big of a process it would be. Will test it out.

schveiguy:
PR created.

PR #1865, opened 2017-07-20
Bump DMD to 2.075.0

wilzbach:


s-ludwig:
Didn't you intend to remove the oldest versions (2.070 instead of 2.071 and ldc 1.0.0 instead of 1.1.0)?

wilzbach:
> Didn't you intend to remove the oldest versions (2.070 instead of 2.071 and ldc 1.0.0 instead of 1.1.0)?

No that was on purpose to save Travis build time. I kept the oldest version to check for compatibility as there was no reason not to keep compatibility. They can always be removed if new compiler / Phobos features need to used ;-)

s-ludwig:
Okay, I decided to drop 2.070.x support anyway due to linker errors in https://github.com/vibe-d/vibe-core/pull/27, so it makes no sense anymore to support it here either. 5 compiler versions is still considerably more than the advertised 3 consecutive versions, so I guess it's okay to make that step.

BTW I didn't drop intermediate versions so far, because of the (unlikely, but not impossible) case that something breaks only in an intermediate version. It wouldn't be detected and if someone stumbles over it, the CI check would have to be added back anyway to make sure it doesn't creep back in after being fixed.

Issue #1866, opened 2017-07-20
router.match doesn't work with delegates anymore


wilzbach:
Worked fine in 0.7.31, but broke in v0.8.0:

```d
#!/usr/bin/env dub
/+ dub.sdl:
name "hello_vibed"
dependency "vibe-d" version="~>0.8.0"
+/
void main()
{
    import vibe.d;
    auto settings = new HTTPServerSettings;
    settings.port = 8080;
    auto router = new URLRouter;
    router.match(HTTPMethod.HEAD, "aa", (req, res) {
        res.writeBody("Hello, World: " ~ req.path);
    });
    listenHTTP(settings, router);
    runApplication();
}
```

Will prepare a PR soon.

s-ludwig:
Fixed by #1867.

PR #1867, opened 2017-07-20
Fix 1866 - router.match doesn't work with delegates anymore

wilzbach:
See also: https://github.com/rejectedsoftware/vibe.d/issues/1866

### DIFF
--- a/http/vibe/http/router.d
+++ b/http/vibe/http/router.d
@@ -133,6 +133,12 @@ final class URLRouter : HTTPServerRequestHandler {
 		return this;
 	}
 
+	/// ditto
+	URLRouter match(HTTPMethod method, string path, HTTPServerRequestDelegate handler)
+	{
+		return match!HTTPServerRequestDelegate(method, path, handler);
+	}
+
 	/// Adds a new route for GET requests matching the specified pattern.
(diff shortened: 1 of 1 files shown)


http/vibe/http/router.d
```diff
@@ -133,6 +133,12 @@ final class URLRouter : HTTPServerRequestHandler {
 		return this;
 	}
 
+	/// ditto
+	URLRouter match(HTTPMethod method, string path, HTTPServerRequestDelegate handler)
+	{
+		return match!HTTPServerRequestDelegate(method, path, handler);
+	}
```
thaven: Looks like the templated match function above this should already do the job just fine. So this seems like a work-around for a compiler bug to me. If so, the bug should be reported (if not known yet) and referenced in a comment here.

s-ludwig: The problem is that the compiler has nothing to infer the argument types of the inline function/delegate in an efficient way. In theory it could determine the common types of all call expressions within the function, but even that would be quite brittle, as the semantics of the template body can in turn depend on the exact delegate type in various ways. In the end it would probably be a combinatoric problem with unpredictable performance characteristics.

PR #1877, opened 2017-07-25
make tests agnostic to the mangling of string[string]

rainers:
This might help https://github.com/dlang/dmd/pull/6998 to pass the tests.

Is the mangling only used for the test in vibe.d or does a change also have an impact on real serialized data?

### DIFF
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -1432,7 +1432,8 @@ unittest { // basic serialization behavior
 	test(12.0f, "V(f)(12)");
 	assert(serialize!TestSerializer(null) ==  "null");
 	test(["hello", "world"], "A(AAya)[2][AE(Aya,0)(V(Aya)(hello))AE(Aya,0)AE(Aya,1)(V(Aya)(world))AE(Aya,1)]A(AAya)");
-	test(["hello": "world"], "D(HAyaAya){DE(Aya,hello)(V(Aya)(world))DE(Aya,hello)}D(HAyaAya)");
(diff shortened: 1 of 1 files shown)


s-ludwig:
It's just used for the generic test serializer and won't affect any real-world serialization.

rainers:
Thanks for merging. Could you also add a new tag for https://github.com/dlang/dmd/pull/6998 to pick up or will this happen anyway soon?

s-ludwig:
DUB releases are tied to DMD releases, but if the project tester also picks up pre-release versions, we could add an alpha version tag. But, @MartinNowak, for synchronized releases, wouldn't it makes sense to test against DUB stable anyway? Then we could just merge this into stable.

rainers:
As I understand @wilzbach in https://github.com/dlang/dmd/pull/6998#issuecomment-319215659 the jenkins test builds fetch the latest tag. But according to the build log, it still uses v0.8.0 (https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6998/7/pipeline/182#step-704-log-7).

The dub failure (https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6998/7/pipeline/173#step-481-log-286) seems to have crept in a couple of days ago. Not sure how it is related, though.

s-ludwig:
Sorry, I was in a hurry and confused the repository we are talking about. So, the next vibe.d release should actually be out soon. I've just tagged v0.8.1-beta.1 (also picked up by code.dlang.org) and will start the beta/RC cycle in the next days once I get a little more time.

PR #1891, opened 2017-08-14
use aliases instead of wrapper functions

MartinNowak:


s-ludwig:
Hm, this got lost between the other related PRs. If it doesn't introduce any issues, it looks like the better approach. Can you verify that the aliases correctly override the original declarations in the deimos headers? I'd hate to break this again just before release.

MartinNowak:
Yes, aliases introduce new symbols that won't overload with symbols in imports (no hijacking rule).
An imported symbol can be aliased into the current module as well, in which case it would participate in overload resolution.

s-ludwig:
Okay thanks, giving it a shot for the next release.

PR #1893, opened 2017-08-17
Added forward proxy support  and included an example

mattremmel:
Overview of changes:
- Generalized 'HTTPReverseProxySettings' into 'HTTPProxySettings' and added a bool for setting reverse proxy mode.
- Moved the 'handleRequest' code out of 'reverseProxyRequest' and into a general 'proxyRequest' function.
- Created a 'forwardProxyRequest' function and updated the 'reverseProxyRequest' function to then call into 'proxyRequest' with the properly set 'HTTPProxySettings' where necessary.
- Created new 'listenHTTPForwardProxy' methods to compliment the existing 'listenHTTPReverseProxy' functions.

Testing:
All unit tests pass. I tested both the existing 'http_reverse_proxy' example and the new 'http_forward_proxy' example to ensure they were both working. NOTE*: the 'http_forward_proxy' example will only work with HTTP sites, not HTTPS. Trying to access an HTTPS site will show error 'proxy is refusing connections'. Setting up the HTTP listener to use SSL will solve this, although that will drop support for regular HTTP sites.

Notes:
Ideally the example would be better off working with both HTTP and HTTPS, so it could truly be used like a general purpose forward proxy. This would require having a listener that can handle both normal and TLS encrypted TCP connections. I don't believe Vibe.d currently supports this (let me know if I'm wrong), but I have some ideas on how to add support for this.

If there are any desired changes or updates to this PR, feel free to let me know. I won't take any offense.

### DIFF
--- a/http/vibe/http/proxy.d
+++ b/http/vibe/http/proxy.d
@@ -25,44 +25,48 @@ import std.exception;
 */
 
 /**
+	Transparently forwards all requests to the proxy to another host.
+
+	The configurations set in 'settings' and 'proxy_settings' determines the exact
+	behavior.
+*/
+void listenHTTPProxy(HTTPServerSettings settings, HTTPProxySettings proxy_settings)
+{
+	// disable all advanced parsing in the server
+	settings.options = HTTPServerOption.None;
+	listenHTTP(settings, proxyRequest(proxy_settings));
+}
+// Compatibility alias - will be deprecated soon.
+alias listenHTTPReverseProxy = listenHTTPProxy;
+
+/**
 	Transparently forwards all requests to the proxy to a destination_host.
 
 	You can use the hostName field in the 'settings' to combine multiple internal HTTP servers
 	into one public web server with multiple virtual hosts.
 */
-void listenHTTPReverseProxy(HTTPServerSettings settings, HTTPProxySettings proxy_settings)
-{
-	// disable all advanced parsing in the server
-	settings.options = HTTPServerOption.None;
-	listenHTTP(settings, reverseProxyRequest(proxy_settings));
-}
-/// ditto
 void listenHTTPReverseProxy(HTTPServerSettings settings, string destination_host, ushort destination_port)
 {
 	URL url;
 	url.schema = "http";
 	url.host = destination_host;
 	url.port = destination_port;
-	auto proxy_settings = new HTTPProxySettings;
-	proxy_settings.reverseProxyMode = true;
+	auto proxy_settings = new HTTPProxySettings(ProxyMode.reverse);
 	proxy_settings.destination = url;
 	listenHTTPReverseProxy(settings, proxy_settings);
 }
 
-/**
-	Transparently forwards all requests to the proxy to the requestURL of the request.
-*/
+/** 
+	Transparently forwards all requests to the proxy to the requestURL of the request. 
+*/ 
 void listenHTTPForwardProxy(HTTPServerSettings settings) {
-	listenHTTP(settings, forwardProxyRequest(new HTTPProxySettings));
-}
-
-/// ditto
-void listenHTTPForwardProxy(HTTPServerSettings settings, HTTPProxySettings proxy_settings) {
-	listenHTTP(settings, forwardProxyRequest(proxy_settings));
+	auto proxy_settings = new HTTPProxySettings(ProxyMode.forward);
+	proxy_settings.handleConnectRequests = true;
+	listenHTTPProxy(settings, proxy_settings);
 }
 
 /**
-	Generic proxy delegate that will work with both forward and reverse proxies
+	Returns a HTTP request handler that forwards any request to the specified or requested host/port.
 */
 HTTPServerRequestDelegateS proxyRequest(HTTPProxySettings settings)
 {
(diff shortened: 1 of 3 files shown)


mattremmel:
I'm not sure what the best way to go about unit testing this is. It doesn't look like anything from the original proxy.d was being unit tested to begin with. I'm open to ideas.

http/vibe/http/proxy.d
```diff
@@ -226,6 +260,8 @@ final class HTTPReverseProxySettings {
 	/// ditto
 	@property void destinationPort(ushort port) { destination.port = port; }
 
+	/// Whether the proxy should act as forward or reverse proxy
+	bool reverseProxyMode = false;
```
s-ludwig: Again, to not break existing code that uses the first `listenHTTPReverseProxy` overload, the default would have to be `true` here.

http/vibe/http/proxy.d
```diff
@@ -185,38 +200,57 @@ HTTPServerRequestDelegateS reverseProxyRequest(HTTPReverseProxySettings settings
...
 		}
 	}
 
 	return &handleRequest;
 }
+
+/**
+	Returns a HTTP request handler that acts as a forward proxy
+*/
+HTTPServerRequestDelegateS forwardProxyRequest(HTTPProxySettings settings)
```
s-ludwig: I'd probably drop this, since the `settings.reverseProxyMode` becomes kind of redundant. In the long run, `reverseProxyRequest` could also be deprecated. Setting up proxies is also uncommon enough that  convenience functions don't really add much value in practice.

http/vibe/http/proxy.d
s-ludwig: It would also be nice to convert this into a `enum ProxyMode { forward, reverse }` and add a constructor that takes a mode, so that forward/reverse becomes an explicit decision. A default constructor would be needed to avoid breaking existing code, but that can be deprecated/removed later down the road.

s-ludwig:
> I'm not sure what the best way to go about unit testing this is. It doesn't look like anything from the original proxy.d was being unit tested to begin with. I'm open to ideas.

I was under the impression that there was a high level test in the "tests" directory, but obviously it was just the TCP proxy test that I had in mind. I would approach this in a similar way, setting up one normal server, one forward/reverse proxy and then firing requests at the proxy and the normal server to compare the results (that would also test the proxy support in the HTTP client). Would be great to have a test, but since none existed up to now, I wouldn't see that as mandatory for this PR.

s-ludwig:
@mattremmel: Do you have the time to make these adjustments? If not, I could also adopt the branch and add the changes on top, as this opportunity for generalization IMO should definitely be taken.

mattremmel:
@s-ludwig Hey, sorry. I tabled this when work got busy and never circled back around. I can finish these changes either today or tomorrow. I'll update the PR when I am done. Thanks.

mattremmel:
So I took another pass at this and finished the changes you requested I believe. I've never worked on a library before, only applications, so the API backwards compatibility shamefully never crossed my mind, but that should all be taken care of now.

I liked the ProxyMode idea, so I've implemented that. I also generalized two of the core functions but added compatibility aliases for backwards compatibility, so let me know what you think. To me, it makes more sense now that a forward proxy mode is an option.

I'm working on writing some unit tests, and I also want to do more on supporting HTTPS, and unit tests will probably help speed up and verify that. I'm not an expert by any means, but typically proxies like Charles or Burp Suite run on a single port (8080) and are capable of forwarding both HTTP and HTTPS traffic through the same port simultaneously. I don't believe this is the same as HTTP CONNECT, although that can be achieved via HTTP CONNECT. However, in my tests, enabling HTTP CONNECT did not solve that problem because my browser was trying to establish a HTTPS connection with the proxy which was failing because it was just an HTTP listener without TLS enabled. My understanding is that proxies like Charles or Burp Suite, will effectively run something like a multiplexer listener, that will detect whether a TLS connection is attempting to be established, and then dispatch the connection to either an SSL/TLS handler or just normal HTTP handler.

Firefox and most other browsers only allow configuring a single HTTP proxy to use, and don't allow specifying a proxy for TLS and a proxy for Non-TLS traffic. Perhaps I misunderstood your suggestion about calling `listenHTTP` twice, but that would only work if both instances could listen on the same port without interference. Maybe they can?

s-ludwig:
> Perhaps I misunderstood your suggestion about calling listenHTTP twice, but that would only work if both instances could listen on the same port without interference. Maybe they can?

No you are right, that won't work on the same port without additional effort currently. I didn't check the traffic for proxy requests for quite a while, so maybe things have indeed changed. However, the question would be how a HTTPS request that simply goes through the proxy would specify its destination to the proxy and at the same time would remain securely encrypted end-to-end. What the mentioned proxies usually have to do is to generate a dummy certificate for the HTTPS site and basically act as a man in the middle. I'd still assume that they simply intercept CONNECT requests that way, but I'll double check that.

The changes look good now. The CI tests just currently fail due to the three trailing spaces in lines 59 to 61 and due to an unrelated error obviously caused by some changes in the Travis-CI setup. From my side we could merge this as a first step and add tackle the HTTPS and testing issues separately.

s-ludwig:
Okay, so I checked with Wireshark on Windows and at least Firefox still performs a CONNECT if I set a proxy with the same port for all protocols. When this is merged I can have another look at where it may have gone wrong there.

s-ludwig:
Turned out that CONNECT handling was simply broken. See #1973 for the fix.

PR #1894, opened 2017-08-17
Make Meson work with 0.8.x

ximion:
The Meson configuration is modularized as well now - there are some improvements that can be done (the files are a bit verbose), and maybe there are even some issues left, but I will find most of those when packaging the new version for Debian.

I also enabled CI for the builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ __dummy.html
 .dub
 dub.selections.json
 
+# Auto-downloaded 3rd-party
+lib/subprojects/
+
 # Mono-D files
 *.userprefs
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,17 @@ matrix:
     - d: ldc-1.3.0
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
+before_install:
+  - pip3 install meson>=0.40
+
+install:
+  - mkdir .ntmp
+  - curl -L https://github.com/ninja-build/ninja/releases/download/v1.7.2/ninja-linux.zip -o .ntmp/ninja-linux.zip
+  - unzip .ntmp/ninja-linux.zip -d .ntmp
+
+before_script:
+  - export PATH=$PATH:$PWD/.ntmp
+
 script: ./travis-ci.sh
 after_success:
  - bash <(curl -s https://codecov.io/bash)
--- a/core/meson.build
+++ b/core/meson.build
@@ -1,0 +1,86 @@
+# Meson file for Vibe Core
+
+vibe_core_src_dir = include_directories('.')
+
+vibe_core_src = [
+    'vibe/appmain.d',
+    'vibe/core/args.d',
+    'vibe/core/concurrency.d',
+    'vibe/core/connectionpool.d',
+    'vibe/core/core.d',
+    'vibe/core/driver.d',
+    'vibe/core/drivers/libasync.d',
+    'vibe/core/drivers/libevent2.d',
+    'vibe/core/drivers/libevent2_tcp.d',
+    'vibe/core/drivers/native.d',
+    'vibe/core/drivers/threadedfile.d',
+    'vibe/core/drivers/timerqueue.d',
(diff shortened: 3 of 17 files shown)


ximion:
With LDC 1.3.0 I get
```
9495 Illegal instruction     (core dumped) ./core/vibe-test_core
```
This seems to be an LDC bug...

ximion:
@s-ludwig You would need to take a look at the test_version script for this to work. Library versioning on Linux doesn't allow letters in the version, so a build will fail if you attempted that. I also set a project-wide version now, which has the same restrictions.
If we want to use the full version in Meson somewhere (we currently don't), we compose if from the given version and a suffix, that can be set in the main file as well.

I don't know what you like here, so I just disabled this for now to let the CI continue its job.
It also appears we hit a couple of LDC bugs here, which don't appear in DMD... I guess those will have to be resolved as well (see the linked bug report above).
I would really like to have this in the next release, so I can get the Debian package into the testing suite properly.

s-ludwig:
So if we would set the first three digits of the full version as the meson project version and set the rest as the suffix, would that work correctly for upgrades (i.e. would it detect that 1.0.0 + "~beta.1" is a lower version than "~rc.1", or can we achieve that with other means)?

I wouldn't mind if the LDC+libevent meson tests were just skipped for now, until the crash gets fixed in LDC. The future direction will be the vibe-core/eventcore configuration anyway, so it wouldn't be a big loss at this point.

ximion:
@s-ludwig The tilde (~) lowers the version number, so `0.1.2 >> 0.1.2~rc1`. This is true in Debian and AFAIR RPM-based distros as well, since the dash is reserved for the distribution revision. We actually rewrite the upstream version name to that format prior to comparing versions, as `0.1.2 << 0.1.2-rc1`, so we get notified properly when a new upstream release is there.
I think I used this in the Meson file just out of habit.

> I wouldn't mind if the LDC+libevent meson tests were just skipped for now, until the crash gets fixed in LDC. The future direction will be the vibe-core/eventcore configuration anyway, so it wouldn't be a big loss at this point.

I can skip them - but does that mean I should switch the Meson configuration to use vibe-core/eventcore? I would like to have it use the recommended codepath for most projects, and the libevent one seemed to be that for a while.

I could also cheat and have the libevent+LDC tests not run the tests until the bug is fixed...

s-ludwig:
> I can skip them - but does that mean I should switch the Meson configuration to use vibe-core/eventcore? I would like to have it use the recommended codepath for most projects, and the libevent one seemed to be that for a while.
>
> I could also cheat and have the libevent+LDC CI job not run the tests until the bug is fixed...

Oh, no, I misinterpreted the results as the Meson build running on both, libevent and vibe-core configurations (didn't remember the constraint in travis.sh). My comment doesn't really make sense then, as the tests may likely fail to build with vibe-core, too.

Just not running the tests would be the cheat that I had in mind then.

ximion:
> Just not running the tests would be the cheat that I had in mind then.

Applied :-) But long-term it looks like I should support vibe-core in Meson as well as libevent then :)

ximion:
On DMD 2.072.2 the `vibe-test_stream` fails:
```
5/13 vibe-test_stream                        FAIL     0.29 s
--- command ---
/home/travis/build/ximion/vibe.d/build/stream/vibe-test_stream
--- stderr ---
/home/travis/build/ximion/vibe.d/build/stream/vibe-test_stream(_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPS4core3sys5posix6signal9siginfo_tPvZv+0x38)[0x5618ac]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x10330)[0x7ff0cd264330]
-------
```
All other DMD versions are fine. Smells like another compiler bug to me, I will make Travis skip the tests on that particular DMD version as well.

ximion:
And now the CI is happy :-)

s-ludwig:
Okay, I'll merge and play around with the version test/tag scripts!

ximion:
Thank you! :-)

Issue #1898, opened 2017-08-22
Projects fail to build without Diet-NG


ximion:
Hi!
If I compile a Vibe project without Diet-NG, I get errors like:
```
/usr/include/d/vibe/vibe/http/server.d(279): Error: undefined identifier 'DefaultFilters'
```
This, looking at the code, makes sense. So I guess you might want to make Diet a hard-dependency, fail with a better error message, or maybe fix this problem somehow, so builds without Diet still work.
Thanks!

s-ludwig:
Hi! Diet-NG has indeed become a hard dependency as of 0.8.1, as the old implementation has been removed completely: https://github.com/rejectedsoftware/vibe.d/blob/c28ce62e33a6d4d9dea827d9f3d13c759ae79fe0/http/dub.sdl#L7
I missed how this is now handled in the meson build when looking through the PR. But tagged diet-ng 1.4.1 has been tagged with the meson.build included, so it could be added to vibe.d.

ximion:
@s-ludwig I added a patch which resolves that in Meson (requires a minor change in Diet as well, for convenience).
I guess it might make sense then in Vibe.d to remove the version block around the Diet-NG code? (so any attempt to compile without Diet will fail for real).

PR #1899, opened 2017-08-22
meson: Always build with Diet

ximion:
Hi!
This makes the Meson build-system automatically fetch the latest Git copy of Diet in case no Diet is found on the system already. This resolves issue #1898

Please do not merge this until https://github.com/rejectedsoftware/diet-ng/pull/39 is merged and I have adjusted the Git URL in this patch to point to upstream Diet instead of my forked copy!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,9 @@ __dummy.html
 dub.selections.json
 
 # Auto-downloaded 3rd-party
-lib/subprojects/
+lib/subprojects/diet/
+lib/subprojects/libevent/
+lib/subprojects/openssl/
 
 # Mono-D files
 *.userprefs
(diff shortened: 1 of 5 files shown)


ximion:
Okay, the URL is updated now, so this is merge-ready :-)
For some more information: This patch uses Meson's Wrap system to have Meson pull in a missing dependency automatically, in case it is missing on the system. The location of the 3rd-party project that is embedded is defined in a wrapfile, in this particular case `lib/subprojects/diet.wrap`.

I've set up the file to always pull the latest Git `master` copy for Diet. If you want a specific tag, or download a tarball instead of checking out the code from Git, you can set these things in the Wrapfile. In any case, the file will always point to a specific version of a dependency and has to be updated in order to switch to a newer version, that's why I just had it follow the master branch for simplicity.

More information about this can be found in Meson's documentation: http://mesonbuild.com/Wrap-dependency-system-manual.html

PR #1901, opened 2017-08-23
Fix it so that std.c.* is no longer required.

jmdavis:
std.c.* has been deprecated for a while and is about to be removed from
Phobos. The bindings that vibe.d was using that were there and still
missing from druntime have been added, and vibe.d needs to be updated to
use them.

I would have just continued to import `std.c.*.socket` for older versions, but unfortunately, at the moment, it looks like the way that `__VERSION__` is handled in dmd master is that it matches the last major release and not the release that it will be for. The druntime bindings did not quite manage to make it into 2.076 unfortunately, and the `std.c.*` modules were actually scheduled to be removed back in June. So, they probably won't be in master much longer. So, using them as long as `__VERSION__` is less than 2077 will fail thanks to the fact that `__VERSION__` is 2076 for master instead of 2077. So, it made more sense to me to just add the declarations here for now, and once vibe.d no longer supports any versions older than 2.077, then those `static if` blocks can be removed.

Either way, vibe.d needs to stop importing from deprecated modules that are about to be removed. Presumably, it's still been using them, because the necessary bindings weren't in druntime like they were supposed to be, but they are now.

### DIFF
--- a/core/vibe/core/drivers/libevent2.d
+++ b/core/vibe/core/drivers/libevent2.d
@@ -61,10 +61,31 @@ version(Windows)
 
 version(OSX)
 {
-	import std.c.osx.socket : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
-} else version(Posix)
+	static if (__VERSION__ < 2077)
+	{
+		enum IP_ADD_MEMBERSHIP = 12;
+		enum IP_MULTICAST_LOOP = 11;
+	}
+	else
+		import core.sys.darwin.netinet.in_ : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
+} else version(FreeBSD)
 {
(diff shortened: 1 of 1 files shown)


jmdavis:
Oh, and a side effect of this is that this code will now compile on FreeBSD (even for older compiler versions). It didn't before, because `std.c.*.socket` didn't have bindings for FreeBSD.

jmdavis:
It looks like the build failures are unrelated.

    Failed to download http://code.dlang.org/packages/libevent/2.0.2+2.0.16.zip: 500 Internal Server Error

s-ludwig:
It's good to finally have those definitions officially available again. While I certainly don't mind having such definitions in the code base for a while, this is an example of an occasional bad habit in Phobos that should really be avoided (removing/deprecating things before there is a replacement ready).

PR #1906, opened 2017-08-30
improve json range support

John-Colvin:
A very simple translation of `skipNumber` to the range API.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1192,7 +1192,7 @@ Json parseJson(R)(ref R range, int* line = null, string filename = null)
 			if( is_float ) {
 				ret = to!double(num);
 			} else if (is_long_overflow) {
-				ret = () @trusted { return BigInt(num); } ();
+				ret = () @trusted { return BigInt(num.to!string); } ();
 			} else {
 				ret = to!long(num);
 			}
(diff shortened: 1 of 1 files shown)


UplinkCoder:
@John-Colvin please profile the performance consequences of this change. 
It seems like it'd hurt quite a bit. please profile with `-Os` as well

s-ludwig:
One small performance improvement would be to skip incrementing `idx` for ranges that define `length` and use `takeExactly(sOrig.length - s.length)` instead. But benchmarking is a topic that I need to get back to anyway. I did a quick test at DConf this year and performance was really bad, whereas I remember pretty good numbers from the last tests.

John-Colvin:
The major performance problem is now fixed by operating on `ubyte[]` instead of auto-decoded strings (Oh how I hate them).

In debug mode `assumeUTF` verifies utf, which is of course expensive. In release, that check doesn't happen.

s-ludwig:
Thanks, John! The performance implication of the additional index is supposedly small enough.

The more important question IMO is anyway to see what the future of `std_data_json` is. There were some points brought up that made it difficult to continue the Phobos integration route, so it might make sense to establish it as an independent library and let it slowly replace this module.

PR #1916, opened 2017-09-07
setCommandLineArgs

burner:
This allows to handle command line args before the vibe.d cmd hanlder kicks in.
This is needed if a user for example wants to handle --help|-h by himself or wants to use a custom parser.


burner:
@s-ludwig thanks

Issue #1922, opened 2017-09-10
vibe.web.rest.Collection and vibe.web.auth


grogancolin:
👋 

Sorry for the large chunk of code I'm gonna paste here, but it's the smallest I could've gotten it while keeping useful info.

Basically, I have an API that is behind a `@requiresAuth`. and inside that API is a sub API that I'd like to also require authentication, but also use `vibe.web.rest.Collection` on.

However, the sub API fails to receive the `auth` parameter from the `authenticate` call.
Also, as I have a log message in the `authenticate` mixin, I can see that the Sub APIs `authenticate` method is being called at the expected time.

These are the commands I used to test:
```bash
# check if I can make an authenticated request
$ curl -b cookies -XGET http://localhost:8080/test\?num=3 
403 - Forbidden

Not authorized to perform this request

# Login
$ curl -c cookies -XPOST -H "Content-Type: application/json" -d '{"user":"admin","pass":"supersecret"}' localhost:8080/login
true

# check if I can make an authenticated request
$ curl -b cookies -XGET http://localhost:8080/test\?num=3
"admin3"

$ curl -b cookies -XGET http://localhost:8080/items/name\?item=this
400 - Bad Request

Missing non-optional query parameter 'auth'.
```

And now for the code. It's basically an amalgamation of:
http://vibed.org/api/vibe.web.rest/Collection
http://vibed.org/api/vibe.web.auth/
```D
import vibe.vibe;
import vibe.web.auth;
import vibe.web.rest;

void main()
{
    auto settings = new HTTPServerSettings;
    settings.port = 8080;
    //settings.sessionStore = new RedisSessionStore("127.0.0.1", 0);
    settings.sessionStore = new MemorySessionStore();

    auto router = new URLRouter;
    router.registerRestInterface(new API());

    listenHTTP(settings, router);
    runApplication();
}

struct AuthInfo{
    string name;
}

/// The Item API that's using vibe.web.rest.Collection
@requiresAuth
interface I_ItemAPI {
    struct CollectionIndices {
        string item;
    }

    @anyAuth
    string getName(string item, AuthInfo auth);

    mixin CreateAuthFunc;
}

class ItemAPI : I_ItemAPI {
    /+
    // I'd expect 'auth' to be populated with the session, but no such luck
       $ curl -b cookies -XGET http://localhost:8080/items/name\?item=this
       400 - Bad Request

        Missing non-optional query parameter 'auth'.
    +/
[...]

PR #1924, opened 2017-09-12
Added test for #1922

grogancolin:
This adds a (currently failing) test for #1922. 
It fails on the call to: 
 `test("/items/name?item=something", "admin");`


s-ludwig:
I'll merge this once #1939 is on master. In the meantime, master has been fixed so that the tests actually run again (this got silently broken during 0.8.1 release time). It was quite fortunate that this PR was actually supposed to fail, but the CI actually passed.

s-ludwig:
Added to master: 75a46f77523417e07c52eb2c3b83d093441c218f

I just had to adjust the code a bit so that the `@requiresAuth` annotation is on the collection interface: bb3fd89ce17544705779a3ab8d8b94abe71ce6de

Thanks for doing the PR!

PR #1927, opened 2017-09-13
travis: Workaround for pip install Travis issue

ximion:
@s-ludwig:
> Travis builds start to fail, because of something that changed with their system packaging: https://travis-ci.org/vibe-d/vibe.d/jobs/274722352
>
> Do you know of a more portable way to access pip/pip3?

Annoying... They should just make a base image based on a current Ubuntu release (like, the last LTS, not a prehistoric one) test it until they like it and then throw it out without modifying much (new browsers and security/integration fixes etc. are exceptions).
They warned about the image changing, but this particular change is missing from the changelog :P https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/

Anyway, I was playing around with this a little, but Travis seems to be on low capacity at time, so I am not sure if this patch works (we'll see).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ matrix:
       env: VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0
 
 before_install:
+  - pyenv global system 3.5
   - pip3 install meson>=0.40
 
 install:


ximion:
Okay, it's broken - I think Travis actually made a mistake here, since this will break quite a lot of stuff using Python and Travis.
I will look into this later when I'm back home, I have a few more ideas on how to fix this or work around the issue.

wilzbach:
If you need a newer pip version, you could have a look at using miniconda

ximion:
So, the problem is actually this:
```
/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.

  SNIMissingWarning
```
This is a Travis bug, reported as https://github.com/travis-ci/travis-ci/issues/8378 and https://github.com/travis-ci/travis-ci/issues/8382
This pull-request now contains a workaround.

s-ludwig:
Okay, thanks a lot! BTW, looks like the DDOX test build also got broken (node.js dependency issues). I would have thought that they'd perform a bunch of test runs for existing repositories before deploying such updates.

PR #1931, opened 2017-09-17
Add methods to support client request digest authentication

tchaloupka:
Added the method to generate digest authentication request header from the response digest parameters.

Also added the example to test it.
It works for the start but I would like to solve also the problem with `cnonce` and `nc` parameters.

If I understand it correctly, cnonce should be the same if the server uses the previous nonce and nc just incremented in following client requests.
For that it is needed to store the values somewhere on the client.
I think that maybe if we add Variant context storage to the HTTPClient and make it accessible from the HTTPClientRequest and HTTPClientResponse it can be used for this as HTTPClient is reused from the pool. Also it should be cleared when the connection is closed.

Is this the correct way of thinking to provide facilities to implement digest authentication for the client calls properly?

### DIFF
--- a/http/vibe/http/auth/digest_auth.d
+++ b/http/vibe/http/auth/digest_auth.d
@@ -11,6 +11,7 @@ module vibe.http.auth.digest_auth;
 
 import vibe.http.server;
 import vibe.core.log;
+import vibe.inet.url;
 
 import std.base64;
 import std.datetime;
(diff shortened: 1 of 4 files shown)


http/vibe/http/auth/digest_auth.d
```diff
@@ -179,3 +180,132 @@ string createDigestPassword(string realm, string user, string password)
...
+	string realm, domain, nonce, opaque;
+	Algorithm algorithm = Algorithm.md5;
+	bool stale;
+	Qop qop;
+
+	/// Parses WWW-Authenticate header value with the digest parameters
+	this(string auth) {
+		import std.algorithm : splitter;
+
+		assert(auth.startsWith("Digest "), "Correct Digest authentication request not provided");
```
s-ludwig: This should probably be an enforcement, because the string comes from external sources.

http/vibe/http/auth/digest_auth.d
```diff
@@ -179,3 +180,132 @@ string createDigestPassword(string realm, string user, string password)
...
+	Params:
+		username = user name
+		password = user password
+		url = requested url
+		auth = value from the WWW-Authenticate response header
+		cnonce = client generated unique data string (required only when some qop is requested)
+		nc = the count of requests sent by the client (required only when some qop is requested)
+		entityBody = request entity body required only if qop==auth-int
+*/
+auto createDigestAuthHeader(U)(HTTPMethod method, U url, string username, string password, DigestAuthParams auth,
```
s-ludwig: I'd split this up into two methods:
```
void addDigestAuth(scope HTTPRequest req, ...);  // analogous to addBasicAuth()
void writeDigestAuthHeader(R)(ref R dst, ...); // dst is an output range that receives the contents
```
The latter can probably be left private for now (to keep the API small and avoid the need for deprecation processes if something ever needs to change).

http/vibe/http/auth/digest_auth.d
```diff
@@ -179,3 +180,132 @@ string createDigestPassword(string realm, string user, string password)
...
+	enum Qop { none = 0, auth = 1, auth_int = 2 }
+	enum Algorithm { none = 0, md5 = 1, md5_sess = 2 }
+
+	string realm, domain, nonce, opaque;
+	Algorithm algorithm = Algorithm.md5;
+	bool stale;
+	Qop qop;
+
+	/// Parses WWW-Authenticate header value with the digest parameters
+	this(string auth) {
```
s-ludwig: A static method `static DigestAuthParams parse(in char[] auth)` would arguably be better here to make it clearer that actual parsing work is going on, and it's not just a field that is initialized.

tchaloupka:
Sorry for the delay on this, I've made the required changes.
Let me know if more is needed, or if I should quash the PR before merge.

s-ludwig:
Thanks again! Looks good to merge now.

tchaloupka:
Failed due to timeout of the test :(

PR #1944, opened 2017-10-10
ignore druntime arguments

MartinNowak:
As vibe.d is using low-level Runtime.args from which the `--DRT-` arguments are not filtered.


WebFreak001:
@s-ludwig bump

mathias-lang-sociomantic:
I just hit this bug trying to pass `--DRT-oncycle=ignore` :/

PR #1946, opened 2017-10-10
Make HTTPServerResponse.path lazy

wilzbach:
I know that `req.path` is about to be deprecated and will disappear with the upcoming rewrite of the http module. However, as `parseURL` is often set by default, we can save quite a bit _today_ without doing any hard work.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -445,7 +445,7 @@ HTTPServerRequest createTestHTTPServerRequest(URL url, HTTPMethod method, InetHe
 @safe {
 	auto tls = url.schema == "https";
 	auto ret = new HTTPServerRequest(Clock.currTime(UTC()), url.port ? url.port : tls ? 443 : 80);
-	ret.path = urlDecode(url.pathString);
+	ret.requestPath = url.path;
 	ret.queryString = url.queryString;
 	ret.username = url.username;
 	ret.password = url.password;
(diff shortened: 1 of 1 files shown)


http/vibe/http/server.d
```diff
@@ -879,30 +879,29 @@ final class HTTPServerRequest : HTTPRequest {
...
+
+		private Nullable!string _path;
 
 		/** The path part of the requested URI.
 		*/
 		InetPath requestPath;
 
 		/** The user name part of the URL, if present.
-
-			Remarks: This field is only set if HTTPServerOption.parseURL is set.
```
wilzbach: At the moment they are always parsed ...

http/vibe/http/server.d
```diff
@@ -532,7 +532,7 @@ alias HTTPServerErrorPageHandler = void delegate(HTTPServerRequest req, HTTPServ
 */
 enum HTTPServerOption {
```
wilzbach: When this module is rewritten, it probably makes sense to rename / fully deprecate this enum as sadly AFAIK individual fields can't be deprecated at the moment

wilzbach: Oh we can hack around it: https://github.com/vibe-d/vibe.d/pull/1947

s-ludwig:
Makes sense, even if this usually won't have a positive effect until `URLRouter` gets fixed to use `requestPath` instead. I'll try to get that done for the next version.

http/vibe/http/server.d
s-ludwig: I really hope that this will be implemented one day, as this is far from being the only enum with deprecated fields in vibe.d.

PR #1947, opened 2017-10-10
Start to deprecate HTTPServerOption

wilzbach:
A hack to start triggering deprecation for `HTTPServerOption`.

I assumed that if we go this way, #1946 would have been incorporated.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -524,6 +524,13 @@ final class HTTPServerErrorInfo {
 alias HTTPServerErrorPageHandler = void delegate(HTTPServerRequest req, HTTPServerResponse res, HTTPServerErrorInfo error) @safe;
 
 
+private enum HTTPServerOptionImpl {
+	none                      = 0,
+	errorStackTraces          = 1<<7,
+	reusePort                 = 1<<8,
+}
+
+// TODO: Should be turned back into an enum once the deprecated symbols can be removed
 /**
 	Specifies optional features of the HTTP server.
 
@@ -533,20 +540,18 @@ alias HTTPServerErrorPageHandler = void delegate(HTTPServerRequest req, HTTPServ
 	will also drain the `HTTPServerRequest.bodyReader` stream whenever a request
 	body with form or JSON data is encountered.
 */
-enum HTTPServerOption {
-	none                      = 0,
-	/// Deprecated: Fills the `.path` and `.queryString` fields in the request
-	parseURL                  = 1<<0,
-	/// Deprecated: Fills the `.query` field in the request
-	parseQueryString          = 1<<1 | parseURL,
-	/// Deprecated: Fills the `.form` field in the request
-	parseFormBody             = 1<<2,
-	/// Deprecated: Fills the `.json` field in the request
-	parseJsonBody             = 1<<3,
-	/// Deprecated: Fills the `.files` field of the request for "multipart/mixed" requests
-	parseMultiPartBody        = 1<<4,
-	/// Deprecated: Fills the `.cookies` field in the request
-	parseCookies              = 1<<5,
+struct HTTPServerOption {
+	static enum none                      = HTTPServerOptionImpl.none;
+	deprecated("This is done lazily. It will be removed in 0.9.")
+	static enum parseURL                  = none;
+	deprecated("This is done lazily. It will be removed in 0.9.")
+	static enum parseQueryString          = none;
+	deprecated("This is done lazily. It will be removed in 0.9.")
+	static enum parseFormBody             = none;
+	deprecated("This is done lazily. It will be removed in 0.9.")
+	static enum parseJsonBody             = none;
+	deprecated("This is done lazily. It will be removed in 0.9.")
+	static enum parseMultiPartBody        = none;
 	/** Deprecated: Distributes request processing among worker threads
 
 		Note that this functionality assumes that the request handler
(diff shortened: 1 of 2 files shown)


http/vibe/http/server.d
```diff
@@ -530,20 +530,18 @@ alias HTTPServerErrorPageHandler = void delegate(HTTPServerRequest req, HTTPServ
...
-	parseFormBody             = 1<<2,
-	/// Deprecated: Fills the `.json` field in the request
-	parseJsonBody             = 1<<3,
-	/// Deprecated: Fills the `.files` field of the request for "multipart/mixed" requests
-	parseMultiPartBody        = 1<<4,
-	/// Deprecated: Fills the `.cookies` field in the request
-	parseCookies              = 1<<5,
+struct HTTPServerOption {
+	static enum none                      = 0;
+	deprecated("It is done lazy.")
```
WebFreak001: Maybe change these messages to `this is done lazily now`

WebFreak001: Plus this whole block of new lower case variants shouldn't exist at all because the upper case variants are being deprecated

http/vibe/http/server.d
```diff
@@ -569,38 +568,34 @@ enum HTTPServerOption {
...
+	static enum ParseFormBody = none;
+	deprecated("It is done lazy.")
+	static enum ParseJsonBody = none;
+	deprecated("It is done lazy.")
+	static enum ParseMultiPartBody = none;
+	deprecated("It is done lazy.")
+	static enum ParseCookies = none;
+
+    int x;
```
WebFreak001: There is a space too much here

http/vibe/http/server.d
```diff
@@ -530,20 +530,18 @@ alias HTTPServerErrorPageHandler = void delegate(HTTPServerRequest req, HTTPServ
...
-	parseQueryString          = 1<<1 | parseURL,
-	/// Deprecated: Fills the `.form` field in the request
-	parseFormBody             = 1<<2,
-	/// Deprecated: Fills the `.json` field in the request
-	parseJsonBody             = 1<<3,
-	/// Deprecated: Fills the `.files` field of the request for "multipart/mixed" requests
-	parseMultiPartBody        = 1<<4,
-	/// Deprecated: Fills the `.cookies` field in the request
-	parseCookies              = 1<<5,
+struct HTTPServerOption {
```
WebFreak001: Maybe add a TODO comment that this should be turned back to an enum once the deprecated symbols are removed

s-ludwig:
I thought about going this route too, but the fact that this loses type safety combined with the risk that this may remain a permanent installation (i.e. never reaches a state where there are no deprecated values) also lead me to abandon the idea. However, I think what would reasonably limit this risk is to define an additional `enum HTTPServerOptionImpl { ... }` that contains only the values that are still relevant and then convert the `HTTPServerOption.x` field to `HTTPServerOptionImpl`.

thaven:
That made me think.... Maybe we should go the opposite direction. Leave `HTTPServerOption` as an enum, and add a second enum for the _deprecated_ members. Those deprecated members appear in `HTTPServerOption` itself as forwards to the new enum, which is to be marked `deprecated`.

So it looks like this:
```D
enum HTTPServerOption {
    none = 0,
    None = HTTPServerOptionDepr.None,
    //....
}

deprecated enum HTTPServerOptionDepr {
    None = HTTPServerOption.none
}
```

Just popped intomy mind. Didn't test it.

s-ludwig:
I'd guess that that would warn at the `None = HTTPServerOption.none` line instead of in user code, but it would be worth a try.

s-ludwig:
Surprisingly, if the deprecated enum is declared after the regular one, there is no deprecation warning at all. But if they are reversed, the warning indeed happens within `HTTPServerOption`, unfortunately.

wilzbach:
> Plus this whole block of new lower case variants shouldn't exist at all because the upper case variants are being deprecated

@WebFreak001 there are no new upper case variants? Did you mean that the deprecated messages on the upper case variants are superfluous because they have been deprecated for a very long time? The problem here is that the user has never seen a deprecation message. 

>  However, I think what would reasonably limit this risk is to define an additional enum HTTPServerOptionImpl { ... } that contains only the values that are still relevant and then convert the HTTPServerOption.x field to HTTPServerOptionImpl.

Did you mean this like it's done now?

> I thought about going this route too, but the fact that this loses type safety combined with the risk that this may remain a permanent installation (i.e. never reaches a state where there are no deprecated values) also lead me to abandon the idea.

Hmm another idea about moving forward would be to go with a new concept as we only have `errorStackTraces` and `reusePort` as options left. The simplest one would be to create sth. like:

```d
enum HTTPServerConfig {
	none                      = 0,
	errorStackTraces          = 1<<7,
	reusePort                 = 1<<8,
	defaults = () { debug return HTTPServerConfig.errorStackTraces; else return HTTPServerConfig.none; } (),

}
```

However, I'm not sure whether we can safely use `HTTPServerConfig` directly or whether we would have to do sth. like this:

```d
        deprecated("Please use .config and HTTPServerConfig")
	HTTPServerOption options = HTTPServerOption.defaults;
	HTTPServerConfig config = HTTPServerConfig.defaults;
```

WebFreak001:
How about making breaking changes and waiting for 0.9.0 to merge this then? It should also like drop all other deprecated stuff

s-ludwig:
@wilzbach I think the current approach should work! It was mainly just the `int` field that was the issue. Hopefully the current push for enum attributes will lead somewhere, then there is also a clear path back to a plain enum type.

@WebFreak001 Things are getting deprecated and removed all the time, because that has been the only reasonable development mode until now. Instead of a 0.9.x branch, the plan is to break out individual sub packages into separate repositories and start with a 1.x.x version numbering there. The API will then at the same time be refactored with all the experience that has been built up so far, to ensure that the 1x branch can live as long as possible.

I'll wait with the merge after 0.8.2 has been released (I'll announce the RC in the coming days and then tag the release a week later).

PR #1948, opened 2017-10-10
Make vibe.web.web.redirect similar to HTTPServerResponse.redirect

wilzbach:
Provide the same API as  `HTTPServerResponse.redirect`

### DIFF
--- a/web/vibe/web/web.d
+++ b/web/vibe/web/web.d
@@ -356,7 +356,7 @@ template render(string diet_file, ALIASES...) {
 	Note that this may only be called from a function/method
 	registered using registerWebInterface.
 */
-void redirect(string url)
+void redirect(string url, int status = HTTPStatus.found)
 @safe {
 	import std.algorithm : canFind, endsWith, startsWith;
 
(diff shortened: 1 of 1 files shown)


web/vibe/web/web.d
```diff
@@ -378,6 +378,12 @@ void redirect(string url)
 	ctx.res.redirect(fullurl);
```
thaven: I suppose `status` should be passed here....

web/vibe/web/web.d
```diff
@@ -378,6 +378,12 @@ void redirect(string url)
...
 }
 
+/// ditto
+void redirect(URL url, int status = HTTPStatus.Found)
+@safe {
+	redirect(url.toString);
```
thaven: ... and here.

web/vibe/web/web.d
s-ludwig: And `found` instead of `Found` ("deprecated")

wilzbach:
Finally found time to rebase this PR - sorry for the forgetting to pass on `status`. Silly :S

PR #1949, opened 2017-10-11
Replace static assert in with unittest

guusw:
I had some issues with this static assert generating undefined references when building with -inline (in 'profile' builds) with dmd.

I created an issue on the dmd bug tracker (See https://issues.dlang.org/show_bug.cgi?id=17895) but this fixes my issue without having to wait for them to fix it and I don't see why this assert couldn't be a unit test instead.


s-ludwig:
Sounds reasonable!

PR #1951, opened 2017-10-13
throw useful error message on MongoDB query failures

MartinNowak:


s-ludwig:
Strange that Meson now fails for the libevent configuration. Master appears to be fine.

PR #1959, opened 2017-10-21
Fix Meson build

ximion:
Meson broke its wrap support in the 0.43 release, this PR excludes the faulty version from being used and also updates the Ninja version used for building.

This error is fixed in newer Meson versions, see https://github.com/mesonbuild/meson/commit/36d85db5d41b1f3a8e3f13a1680d1758353b23f7 for details.

Cheers,
    Matthias

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,10 @@ services:
 addons:
   apt:
     packages:
-    - libevent-dev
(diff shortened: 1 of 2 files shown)


s-ludwig:
Thanks!

PR #1962, opened 2017-10-24
Use more universal inout modifier instead of const

schveiguy:
Removes the extra methods just due to const. This is inout's wheelhouse.

RE: http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/post/46769

### DIFF
--- a/core/vibe/core/task.d
+++ b/core/vibe/core/task.d
@@ -66,10 +66,8 @@ struct Task {
 		}
 
 		// FIXME: this is not thread safe!
-		@property ref ThreadInfo tidInfo() { return m_fiber ? fiber.tidInfo : s_tidInfo; }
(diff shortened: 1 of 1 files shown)


s-ludwig:
Tried that, too ;) I decided to go with a `const` overload to avoid the cast that would have been needed otherwise.

s-ludwig:
For reference: https://travis-ci.org/vibe-d/vibe.d/jobs/292105177#L570
> core/vibe/core/task.d(69,60): Error: cannot implicitly convert expression this.m_fiber ? this.fiber().tidInfo() : s_tidInfo of type const(ThreadInfo) to inout(ThreadInfo)

schveiguy:
OK, I wasn't getting that the return value could depend on static data. This makes it unsuitable to `inout`, unfortunately, as there is no mechanism to match `mutable` and `inout` together like there is to match `immutable` and `inout`.

Issue #1958, opened 2017-10-21
TLS is almost unusable on ArchLinux


Geod24:
This is sort of a meta-issue, for other to find, and to summarize my experience here.
I got a Vibe.d application working flawlessly on Ubuntu 16.04. I tried it at home and have been trying to fix it for a couple of hours now.

- At first, it  didn't compile.
- Found undocumented `version = VibeUseOpenSSL11` (guess it should be on vibed.org?).
- It still didn't compile, now in Deimos:
```
Linking...
../../../../home/geod24/.dub/packages/vibe-d-0.8.2-alpha.2/vibe-d/tls/.dub/build/openssl-debug-linux.posix-x86_64-dmd_2076-CB8379A6F8B408EDF6C4E5778ED681DF/libvibe-d_tls.a(safestack_46d_449.o): In function `_D6deimos7openssl9safestack57__T10SKM_sk_numTS6deimos7openssl6x509v315GENERAL_NAME_stZ16__T10SKM_sk_numZ10SKM_sk_numFNbPS6deimos7openssl9safestack54__T8STACK_OFTS6deimos7openssl6x509v315GENERAL_NAME_stZ8STACK_OFZi':
/mnt/shared/projects/slacklib/../../../../home/geod24/.dub/packages/openssl-1.1.5_1.0.1g/openssl/deimos/openssl/safestack.d:140: undefined reference to `sk_num'
../../../../home/geod24/.dub/packages/vibe-d-0.8.2-alpha.2/vibe-d/tls/.dub/build/openssl-debug-linux.posix-x86_64-dmd_2076-CB8379A6F8B408EDF6C4E5778ED681DF/libvibe-d_tls.a(safestack_46f_516.o): In function `_D6deimos7openssl9safestack59__T12SKM_sk_valueTS6deimos7openssl6x509v315GENERAL_NAME_stZ18__T12SKM_sk_valueZ12SKM_sk_valueFNbPS6deimos7openssl9safestack54__T8STACK_OFTS6deimos7openssl6x509v315GENERAL_NAME_stZ8STACK_OFiZPS6deimos7openssl6x509v315GENERAL_NAME_st':
/mnt/shared/projects/slacklib/../../../../home/geod24/.dub/packages/openssl-1.1.5_1.0.1g/openssl/deimos/openssl/safestack.d:142: undefined reference to `sk_value'
collect2: error: ld returned 1 exit status
Error: linker exited with status 1
dmd failed with exit code 1.
```
[...]

PR #1965, opened 2017-10-25
Add OpenSSL 0.9 and 1.1 configurations to avoid the need for defining…

s-ludwig:
… version constants.

See #1958 and D-Programming-Deimos/openssl#35.


WebFreak001:
Shouldn't this be merged manually? I don't think the new versions have to do anything at all with that travis issue

PR #1969, opened 2017-11-01
IAllocator owns takes const void[]

edi33416:
After this PR in Phobos, https://github.com/dlang/phobos/pull/5831, `IAllocator`s `owns` must take a `const void[]`

### DIFF
--- a/utils/vibe/internal/utilallocator.d
+++ b/utils/vibe/internal/utilallocator.d
@@ -115,7 +115,11 @@ final class RegionListAllocator(Allocator, bool leak = false) : IAllocator {
 	} else {
 		override Ternary resolveInternalPointer(void* p, ref void[] result) { return Ternary.unknown; }
 	}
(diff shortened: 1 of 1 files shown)


edi33416:
Any ideas as to why CI is failing?

s-ludwig:
Something for Travis-CI must have changed, the other branches also fail now. I'll just merge this manually.

@wilzbach Fully agree. I see the `std.experimental` approach pretty much as a failure after seeing how this turned out with the allocator package. The logging module is also just sitting there with little evidence that it gets any significant use, so it's going to be tough to determine whether it's ready to go to non-experimental.

To be fair, allocators should have never entered `std.experimental` in the current condition, there are still glaring issues in the API (probably most importantly, handling attributes properly), certainly not worthy of a standard library module. On the other hand it's good to at least have something official, as this is definitely a pressing issue.

PR #1973, opened 2017-11-06
Fix proxy CONNECT request handling.

s-ludwig:
There were two issues:
- The destination resource ("host:port") was not parsed correctly
- The response header was missing

See #1893.


s-ludwig:
There is still an issue in the closing procedure where an assertion in the libevent driver is triggered, but this appears to be specific to the libevent driver. Using vibe-core does not show a similar issue.

PR #1990, opened 2017-12-02
Added spec defined websocket close enum

andrewbenton:
This enum and function provide access to the defined websocket close
codes and will also generate the reason string based off of the code if
one isn't already defined.  This saves library users from having to look
up documentation.

### DIFF
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -384,6 +384,7 @@ HTTPServerRequestDelegateS handleWebSockets(void function(scope WebSocket) @syst
  */
 enum WebSocketCloseReason : short
 {
+	none = 0,
 	normalClosure = 1000,
 	goingAway = 1001,
 	protocolError = 1002,
@@ -678,20 +679,16 @@ final class WebSocket {
 			code = Numeric code indicating a termination reason.
 			reason = Message describing why the connection was terminated.
 	*/
-	void close(short code = 0, scope const(char)[] reason = "")
+	void close(short code = WebSocketCloseReason.normalClosure, scope const(char)[] reason = "")
 	{
-		//assume the default is normal, intentional closure
-		if(code == 0)
-			code = WebSocketCloseReason.normalClosure;
-
-		if(reason is null || reason.length == 0)
+		if(reason.length == 0)
 			reason = (cast(WebSocketCloseReason)code).toString;
 
 		//control frame payloads are limited to 125 bytes
 		version(assert)
 			assert(reason.length <= 123);
 		else
-			reason = reason[0 .. 123];
+			reason = reason[0 .. min($, 123)];
 
 		if (connected) {
 			send((scope msg) {
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -359,6 +359,102 @@ HTTPServerRequestDelegateS handleWebSockets(void function(scope WebSocket) @syst
 	});
 }
 
+/**
+ * Provides the reason that a websocket connection has closed.
+ *
+ * Further documentation for the WebSocket and it's codes can be found from:
+ * https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+ *
+ * ---
+ *
+ * void echoSocket(scope WebSocket sock)
+ * {
+ *   import std.datetime : seconds;
+ *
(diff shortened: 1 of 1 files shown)


andrewbenton:
@s-ludwig Would you mind re-running that single failing test?  I can't determine from looking at the logs why it might have failed.

http/vibe/http/websockets.d
```diff
@@ -567,8 +680,18 @@ final class WebSocket {
 	*/
 	void close(short code = 0, scope const(char)[] reason = "")
 	{
+		//assume the default is normal, intentional closure
+		if(code == 0)
+			code = WebSocketCloseReason.normalClosure;
```
s-ludwig: This would remove the possibility to send a closing frame *without*  a code/reason, which may be desirable for efficiency if a closing reason isn't used anyway. Do you think that sending no status is an issue for existing endpoints in practice? In that case, the other option would be to change the default parameter to `code = WebSocketCloseReason.normalClosure` and to define `WebSocketCloseReason.none = 0` to allow explicitly skipping the status.

http/vibe/http/websockets.d
```diff
@@ -567,8 +680,18 @@ final class WebSocket {
...
+
+		if(reason is null || reason.length == 0)
+			reason = (cast(WebSocketCloseReason)code).toString;
+
 		//control frame payloads are limited to 125 bytes
-		assert(reason.length <= 123);
+		version(assert)
+			assert(reason.length <= 123);
+		else
+			reason = reason[0 .. 123];
```
s-ludwig: Should be `reason = reason[0 .. min($, 123)];` to avoid a range error.

http/vibe/http/websockets.d
```diff
@@ -567,8 +680,18 @@ final class WebSocket {
 	*/
 	void close(short code = 0, scope const(char)[] reason = "")
 	{
+		//assume the default is normal, intentional closure
+		if(code == 0)
+			code = WebSocketCloseReason.normalClosure;
+
+		if(reason is null || reason.length == 0)
```
s-ludwig: Nitpick: the `reason is null` is redundant here, `if (reason.length == 0)` would be sufficient.

http/vibe/http/websockets.d
andrewbenton: I like the idea of adding a `WebSocketCloseReason.none` and defaulting to normal.  I'm not sure if sending no status is an issue, though it probably acts as an unexpected closure.  From an API standpoint, I had expected that the default was a normal closure.

s-ludwig: My expectation was that it is usually trated as a normal closure, except if the peer application actually expects a status code, but I would be fine with the `.none` solution.

BTW, a related question is if the reason string should be omitted if none is given explicitly. Sending a generic string doesn't add much value and may be unwanted use of bandwidth for applications where this is critical.

andrewbenton: With the generic string, are you referring to the `0` case, or the spec-recognized cases, or both?  I admit that I like the idea of providing that string, though saving bandwidth is appealing too.

My motivations on this are driven by having to look up the close reason codes on the server when we receive a websocket close and also on the browser's side when we get a number only and I have to look it up.

Do you think that adding another method would be better, or maybe using the defaulting string behavior only for non-release versions?

http/vibe/http/websockets.d
andrewbenton: Thanks for this catch!

http/vibe/http/websockets.d
andrewbenton: Please let me know how you want to go with this since everything else is green on the PR.

s-ludwig: As for the string, I think it would be fine to just treat `null` as a special case to still allow people to disable the reason phrase if bandwidth is critical. So the default phrase would just be set `if (reason !is null && reason.length == 0)`. Using different behavior for debug/release is something I'd rather not do in this case, since I'd rate the reason phrase to be slightly more than just a debugging aid (as opposed to sending a stack trace or similar things).

andrewbenton:
Unless you have any further change requests, I think I'm done with this PR.

s-ludwig:
Merging now, thanks again! I was just waiting until the 0.8.2 release was out, so that I didn't have to create a release branch.

PR #1998, opened 2017-12-11
Remove 2.071.x compatibility code

s-ludwig:


s-ludwig:
Meson failing again?

s-ludwig:
May be an issue with the tag_version script, since it started to fail in 38d5da2bd87595cf4c9c6c744301d3ee9300e730 with:
```
Meson encountered an error in file meson.build, line 1, column 0:

Subproject_dir must not contain a path segment.
```

I'll have a look.

s-ludwig:
Hm, it's actually still https://github.com/mesonbuild/meson/issues/2719 and there is just a new Meson version that still has the issue.

.travis.yml
```diff
@@ -52,7 +52,7 @@ matrix:
 
 before_install:
   - pyenv global system 3.5
-  - pip3 install 'meson>=0.42,!=0.43'
+  - pip3 install 'meson>=0.42,!=0.43,!=0.44'
```
wilzbach: `pip install 'meson==0.42'` would be a lot safer ...

s-ludwig: That's true, I was also thinking about that. I guess the idea behind that was to not forget about this issue and enforce getting back to the latest release at some point. To me it would be fine to keep it like this for another version, but then I'd also rather like to go with `==0.42` to avoid this becoming a regular distraction.

PR #1999, opened 2017-12-12
Portable way to get the processor count on Linux

rracariu:
The initial code was not very portable across various C run-time libs, for example uClibc doesn't provide the `get_nprocs` function


s-ludwig:
Thanks, that makes sense! The only thing missing is a `cast(uint)`, because the CI is failing due to `sysconf` returning a `c_long`.

rracariu:
Cool, thanks! Added the cast, hope all is good.

PR #2003, opened 2017-12-14
Switch to Python 3.6 env to work around Travis-CI failures

s-ludwig:


mongodb/vibe/db/mongo/sessionstore.d
```diff
@@ -15,6 +15,20 @@ import std.datetime : Clock, SysTime, UTC;
...
+	import vibe.http.server : HTTPServerSettings;
+	import core.time : hours;
+
+	auto store = new MongoSessionStore("mongodb://127.0.0.1/mydb", "sessions");
+	store.expirationTime = 5.hours;
+
+	auto settings = new HTTPServerSettings("127.0.0.1:8080");
+	settings.sessionStore = store;
+
+	//listenHTTP(settings, router);
```
wilzbach: Idea: you could define a stub for `listenHTTP` which is only built for the vibe.d testsuite?

s-ludwig: Another idea that just occurred to me now is to wrap the code inside a nested `main()`  function instead of the `test()` function that I've sometimes used for similar examples. That way it can ideally be used as a copy-paste example, but is not actually executed.

I'll try that in a separate PR. This commit accidentally slipped in, because I still had it in master when creating the branch.

Issue #2004, opened 2017-12-14
HTTPServerRequest.context.get cannot be instantiated with classes and interfaces


quickfur:
Code:
````
interface MyInterface { }
void func(HTTPServerRequest req) {
    auto intf = req.context.get!MyInterface("abc");
}
````

Compiler output:
````
/usr/src/d/vibe.d/utils/vibe/utils/dictionarylist.d(198): Error: cannot implicitly convert expression `this.get(key, delegate const(MyInterface)() => null)` of type `const(MyInterface)` to `inout(MyInterface)`
main.d(41): Error: template instance vibe.utils.dictionarylist.DictionaryList!(VariantN!32LU, true, 2LU, false).DictionaryList.get!(MyInterface) error instantiating
````

The same error occurs if the `interface` is changed to `class`, and if the class/interface isn't directly stored in `req.context`, but is part of a `struct`.

Expected behaviour: it should be possible to store class and interface references in `HTTPServerRequest.context`.

Currently-known workaround: specify the default argument with `T.init` explicltly.
````
interface MyInterface { }
void func(HTTPServerRequest req) {
    auto intf = req.context.get!MyInterface("abc",
        MyInterface.init); // <-- this works
}
````

I looked at the code and found comments mentioning a dmd bug and a bunch of overloads that supposedly works around said bug. However, the workaround doesn't seem to work for classes/interfaces.

PR #2005, opened 2017-12-14
Fix issue 2004: DictionaryList.get doesn't work with class/interface

quickfur:
Turns out that the fix for issue #2004 is trivial: the default argument just needs to be specified with the `inout` qualifier.

### DIFF
--- a/utils/vibe/utils/dictionarylist.d
+++ b/utils/vibe/utils/dictionarylist.d
@@ -402,7 +402,7 @@ unittest {
 }
 
 // Issue 2004
-@safe unittest {
+unittest {
 	import std.variant : Variant;
 	DictionaryList!Variant l;
 	class C {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks for the PR! Looks like this should indeed work on all DMD versions. Can you do a quick rebase on top of the current master, though? Today something on the Travis-CI image changed, breaking the build, and the workaround has just been merged

quickfur:
Rebased.

utils/vibe/utils/dictionarylist.d
```diff
@@ -400,3 +400,15 @@ unittest {
 	assert(text(l) == `["foo": 42, "bar": 43]`, text(l));
 	assert(l.toString() == `["foo": 42, "bar": 43]`, l.toString());
 }
+
+// Issue 2004
+@safe unittest {
```
s-ludwig: I think the failures are coming from `Variant` not being annotated `@trusted`. Since this is just about the `inout` issue, I'll just quickly remove the `@safe`.

quickfur:
Whoa, this is still sitting here?  What gives with the AppVeyor problem? Anything I can do on my end?

quickfur:
Saw some commits related to working around AppVeyor issues; rebased on master to see if that helps.

PR #2008, opened 2017-12-17
Use std.parallelism.totalCPUs, it's optimized

kubo39:
We don't have to call sysconf(3) every workerThreadConut calls.

### DIFF
--- a/core/vibe/core/core.d
+++ b/core/vibe/core/core.d
@@ -691,8 +691,8 @@ public void setupWorkerThreads(uint num = logicalProcessorCount())
 public @property uint logicalProcessorCount()
 {
 	version (linux) {
-		import core.sys.posix.unistd;
-		return cast(uint) sysconf(_SC_NPROCESSORS_ONLN);
+		import std.parallelism : totalCPUs;
+		return totalCPUs;
 	} else version (OSX) {
 		int count;
 		size_t count_len = count.sizeof;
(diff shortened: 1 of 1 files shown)


core/vibe/core/core.d
```diff
@@ -691,8 +691,8 @@ public void setupWorkerThreads(uint num = logicalProcessorCount())
 public @property uint logicalProcessorCount()
 {
 	version (linux) {
-		import core.sys.posix.unistd;
-		return cast(uint) sysconf(_SC_NPROCESSORS_ONLN);
+		import std.parallelism : totalCPUs;
+		return totalCPUs;
 	} else version (OSX) {
```
wilzbach: Why don't you use it for the other OSes as well?

kubo39: Sure, why not! Thanks pointed out!

wilzbach:
BTW two other things:

- vibe-core will be the default for vibe.d 0.8.3, so you would have to change it here: https://github.com/vibe-d/vibe-core/blob/master/source/vibe/core/core.d#L603
- this function is only called once during setup, but if we can use Phobos for it, it would make sense to deprecate it too (I'm on my phone atm, so I didn't check the history for the reason of inclusion).

s-ludwig:
Agree with the change, no need to maintain a separate version for this, regardless of performance. I actually never saw this in `std.parallelism` up to now, just `coresPerCPU` in the `std.cpuid` module. I'll have a look at vibe-core.

PR #2013, opened 2017-12-20
Fix RedisHash!int.opIndexOpAssign

Geoffrey-A:
Hello,

A fix as per title.

Thank you very much for dub and vibe.d!

Best regards,
Geoffrey


s-ludwig:
Thanks!

PR #2016, opened 2017-12-21
more control over running test components

MartinNowak:
- replace ad-hoc BUILD_EXAMPLE & RUN_TEST mechanism
- allows to test components in parallel

### DIFF
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -2,47 +2,55 @@
 
 set -e -x -o pipefail
 
-DUB_ARGS=${DUB_ARGS:-}
-
-./scripts/test_version.sh
-
-# Check for trailing whitespace"
-grep -nrI --include=*.d '\s$'  && (echo "Trailing whitespace found"; exit 1)
-
-# test for successful release build
-dub build --combined -b release --compiler=$DC --config=${VIBED_DRIVER=libevent}
-dub clean --all-packages
-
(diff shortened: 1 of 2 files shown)


travis-ci.sh
```diff
@@ -2,47 +2,55 @@
...
-
-./scripts/test_version.sh
-
-# Check for trailing whitespace"
-grep -nrI --include=*.d '\s$'  && (echo "Trailing whitespace found"; exit 1)
+DUB_ARGS="--build-mode=${DUB_BUILD_MODE:-separate} ${DUB_ARGS:-}"
+# default to run all parts
+: ${PARTS:=lint,builds,unittests,examples,tests,meson}
+# default to libevent driver
+: ${VIBED_DRIVER:=libevent}
```
wilzbach: vibe-core is the new default: https://github.com/vibe-d/vibe.d/pull/2002

s-ludwig:
Seconded, makes this a lot cleaner to control. BTW, I think we need to make some DUB improvements before parallel builds work reliably. Last time I tried that I got all sorts of mutual file access issues.

Two of those things would be locking build directories and feeding the specific build output (`.dub/build/xxx/yyy.lib`) into dependent packages instead of the main target file, which could have been changed by a concurrent build.

MartinNowak:
Yes, dub doesn't lock builds atm. and dmd doesn't write output files atomically, so build fail.
I'm more interested in concurrent (isolated) builds of vibe.d to reduce the test time on https://ci.dlang.io for now though.

PR #2021, opened 2017-12-29
add @successCode UDA to return other 2XX codes

MartinNowak:


web/vibe/web/rest.d
```diff
@@ -1398,15 +1398,17 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
 			static if (is(RT == void)) {
 				() @trusted { __traits(getMember, inst, Method)(params); } (); // TODO: remove after deprecation period
 				returnHeaders();
+				res.statusCode = extractSuccessCode!CFunc();
 				res.writeBody(cast(ubyte[])null);
 			} else {
 				auto ret = () @trusted { return __traits(getMember, inst, Method)(params); } (); // TODO: remove after deprecation period
 
-				static if (!__traits(compiles, () @safe { evaluateOutputModifiers!Func(ret, req, res); } ()))
+				static if (!__traits(compiles, () @safe { evaluateOutputModifiers!CFunc(ret, req, res); } ()))
```
MartinNowak: That's a drive-by to match the actual invocation below.

MartinNowak:
Not too certain whether a `status = code` solution like for vibe.web.web would be better.
Definitely still need some plain `(req, res)` to deal with endpoints returning either 204 or 201.
Also optional void returns (204) aren't really supported.

Geod24:
This can be done using `@viaStatus` UDA.

PR #2025, opened 2018-01-09
Add support for parsing cookies.

Computermatronic:
Adds support for parsing cookies from headers for the vibe HTTP client.

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -888,7 +888,7 @@ final class HTTPClientResponse : HTTPResponse {
 
 	/// All cookies that shall be set on the client for this request
 	override @property Cookie[string] cookies() {
-		if ("Set-Cookie" in this.headers && cookies is null) {
+		if ("Set-Cookie" in this.headers && m_cookies is null) {
 			foreach(cookieString; this.headers.getAll("Set-Cookie")) {
 				auto cookie = parseHTTPCookie(cookieString);
 				m_cookies[cookie.name] = cookie;
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -23,6 +23,7 @@ import vibe.stream.operations;
 import vibe.stream.wrapper : createConnectionProxyStream;
 import vibe.stream.zlib;
 import vibe.utils.array;
+import vibe.utils.dictionarylist;
 import vibe.internal.allocator;
 import vibe.internal.freelistref;
 import vibe.internal.interfaceproxy : InterfaceProxy, interfaceProxy;
@@ -876,7 +877,6 @@ final class HTTPClientResponse : HTTPResponse {
 		InterfaceProxy!InputStream m_bodyReader;
 		bool m_closeConn;
 		int m_maxRequests;
-		Cookie[string] m_cookies;
 	}
 
 	/// Contains the keep-alive 'max' parameter, indicates how many requests a client can
@@ -887,11 +887,11 @@ final class HTTPClientResponse : HTTPResponse {
 
 
 	/// All cookies that shall be set on the client for this request
-	override @property Cookie[string] cookies() {
-		if ("Set-Cookie" in this.headers && m_cookies is null) {
+	override @property DictionaryList!Cookie cookies() {
+		if ("Set-Cookie" in this.headers && m_cookies.length == 0) {
 			foreach(cookieString; this.headers.getAll("Set-Cookie")) {
 				auto cookie = parseHTTPCookie(cookieString);
-				m_cookies[cookie.name] = cookie;
+				m_cookies[cookie[0]] = cookie[1];
 			}
 		}
 		return m_cookies;
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -876,12 +876,25 @@ final class HTTPClientResponse : HTTPResponse {
 		InterfaceProxy!InputStream m_bodyReader;
 		bool m_closeConn;
 		int m_maxRequests;
+		Cookie[string] m_cookies;
 	}
 
 	/// Contains the keep-alive 'max' parameter, indicates how many requests a client can
 	/// make before the server closes the connection.
 	@property int maxRequests() const {
 		return m_maxRequests;
+	}
+
+
+	/// All cookies that shall be set on the client for this request
+	override @property Cookie[string] cookies() {
+		if ("Set-Cookie" in this.headers && cookies is null) {
+			foreach(cookieString; this.headers.getAll("Set-Cookie")) {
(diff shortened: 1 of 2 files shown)


wilzbach:
> Fix unittest again? I swear dub test dosen't work!

Try:

```
dub test :http
```

`dub test` doesn't test all subpackage - the first time I also ran into this. It's a bit tricky ;-)

Computermatronic:
Thanks.

http/vibe/http/common.d
```diff
@@ -712,6 +712,45 @@ final class Cookie {
 		}
 	}
 
+	/// Parses the cookie from a header field, returning the name of the cookie.
+	string parse(string headerString) {
+		auto parts = headerString.splitter("; ");
```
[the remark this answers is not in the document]
thaven: That would be except the first one, thus making the first one require different treatment. So I think it may be better to leave this as is, and throw the exception if any part contains ';', though that would be a little slower, as you have to walk the string instead of just comparing the first char.

http/vibe/http/common.d
```diff
@@ -712,6 +712,45 @@ final class Cookie {
 		}
 	}
 
+	/// Parses the cookie from a header field, returning the name of the cookie.
+	string parse(string headerString) {
```
WebFreak001: I think it's really confusing that parse returns a string and alters the internal state, wouldn't it be smarter to make this a static method and return a `Cookie` instance?

s-ludwig: A tuple of `string` and `Cookie` that is, but I agree that this would be nicer. Maybe this method should just be `package` for now, so that we can move on with the main functionality without worriying too much about the API here (which may have to change in the future anyway when `Cookie` for example gets changed to a `struct`).

http/vibe/http/common.d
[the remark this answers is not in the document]
s-ludwig: That plus `indexOf`` with a string argument is also considerably slower than the overload taking a single character. But isn't the first part already handled separately outside of the loop anyway?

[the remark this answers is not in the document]
Computermatronic: I've been doing some looking, and according to the RFC ([https://tools.ietf.org/html/rfc6265#section-4.1.1](url)) for Set-Cookie, a space after the semi-colon is not required, and all leading / trailing spaces should be stripped. I'll probably switch to a `splitter(';')` and use `trim(key)` and `trim(value)`

http/vibe/http/common.d
Computermatronic: From what I have observed, most parsing functions are free-standing, so it makes more sense to make `name` part of `Cookie` class and have a `parseHTTPCookie` free-standing function, which returns a cookie

http/vibe/http/client.d
```diff
@@ -917,7 +917,12 @@ final class HTTPClientResponse : HTTPResponse {
 		foreach (k, v; this.headers)
 			logTrace("%s: %s", k, v);
 		logTrace("---------------------");
-
+		if ("Set-Cookie" in this.headers) {
```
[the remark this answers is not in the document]
Computermatronic: Actually `HTTPServerRequest` seems to do `HTTPRequest.cookies` for session management, however it is easy enough to just put `m_cookies` in `HTTPRequest` with a `@property` getter and override it in `HTTPClientRequest`.

http/vibe/http/client.d
```diff
@@ -884,6 +885,18 @@ final class HTTPClientResponse : HTTPResponse {
 		return m_maxRequests;
 	}
 
+
+	/// All cookies that shall be set on the client for this request
+	override @property Cookie[string] cookies() {
+		if ("Set-Cookie" in this.headers && cookies is null) {
```
WebFreak001: this is recursive, it should check m_`cookies` instead

Computermatronic: Oops, I'll fix that.

Computermatronic:
`Cookie` and `CookieValueMap.Cookie` should be merged, and `CookieValueMap` replaced with a `DictionaryList` along with the `string[string]` used in `HTTPResponse`, so that there is a single way to deal with cookies, multiple cookies with the same name are supported.

s-ludwig:
The reason `CookieValueMap` is the additional support for specifying encodings. But of course it could, and should, be changed to use `DictionarlyList` under the hood. `DictionaryList` for `HTTP(Server)Response.cookies` makes sense, too.

As for `Cookie` vs. `CookieValueMap.Cookie`, the problem is that, at least in the low-level API, all fields except for the `value` field would be unused on the client side, which would be rather confusing/error prone. As soon as a cookie store is involved everything would fit though.

http/vibe/http/common.d
```diff
@@ -618,11 +620,39 @@ FreeListRef!ChunkedOutputStream createChunkedOutputStreamFL(OS)(OS destination_s
 	return FreeListRef!ChunkedOutputStream(interfaceProxy!OutputStream(destination_stream), allocator, true);
 }
 
+/// Parses the cookie from a header field, returning the name of the cookie.
+Cookie parseHTTPCookie(string headerString)
```
s-ludwig: There is an extra space at the end of the line here, which makes the travis build fail.

http/vibe/http/common.d
```diff
@@ -638,6 +668,11 @@ final class Cookie {
 		none = raw
 	}
 
+	/// Cookie anme
+	@property void name(string name) { m_name = name; }
+	/// ditto
+	@property string name() const { return m_name; }
```
s-ludwig: I don't really like the data redundancy that this introduces. For now it would really be better to keep the API as it is and let `parseHTTPCookie` return a pair of name and cookie. It should also be marked `package`, so that we can discuss the ideal API in a separate session, no need to conflate this with the functional changes of this PR.

Computermatronic:
Why have the appveyor checks failed?

WebFreak001:
dub was probably down again

wilzbach:
I restarted AppVeyor.

wilzbach:
But I think https://github.com/vibe-d/vibe.d/pull/2038 is needed to fix Travis.

Computermatronic:
I have incorporated the recommended changes not too long ago, but I've only got back to this now.
Any other changes that need to take place?

wilzbach:
> Any other changes that need to take place?

Any chance we can add a "high-level" test for this in `tests`? e.g. the server sending a request with two cookies set, the client looking at them, manipulating them a bit and sending them back?

wilzbach:
Ping @Computermatronic. I think the only bit missing here is a more high-level test.

WebFreak001:
can be closed now (my PR of this with updates got merged)

PR #2026, opened 2018-01-10
Fix: forward RedisValue methods hidden in redis types by overloads

Geoffrey-A:
All supported redis values now work as per the test below:
```d
auto db = connectRedis("127.0.0.1").getDatabase(0);
db.getAsList("test_list").remove();
db.getAsSet("test_set").remove();
db.getAsZSet("test_zset").remove();
db.getAsHash("test_hash").remove();
db.getAsString("test_string").remove();
```

### DIFF
--- a/redis/vibe/db/redis/types.d
+++ b/redis/vibe/db/redis/types.d
@@ -602,6 +602,7 @@ struct RedisZSet(T = string) {
 
 	long insert(ARGS...)(ARGS args) { return m_db.zadd(m_key, args); }
 	long remove(ARGS...)(ARGS members) { return m_db.zrem(m_key, members); }
+	void remove()() { value.remove(); }
 	long length() { return m_db.zcard(m_key); }
 
 	long count(string INT = "[]")(double min, double max)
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thanks! LGTM.

Geoffrey-A:
Seeing it was not merged after a few days, I added a few commits. They fix the forwarding of the hidden RedisValue methods in the redis types which were hiding them:
1. forward the boolean result returned by remove();
2. forward exists() in RedisHash, which was hiding it with its own overload.

I also checked there was no other method from RedisValue that was hidden by any redis type.

Since I committed, the dlang bot removed the auto-merge tag.

Please allow me a few  questions:
What is the benefit of the auto-merge tag over simply merging?
Why was not this PR merged? I noticed there was at that time some CI failures, maybe that was the reason? However the failures seemed completely unrelated to this PR.

s-ludwig:
The "auto-merge" is supposed to provide a "shoot and forget" mode, where you don't have to wait until all CI checks complete. Unfortunately there are currently still too many network related failures, so that this often fails and then sits in the queue unnoticed. But this will hopefully be sorted out in a while.

Fortunately I can merge immediately now, since the tests have already finished...

PR #2027, opened 2018-01-10
Use old MongoAuthMechanism if explictly stated

wilzbach:
Applied the comment from
https://github.com/vibe-d/vibe.d/issues/1967#issuecomment-349782134


Computermatronic:
9kt

PR #2028, opened 2018-01-10
Fix compilation on DragonFlyBSD

dkgroot:
- Also do a little formatting cleanup
- Add static assert to prevent compilation when IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP are missing (not sure if this is preferred or not)

### DIFF
--- a/core/vibe/core/drivers/libevent2.d
+++ b/core/vibe/core/drivers/libevent2.d
@@ -77,6 +77,13 @@ version(Windows) {
 	}
 	else
 		import core.sys.freebsd.netinet.in_ : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
+} else version (DragonFlyBSD) {
+	static if (__VERSION__ < 2077) {
+		enum IP_ADD_MEMBERSHIP  = 12;
+		enum IP_MULTICAST_LOOP  = 11;
+	}
+	else
+		import core.sys.dragonflybsd.netinet.in_ : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
 } else version (linux) {
 	static if (__VERSION__ < 2077) {
 		enum IP_ADD_MEMBERSHIP =  35;
--- a/core/vibe/core/drivers/libevent2.d
+++ b/core/vibe/core/drivers/libevent2.d
@@ -60,41 +60,53 @@ version (Windows)
 else
 	version(VibePragmaLib) pragma(lib, "event");
 
-version(Windows) {
+version(Windows)
+{
 	import core.sys.windows.winsock2;
 	alias EWOULDBLOCK = WSAEWOULDBLOCK;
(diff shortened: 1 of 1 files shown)


core/vibe/core/drivers/libevent2.d
```diff
@@ -77,6 +77,13 @@ version(Windows) {
 	}
 	else
 		import core.sys.freebsd.netinet.in_ : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
+} else version (DragonFlyBSD) {
+	static if (__VERSION__ < 2077) {
```
Geod24: Shouldn't this be 2079 ? Since it's not merged yet, and 2078 was already released.

core/vibe/core/drivers/libevent2.d
```diff
@@ -60,34 +60,53 @@ version (Windows)
...
 		enum IP_ADD_MEMBERSHIP  = 12;
 		enum IP_MULTICAST_LOOP  = 11;
 	}
 	else
 		import core.sys.freebsd.netinet.in_ : IP_ADD_MEMBERSHIP, IP_MULTICAST_LOOP;
-} else version (linux) {
+}
+else version (DragonFlyBSD)
+{
+	static if (__VERSION__ < 2077) {
```
thaven: Somehow the comment of @Geod24 is shown as outdated, but he is right, this should be `__VERSION__ < 2079` (or even later, if `core.sys.dragonflybsd` doesn't get merged into druntime master before 2.079)

core/vibe/core/drivers/libevent2.d
dkgroot: You raising this question, made me think about this again. As this source is only going to be compiled/used when the changes have already landed in the compiler + druntime. It does make sense to even check the compiler version. I will remove this static if and just import core.sys.dragonflybsd.netinet.in_. There really is not need for the fall back.

dkgroot:
@thaven I fully agree. It would make more sense and be a lot nicer if there was an generic interface that could be referred to.

core/vibe/core/drivers/libevent2.d
dkgroot: [See](https://github.com/vibe-d/vibe.d/pull/2028#discussion_r161077152)
No need to check the compiler version. It's like a chicken/egg issue, without compiler support there is not way to get to this code. When compiler support lands, then core.sys.dragonflybsd.netinet.in_ will exist.

s-ludwig:
Usually those constants actually are in druntime, but unfortunately they are just being added on a per-case basis instead of thoroughly going through each C header. I used to make druntime PR's for such missing constants, but lately I've been getting too busy and had to focus on just getting vibe.d to work.

dkgroot:
Ok understood. Thanks for merging this PR !

PR #2029, opened 2018-01-10
small optimization for jsonEscape

Boris-Barboris:
According to Callgrind this function is the heavies one in the process of json serialization of string-heavy objects. This pull request is aimed to improve it's performance a little by reducing the frequency of dst.put calls and getting rid of lastch reassignment on every iteration. This reduces the time (exclusive) spent in this function by ~2% and reduces the number of "put" invocations considerably, giving ~10% improvement of total (inclusive) time on my test case.   

Also, are you sure about that 
```
pos -= 1;
```
in escape_unicode version of this function? I'm getting an infinite loop feeling.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -2164,15 +2164,24 @@ string convertJsonToASCII(string json)
 /// private
 private void jsonEscape(bool escape_unicode = false, R)(ref R dst, string s)
 {
-	char lastch;
+	size_t startPos = 0;
+	
+	void putInterval(size_t curPos)
+	{
+		if (curPos > startPos)
+			dst.put(s[startPos..curPos]);
+		startPos = curPos + 1;
+	}
+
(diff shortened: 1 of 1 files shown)


data/vibe/data/json.d
```diff
@@ -2164,15 +2164,24 @@ string convertJsonToASCII(string json)
...
 			default:
 				static if (escape_unicode) {
-					if (ch > 0x20 && ch < 0x80) dst.put(ch);
-					else {
+					if (ch <= 0x20 || ch >= 0x80)
+					{
+						putInterval(pos);
 						import std.utf : decode;
 						int len;
 						dchar codepoint = decode(s, pos);
```
s-ludwig: `startPos` needs to be adjusted accordingly here, because `decode` may advance more that the `char` that is assumed by `putInterval`.

Boris-Barboris: oh,  got it

PR #2031, opened 2018-01-13
Disable the default caching of the HTTP file server.

s-ludwig:
Caching as an optimization with possibly undesired side-effects should be enabled consciously instead of being the default.


wilzbach:
BTW should we disable all branches except for master on AppVeyor and Travis? It would save a bit of our precious CI resources.

wilzbach:
> BTW should we disable all branches except for master on AppVeyor and Travis? It would save a bit of our precious CI resources.

https://github.com/vibe-d/vibe.d/pull/2033

PR #2032, opened 2018-01-13
Fix a regression in the HTTPServerRequest.rootDir property.

s-ludwig:
Got introduced with the addition of vibe-core compatibility.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -1141,9 +1141,17 @@ final class HTTPServerRequest : HTTPRequest {
 	*/
 	@property string rootDir()
 	const @safe {
-		import std.range.primitives : walkLength;
-		auto depth = requestPath.bySegment.walkLength;
+		import std.algorithm.searching : count;
+		auto depth = requestPath.bySegment.count!(s => s.name.length > 0);
+		if (!requestPath.endsWithSlash) depth--;
 		return depth == 0 ? "./" : replicate("../", depth);
+	}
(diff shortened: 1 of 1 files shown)


http/vibe/http/server.d
```diff
@@ -1141,10 +1141,18 @@ final class HTTPServerRequest : HTTPRequest {
 	*/
 	@property string rootDir()
 	const @safe {
-		import std.range.primitives : walkLength;
-		auto depth = requestPath.bySegment.walkLength;
+		import std.algorithm.searching : count;
+		auto depth = requestPath.bySegment.count!(s => s.name.length > 0);
+		if (!requestPath.endsWithSlash) depth--;
 		return depth == 0 ? "./" : replicate("../", depth);
```
wilzbach: Hmm that looks potentially dangerous. `count` returns an unsigned number.
Consider:

```d
/+ dub.sdl:
dependency "vibe-d:core" version="~master"
+/
import vibe.core.path;
import std.stdio;
import std.algorithm;

void main()
{
	auto p = InetPath("");
	p.bySegment.writeln;
	p.endsWithSlash.writeln;
	size_t depth = p.bySegment.count;
	if (!p.endsWithSlash) depth--;
	depth.writeln;
}
```

https://run.dlang.io/is/toQxsC

s-ludwig: Good catch! Fixed and rebased.

PR #2033, opened 2018-01-13
Only build master on Travis

wilzbach:
Should reduce this redundancy from four to two CI runs:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,11 @@ addons:
       - libevent-dev
       - libssl-dev
 
+branches:
+  only:
+  - master
+  - 0.7.x
+
 d:
   # order: latest DMD, oldest DMD, LDC/GDC, remaining DMD versions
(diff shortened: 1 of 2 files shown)


s-ludwig:
I actually enabled other branches a while ago, because most notably the 0.7.x branch wasn't tested. It was actually quite confusing to discover the reason inside of the appveyor config file on master. What I wonder is why the setting in their UI doesn't restrict the build to those two branches: https://ci.appveyor.com/project/s-ludwig/vibe-d/settings

wilzbach:
> because most notably the 0.7.x branch wasn't tested.

I added that one to the whitelist.

> What I wonder is why the setting in their UI doesn't restrict the build to those two branches: 

I don't know it either :/ 
Maybe some settings aren't applied if you use a `appveyor.yml`?

wilzbach:
I guess the "workaround" then would be to do PRs from your user repository?

PR #2038, opened 2018-01-15
Bump versions to 2.078.0

wilzbach:
As a side effect this should fix the Travis failures.
Lacking a Windows VM, I did the changes to `appveyor.yml` blindly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,17 +19,17 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.077.1
-  - dmd-2.072.2
+  - dmd-2.078.0
+  - dmd-2.073.2
+  - ldc-1.7.0
   - ldc-1.6.0
   - ldc-1.5.0
   - ldc-1.4.0
   - ldc-1.3.0
-  - ldc-1.2.0
+  - dmd-2.077.1
   - dmd-2.076.1
   - dmd-2.075.1
   - dmd-2.074.1
-  - dmd-2.073.2
   - dmd-beta
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ matrix:
 
 before_install:
   - pyenv global system 3.6
-  - pip3 install 'meson>=0.42,!=0.43,!=0.44'
+  - pip3 install 'git+git://github.com/mesonbuild/meson.git@ca2db0f482e17b89d1b05ad056e4815bac4ad95d'
 
 install:
   - mkdir .ntmp
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ d:
   - dmd-beta
 
 env:
-  - VIBED_DRIVER=libevent PARTS=lint,builds,unittests,examples,tests,meson
+  - VIBED_DRIVER=libevent PARTS=lint,builds,unittests,examples,tests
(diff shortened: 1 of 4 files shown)


wilzbach:
Travis fails with:

```
Meson encountered an error in file meson.build, line 1, column 0:
Unknown compiler "dmd"
```

CC @ximion

ximion:
@wilzbach This was fixed in Meson master: https://github.com/mesonbuild/meson/pull/2889 - Unfortunately, we'll have to wait for a new release to get this fixed on the CI, or have the CI use the Meson master branch.

wilzbach:
>  or have the CI use the Meson master branch.

I went with locking your PR commit:

```
pip3 install 'git+git://github.com/mesonbuild/meson.git@5133d8650fb12e1635fda6a36794fdff9262bb7a
```

Shouldn't be too bad & at least we don't run the risk of Meson randomly breaking with new released versions again.

ximion:
It shouldn't break, actually... But Meson 0.44 was an insane release, with quite a few weird regressions (most are fixed now).
So pinning that commit for now at least is likely a good call.

wilzbach:
@ximion using master doesn't work either:

```
Meson encountered an error in file meson.build, line 1, column 0:
Subproject_dir must not contain a path segment.
```

ximion:
@wilzbach That's because you don't use the master branch, but some pinned commit. Either switch to the real master branch, or pin commit `ca2db0f482e17b89d1b05ad056e4815bac4ad95d` instead.
`pip3 install 'git+git://github.com/mesonbuild/meson.git@ca2db0f482e17b89d1b05ad056e4815bac4ad95d`
They made Meson overreact a bit when the subproject dir was changed, to fix a different issue of people setting it to absolute paths or some other crazy locations outside of the project directory.
Unfortunately, this broke a lot of legitimate uses.
(I do hope we'll get a 0.44.1 bugfix release soon)

wilzbach:
> That's because you don't use the master branch, but some pinned commit. 

Hmm, I thought I used the merge commit of your PR.
In any case, I gave `ca2db0f482e17b89d1b05ad056e4815bac4ad95d` a try now - thanks for the quick response!

ximion:
@wilzbach That other bug was fixed in a different PR that was merged later :-)

wilzbach:
Still not working :/

```
Executing subproject diet.
Project name: Diet-NG
Native D compiler: dmd (dmd 2.078.0 "DMD64 D Compiler v2.078.0")
Also couldn't find a fallback subproject in lib/subprojects/diet for the dependency diet
Meson encountered an error in file meson.build, line 60, column 0:
Native dependency 'diet' not found
```

ximion:
@wilzbach Interesting, I can't reproduce this locally... - I'll try a few things.

wilzbach:
@ximion I will temporarily comment out building meson, s.t. we have green builds here again. Please feel submit a PR once you got things working on your side.

ximion:
@wilzbach It's another bug related to the subproject sandboxing feature... This is getting really annoying.
This doesn't mean all the projects using Meson are broken (obviously not, we can still compile Mesa and all of GNOME), but the ones using a custom subprojects directory are.
We could work around this in Vibe by just having a new toplevel directory called "subprojects" and put the wrapfiles in there.

In any case, now that I have tracked down this bug, I will send in another PR for Meson - and maybe an unittest for this scenario later as well.

ximion:
@wilzbach New PR with the bugfix: https://github.com/mesonbuild/meson/pull/2943 - I will wait with a PR to enable Meson again in Vibe until this has been addressed on the Meson side and we have a new release that includes the bugfix.

PR #2046, opened 2018-01-25
Improve CodeCov setup - hide codecov/project

wilzbach:
See also: https://github.com/dlang/dub/pull/1347


wilzbach:
There are way too many random failures on AppVeyor:

image

s-ludwig:
We should update to the latest DUB, so that AppVeyor can benefit from the retry/fallback logic.

PR #2049, opened 2018-01-26
Correctly parse the host part of file URLs. Fixes #2048.

### LINKED ISSUE #2048: Host part of file URLs not handled correctly
Currently, `file://` URLs are treated as if they did not have a host portion, resulting in the wrong interpretation if the URL does not begin with `file:///`. Instead, the string between the second and the third slash should be interpreted as a host name.

s-ludwig:


s-ludwig:
Nice, I didn't know that that also ignores indentation changes.

(merging manually due to the version check error and will tag a new alpha in a moment)

PR #2051, opened 2018-01-26
Fix the HTTPServerOption.distribute deprecation message

wilzbach:
`HTTPServerOption.distribute` has been deprecated as it has no effect anymore.
This PR removes the last usage from Vibe.d and removes the deprecation message.


s-ludwig:
This reverts the change that I did shortly ago: bf5896e225f7853a18112ee32c690f084bbd0c81

Removing this immediately is not a true deprecation process. It should stay functional for at least another release to give library users time to switch to a different approach. The `HTTPServerOption.distribute` symbol itself should stay for a few more releases until the 0.7.x->0.8.x transition is reasonably finished (there are still active users of 0.7.x).

Issue #2054, opened 2018-02-01
assertion in tests/tcp


MartinNowak:
Yikes, no useful output because of `scope(failure) assert(false);` which gives a wrong backtrace.
https://github.com/vibe-d/vibe.d/blob/f64790b66a58dc85d61ca098fedb6bbef210b93d/tests/tcp/source/app.d#L20
Here is the failing test log on vibe.d-
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/master/871/pipeline#step-1023-log-443
```
[INFO] Running test tcp
+ cd tests/tcp
+ dub --compiler=dmd --override-config=vibe-d:core/libevent --build-mode=separate
Performing "debug" build using dmd for x86_64.
stdx-allocator 2.77.0: building configuration "library"...
vibe-d:utils 0.8.3-alpha.3: building configuration "library"...
vibe-d:data 0.8.3-alpha.3: building configuration "library"...
vibe-d:core 0.8.3-alpha.3: building configuration "libevent"...
../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
vibe-d:stream 0.8.3-alpha.3: building configuration "library"...
../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
tests ~master: building configuration "application"...
../../core/vibe/core/log.d(565,21): Deprecation: Built-in hex string literals are deprecated, use `std.conv.hexString` instead.
source/app.d(6,8): Deprecation: struct `std.datetime.StopWatch` is deprecated - Use std.datetime.stopwatch.StopWatch.
source/app.d(6,8): Deprecation: struct `std.datetime.StopWatch` is deprecated - Use std.datetime.stopwatch.StopWatch.
source/app.d(33,15): Deprecation: struct `std.datetime.StopWatch` is deprecated - Use std.datetime.stopwatch.StopWatch.
source/app.d(114,14): Deprecation: struct `std.datetime.StopWatch` is deprecated - Use std.datetime.stopwatch.StopWatch.
Linking...
Running ./tests
core.exception.AssertError@source/app.d(20): Assertion failure
----------------
??:? _d_assertp [0x639ff5]
source/app.d:20 void app.test1() [0x5791d2]
source/app.d:142 void app.test() [0x579cb4]
[...]

PR #2055, opened 2018-02-01
remove (pointless?) scope(failure) assert(false)

MartinNowak:
- hides any real backtrace
- test seem flaky (at least failed once on ci.dlang.org, see #2054)


s-ludwig:
Indeed seems useless...

PR #2063, opened 2018-02-06
On systems with secure arc4random_buf use it instead of /dev/urandom

n8sh:
Note: on the listed platforms `arc4random_buf` is not based on RC4 but on a modern and cryptographically secure cipher.


s-ludwig:
Makes sense!

PR #2069, opened 2018-02-13
Add writeTimeout for TCP connections

TETYYS:
Untested

### DIFF
--- a/core/vibe/core/drivers/libasync.d
+++ b/core/vibe/core/drivers/libasync.d
@@ -1138,7 +1138,7 @@ final class LibasyncTCPConnection : TCPConnection/*, Buffered*/ {
 	}
 
 	override @property Duration readTimeout() const { return m_settings.readTimeout; }
-	
+
 	override @property void writeTimeout(Duration dur)
 	{
 		m_settings.writeTimeout = dur;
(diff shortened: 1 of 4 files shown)


wilzbach:
Hi there,

Just a few quick comments (from my phone):

1) vibe-core is more or less deprecated. The vibe-core package is the default for master. So this would need to be added to that package too 
2) If you want this to be added, it should have tests. I know that other functionality might not have tests, but ask yourself this question: would you pull untested code?

Check `tests` for examples and ping us if you need help.

s-ludwig:
> vibe-core is more or less deprecated.

 typo: should be `vibe:core`

I agree that it doesn't make much sense to focus future work on `vibe:core`, but in general it can't hurt to have such functionality there, too, as I expect existing code to still rely on it for a while, until it gets finally removed. But `writeTimeout` needs to be added to `vibe-core`/`eventcore`, too, so that no accidental backwards-compatiblity holes get added. I can also do this when I get the time, but I'd hold off with pulling this until it is implemented there.

As for the tests, it should be noted that unfortunately timing behavior on the CI servers is often erratic, so the timeout must be expected to be at least a second or so off.

TETYYS:
I will try to add the timeout to vibe-core, but I'm not sure if I can test it. I mean, I know that such thing as write timeout exists, but what triggers it?

Geod24:
`vibe-core` now has write timeout (see https://github.com/vibe-d/vibe.d/pull/2344 and related).

PR #2073, opened 2018-02-17
Add Unix domain socket support for listenTCP/listenHTTP for libevent on Posix

lesderid:


core/vibe/core/drivers/libevent2.d
```diff
@@ -389,8 +389,29 @@ final class Libevent2Driver : EventDriver {
...
+		version(Posix)
+		{
+			import core.sys.posix.sys.un;
+			import core.stdc.string : strcpy;
+
+			if (address[0] == '/')
+			{
+				bind_addr.family = AF_UNIX;
+				sockaddr_un* s = bind_addr.sockAddrUnix();
+				enforce(s.sun_path.length > address.length, "Unix sockets cannot have that long a name.");
```
wilzbach: "that long a name" - it's generally good to be explicit in the error messages. The second part of `enforce` is lazy, so you can use things like `format`.

core/vibe/core/drivers/libevent2.d
```diff
@@ -389,8 +389,29 @@ final class Libevent2Driver : EventDriver {
...
+				s.sun_family = AF_UNIX;
+				() @trusted { strcpy(cast(char*)s.sun_path.ptr,address.toStringz()); } ();
```
wilzbach: The `toStringz` seems to be unnecessary here. Why don't you simply copy it over (e.g. `s[] = address[]`) and then set the next char to `\0`?

core/vibe/core/drivers/libevent2.d
wilzbach: Also this needs to be `>=` as `\0` takes up space too.

examples/unix_socket_server/dub.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "unit-socket-server-example",
+	"description": "A minimal HTTP server using Unix domain sockets",
+	"dependencies": {
+		"vibe-d:http": {"path": "../../"},
+		"vibe-d:web": {"path": "../../"},
+		"vibe-d:core": {"path": "../../"}
```
wilzbach: FWIW you would only need `vibe-d:web` -  it properly specified the dependencies on the other modules.

examples/unix_socket_server/source/app.d
```diff
@@ -0,0 +1,20 @@
+import vibe.inet.url;
+import vibe.http.server;
+import vibe.http.router;
+import std.stdio;
```
wilzbach: Use `vibe.core.log` for logging and in this case this looks like a left over.

examples/unix_socket_server/source/app.d
```diff
@@ -0,0 +1,20 @@
...
+
+shared static this()
+{
+	auto router = new URLRouter;
+	router.get("/hello", &handleHelloRequest);
+
+	auto settings = new HTTPServerSettings;
+	settings.bindAddresses = ["/tmp/vibe.sock"];
+
+	listenHTTP(settings, router);
```
wilzbach: Have a look at the other tests for examples on how you can verify the output of the server with `requestHTTP`

core/vibe/core/drivers/libevent2.d
lesderid: This is the same enforce and message that was used for the UDS implementation of requestHTTP.

core/vibe/core/drivers/libevent2.d
lesderid: Also copied from the UDS implementation of requestHTTP.

examples/unix_socket_server/source/app.d
lesderid: Yeah sorry, that's a leftover, will fix that.

examples/unix_socket_server/dub.json
lesderid: Hmm, I think I copied this from another example's dub.json. Might be worth taking a look at others to see if they have unnecessary dependencies as well then.

examples/unix_socket_server/source/app.d
lesderid: Will check that out, thanks.

core/vibe/core/drivers/libevent2.d
lesderid: `s.sun_path` is a `byte[]`. It could be changed to `() @trusted { s.sun_path[0 .. address.length] = (cast(byte[])address)[]; s.sun_path[address.length] = '\0'; } ();`, which is more or less [what Phobos does](https://github.com/dlang/phobos/blob/master/std/socket.d#L1988). Would that be better?

core/vibe/core/drivers/libevent2.d
lesderid: [Phobos doesn't seem to specify the length either](https://github.com/dlang/phobos/blob/master/std/socket.d#L1986), but I could change it to `enforce(s.sun_path.length > address.length, "Unix sockets cannot have names longer than %d characters.".format(s.sun_path.length - 1));`?

lesderid: It shouldn't have to be `>=`, `s.sun_path` is a `byte[]`.

examples/unix_socket_server/dub.json
lesderid: `vibe-d:core` is still required for the server example so the `libevent` configuration can be used for it, otherwise dub would say `The sub configuration directive "vibe-d:core" -> "libevent" references a package that is not specified as a dependency and will have no effect.`. Removing the `vibe-d:http` dependency from the server example and `vibe-d:web` from the client example though.

examples/unix_socket_server/source/app.d
lesderid: Hmm, I'm not sure how much sense it would make adding tests for this PR considering the client side (i.e. `requestHTTP`, in master since https://github.com/vibe-d/vibe.d/commit/33027eeb13c935cbcd437ba472c42cc50059ce22) for UDS doesn't have tests either.

lesderid:
@wilzbach `vibe-core` support seems trivial (`connectTCP` in `vibe-core` seems to be broken in master btw), will make a PR once the comments here are resolved.

core/vibe/core/drivers/libevent2.d
s-ludwig: AFAICS, the path must always be `\0` terminated regardless of the buffer length defined in `sun_path`, so using `>=` seems to be necessary indeed. The `stdcpy` below could otherwise also produce a buffer overflow when writing the zero byte.

core/vibe/core/drivers/libevent2.d
s-ludwig: Would be better in many cases because it always avoids a GC allocation, while `toStringz` needs to allocate if the string is not already zero terminated in memory for some reason (e.g. string literal or in a buffer that is padded with zeros).

examples/unix_socket_server/source/app.d
s-ludwig: Would surely be good to add a test that tests both sides at once with a simple request. But since it depends on vibe-core fixes to make it work there, too, I'd defer that, though and treat this topic WIP in the meantime.

core/vibe/core/drivers/libevent2.d
```diff
@@ -389,8 +389,29 @@ final class Libevent2Driver : EventDriver {
...
 	{
-		auto bind_addr = resolveHost(address, AF_UNSPEC, false);
-		bind_addr.port = port;
+		NetworkAddress bind_addr;
+		version(Posix)
+		{
+			import core.sys.posix.sys.un;
+			import core.stdc.string : strcpy;
+
+			if (address[0] == '/')
```
nemanja-boric-sociomantic: This disallows relative paths, is that necessary?

dd86k:
I'm interested in UNIX socket implementations for vibe.d and landed on this page. Is this still, or already a thing? Simply to be informed.

s-ludwig:
AFAICS, there are three open review comments that keep this from getting merged. @lesderid, if you don't have time to address them, I could probably take care of them in the near future, so that the remaining changes don't get lost here.

lesderid:
@s-ludwig I should have time to get back to this PR this week or the next. Sorry for waiting so long with this.

core/vibe/core/drivers/libevent2.d
lesderid: We're copying from `address` (`string`) to `sun_path` (`byte[somelength]`). Wouldn't `>` rather than `>=` make sure `sun_path` is large enough to hold `address` *and* `\0`?

core/vibe/core/drivers/libevent2.d
lesderid: No, but we couldn't simply check if it's a valid path because an IP address might also be a valid path.

@s-ludwig Any objections to using `std.path.isValidPath(address) && std.socket.getAddressInfo(address, AddressInfoFlags.NUMERICHOST).empty` here? Or is there a way to do this with `vibe.core`?

We could also just change it to `address.canFind('/')`, but then we wouldn't support Windows paths and paths below the working directory without `./`.

lesderid:
Is this PR still relevant? If so, could someone take a look at my review comments?

Ping @s-ludwig @wilzbach

PR #2079, opened 2018-02-22
preserve trailing back-slashes in NativePath

MartinNowak:
Slash behavior originally introduced here vibe-d/vibe.d@970a022#diff-3620332feace439ec637b023bc2e7a01R262, and apparently broken here vibe-d/vibe.d@344fe34#diff-47d6be40e29a186df00d5f3bfad92671R36.

The weird usage of Path for URLs and file paths on Windows makes this difficult to handle properly (though `/` and `\` usage should be mut-ex IIRC). No idea how much impact this breaking change would have. Any suggestion for a better migration path?

### DIFF
--- a/core/vibe/core/path.d
+++ b/core/vibe/core/path.d
@@ -131,6 +131,7 @@ struct Path {
 		m_nodes = splitPath(pathstr);
 		m_absolute = (pathstr.startsWith("/") || m_nodes.length > 0 && (m_nodes[0].toString().canFind(':') || m_nodes[0] == "\\"));
 		m_endsWithSlash = pathstr.endsWith("/");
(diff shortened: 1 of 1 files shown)


s-ludwig:
This is why there are three default path types in vibe-core (`WindowsPath`, `PosixPath` and `InetPath`). This conflation was a really bad idea, one that `std.path` also suffers from.

It should be `m_endsWithSlash |= pathstr.endsWith('\\');` though, so that a trailing slash is also still allowed on Windows.

MartinNowak:
> It should be m_endsWithSlash |= pathstr.endsWith('\\'); though, so that a trailing slash is also still allowed on Windows.

Of course :roll_eyes:, done.

Issue #2080, opened 2018-02-22
requestHTTP seems to have problems with IPv6 urls


MartinNowak:
```d
/++ dub.sdl:
dependency "vibe-d" version="~>0.8.3-alpha.4"
+/
import vibe.vibe;

void main()
{
    auto settings = new HTTPServerSettings;
    settings.port = 8080;
    settings.bindAddresses = ["::1"];
    listenHTTP(settings, (req, res) { res.writeBody("Hello, World!"); });

    runTask({
        auto resp = requestHTTP("http://[::1]:8080/");
        assert(resp.statusCode == 200, resp.toString);
    });
    runApplication();
}
```
Fails with `HTTP/1.1 400 Bad Request`.

PR #2081, opened 2018-02-22
run tests on ephemeral ports

MartinNowak:
- for obvious reasons
- dropped IPv6 binds as requestHTTP to IPv6 addresses doesn't work (#2080)
  (also IPv6 support should be tested once, not everywhere)

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -450,18 +450,18 @@ class Example7 : Example7API {
 }
 
 
-void runTests()
+void runTests(string url)
 {
 	// Example 1
 	{
-		auto api = new RestInterfaceClient!Example1API("http://127.0.0.1:8080");
+		auto api = new RestInterfaceClient!Example1API(url);
 		assert(api.getSomeInfo() == "Some Info!");
 		assert(api.getter == "Getter");
 		assert(api.postSum(2, 3) == 5);
 	}
 	// Example 2
 	{
-		auto api = new RestInterfaceClient!Example2API("http://127.0.0.1:8080", MethodStyle.upperUnderscored);
+		auto api = new RestInterfaceClient!Example2API(url, MethodStyle.upperUnderscored);
 		Example2API.Aggregate[] data = [
 			{ "one", 1, Example2API.Aggregate.Type.Type1 },
 			{ "two", 2, Example2API.Aggregate.Type.Type2 }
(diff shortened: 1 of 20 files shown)


wilzbach:
> for obvious reasons

Yeah, just a recent example:

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7939/5/pipeline

> dropped IPv6 binds as requestHTTP to IPv6 addresses doesn't work

Though there seems to be something wrong with vibe-core and the older libevent driver:

```
+cd tests/tcpproxy
...
Running ./tests 
Test protocol implementation on server
Test failed: Failed to connect to [0:0:0:0:0:0:0:0]:41441 (Cannot assign requested address)
Program exited with code 1
```

(all libasync jobs pass)

I tried to reproduce this locally, but failed.

s-ludwig:
The libasync runs pass only because the high level tests are not executed. The tcpproxy example is just missing an explicit "127.0.0.1" bind address (as well as some others maybe).

MartinNowak:
> +cd tests/tcpproxy ... Running ./tests Test protocol implementation on server Test failed: Failed to connect to [0:0:0:0:0:0:0:0]:41441 (Cannot assign requested address) Program exited with code 1

It's weird that listenTCP returns an IPv6 listener with addr 0 on Travis-CI which doesn't have IPv6 enabled. I've added a workaround and filtered the listeners for an IPv4 one.

MartinNowak:
Please rerun or ignore this job https://travis-ci.org/vibe-d/vibe.d/jobs/345425216.

wilzbach:
> It's weird that listenTCP returns an IPv6 listener with addr 0 on Travis-CI which doesn't have IPv6 enabled (filed as #2083

Especially given that some tests like ` tests/vibe.core.net.1726` which also use `listenTCP` pass fine :O 

> Please rerun or ignore this job https://travis-ci.org/vibe-d/vibe.d/jobs/345425216.

@s-ludwig any chance @MartinNowak can be added to team? 
(of course ideally we deploy https://github.com/dlang/installer/pull/273 and https://github.com/dlang/installer/pull/301 ...)

s-ludwig:
Sent an invitation

MartinNowak:
> Especially given that some tests like tests/vibe.core.net.1726 which also use listenTCP pass fine :O

That surprised me as well, the API is a bit weird, apparently it does listen on all interfaces and for IPv4 and IPv6 separately if you don't specify an address.
http://vibed.org/api/vibe.core.net/listenTCP
Other tests explicity select 127.0.0.1, so they don't run into the IPv6 bug.
https://github.com/vibe-d/vibe.d/blob/2f58e725197cf01a2a717cb40545aa5a7203b247/tests/vibe.core.net.1726/source/app.d#L41

s-ludwig:
Yeah, I'd really like to deprecate this and the `["::", "0.0.0.0"]` default for `HTTPServerSettings.bindAddresses`. It's legacy code that doesn't make sense at all today.

wilzbach:
FYI I tagged `v0.8.3-alpha.5`, s.t. this can be used on the Project Tester.

> Yeah, I'd really like to deprecate this and the ["::", "0.0.0.0"] default for HTTPServerSettings.bindAddresses. It's legacy code that doesn't make sense at all today.

Cool! Let's give this a try:

https://github.com/vibe-d/vibe.d/pull/2086
https://github.com/vibe-d/vibe.d/pull/2087

dejlek:
I disagree. Binding to all available interfaces makes more sense than binding to the 127.0.0.1.

PR #2085, opened 2018-02-24
Convert (to/from)JSONValue to opCast/constructor.

s-ludwig:
This allows other similar conversions to be implemented later, in particular for `stdx.data.json`. It also enables `json.to!JSONValue` and `jsonvalue.to!JSON` conversions.

See #1904.


s-ludwig:
@WebFreak001 @wilzbach FYI

WebFreak001:
:+1: too bad there are no implicit cast operators in D

PR #2086, opened 2018-02-25
Change default bindAddress to 127.0.0.1

wilzbach:
An alternative if you want people to be explicit would be to use a empty bind address, check for it in `listenHTTP`, still set the old bindAddress, but trigger a "deprecation message" to the log.

Arguments for 127.0.0.1
-----------------------

- it's the sane default. A user should explicitly expose his server to the world
- `listenHTTP(":8080") is a convenient shorthand for which Vibe.d should just listen on localhost

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -163,7 +163,7 @@ unittest
 	void test()
 	{
 		static void testSafeFunction(HTTPServerRequest req, HTTPServerResponse res) @safe {}
-		listenHTTP("0.0.0.0:8080", &testSafeFunction);
+		listenHTTP("127.0.0.1:8080", &testSafeFunction);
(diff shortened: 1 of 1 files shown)


dejlek:
Why is binding to localhost a better alternative than binding to any available interface?? Imagine your CI/CD pipeline includes generation of a docker image, you run it, and want to test... - You will not be able to use the 127.0.0.1

wilzbach:
> Why is binding to localhost a better alternative than binding to any available interface?? 

Because you don't want to expose your application to the __public__ interface by default.
That should be a conscious choice by the programmer.

>  However, if the code by default binds to any interface, you should be able to connect to your docker container.

Yep, and that should be a conscious choice as otherwise it's really easy to accidentally expose your application to the world.

PR #2093, opened 2018-03-01
Use getrandom syscall on linux if available

n8sh:
Based on code originally written by @wilzbach
libmir/mir-random#13


s-ludwig:
To avoid last minute breaking I'd defer the merge until after the 0.8.3 release. In two years or so we should probably start thinking about switching to `getrandom` from `sys/random.h`.

s-ludwig:
Should be "to avoid *the possibility* of last minute breaking"

WebFreak001:
*ping* 0.8.3 has been released

PR #2115, opened 2018-03-11
Download DUB from GitHub instead of code.dlang.org.

s-ludwig:


wilzbach:
Looks like AppVeyor is using Openssl 1.1

s-ludwig:
The problem appeared to be PowerShell not using TLS 1.2 by default, and GitHub requiring it. Now the question is why the LDC download works by default...

PR #2121, opened 2018-03-14
Fix openssl cert peer validation

marcioapm:
This is for 0.7.32 and is important due to the imminent european GDPR deadline...
The same bug is also present in master.

### DIFF
--- a/source/vibe/stream/openssl.d
+++ b/source/vibe/stream/openssl.d
@@ -898,7 +898,7 @@ private bool verifyCertName(X509* cert, int field, in char[] value, bool allow_w
 	}
 
 	X509_NAME* name = X509_get_subject_name(cert);
-	int i;
+	int i = -1;
 	while ((i = X509_NAME_get_index_by_NID(name, cnid, i)) >= 0) {
 		X509_NAME_ENTRY* ne = X509_NAME_get_entry(name, i);
 		ASN1_STRING* str = X509_NAME_ENTRY_get_data(ne);


wilzbach:
> lastpos should initially be set to -1. If there are no more entries -1 is returned.

From https://linux.die.net/man/3/x509_name_get_index_by_nid

So shouldn't the while loop in general be kept?

s-ludwig:
Think so, too. AFAICT the mistake is that `i` is not initialized to `-1`.

marcioapm:
Fixed!

s-ludwig:
Thanks! I'll merge it to master, too.

marcioapm:
This is quite important for GDPR compliance... are you able to do another 0.7.x release?

s-ludwig:
> are you able to do another 0.7.x release?

Yes, I can do that. If you see any fixes in the change log for the 0.8.2 or 0.8.3 releases that look important, I could possibly merge those, too, to make the release more valuable.

marcioapm:
I will have a look and let you know! Thanks Sönke! :)

wilzbach:
I guess this should be tagged as 0.8.4 or 0.8.3-1 then too?

marcioapm:
@s-ludwig I found the following commits that would be immediately valuable for us.
I also hope it compiles with latest DMD - we are still on 0.76.1 but thinking about upgrading soon.

*important*
fix leakage of connections
https://github.com/vibe-d/vibe.d/commit/ab5ebe9e08f76e35f3430ed336131d52c049a2c1

Fix support for outbound SSL w/ SSL1.1.x
https://github.com/vibe-d/vibe.d/commit/4b0466e558abfc6bcd23b0d809d8547bfce064c3

*good to have*
throw useful error message on MongoDB query failures
https://github.com/vibe-d/vibe.d/commit/31281f21c8dc7c89e2eaa6ab0c96be83e1cea450

Properly handle TLS disconnects in OpenSSLStream.leastSize.
https://github.com/vibe-d/vibe.d/commit/1c55fb0409245144264e0f68c01910b5c7506410

Use non-GC memory for MatchGraphBuilder.
https://github.com/vibe-d/vibe.d/commit/e7a0f30204d0668168409f5c6bc2f08f5bec95b2

Use small integer types for MatchGraph(Builder) to reduce memory usage.
https://github.com/vibe-d/vibe.d/commit/396400bbcf9f43ac44e0e85a8884e8cb97aa7459

*good if compatible*
Fix closing of user-triggered non-keepalive connections.
https://github.com/vibe-d/vibe.d/commit/2a768b2f7b299047e9accb3d6959fa7917c9eaa5

Thanks Sönke.

s-ludwig:
Opened corresponding PRs: #2125, #2126, #2127, #2128, #2129

PR #2125, opened 2018-03-18
Fix serialization test for the latest DMD versions.

s-ludwig:


wilzbach:
This is still failing due to IPv6 listening errors on Travis. I think https://github.com/vibe-d/vibe.d/pull/2129 tries to fix that too.

PR #2130, opened 2018-03-22
Replace deprecated std.exception.enforceEx with enforce

wilzbach:
With 2.079 and higher `enforce` is a strict super set of `enforce`.


core/vibe/core/drivers/utils.d
```diff
@@ -7,7 +7,7 @@
 */
 module vibe.core.drivers.utils;
 
-import std.exception;
+import vibe.internal.exception : enforce, ErrnoException;
```
s-ludwig: Errors out for `ErrnoException` here.

wilzbach: Fixed.

wilzbach:
BTW should we create a special "compat" module into which this entire compatibility layer over Phobos goes?
I guess this would make it easier to keep track of the removing old compatibility workarounds and would create a "go-to" place when compatibility is required.
tl;dr: it felt a bit wrong to create `utils/vibe/internal/exception.d` and I think `utils/vibe/internal/compat.d` might be better.

PR #2135, opened 2018-03-29
reduce hash calculations in json opIndex

Boris-Barboris:
This small fix improves efficiency of Json object field assignment by reusing an AA  bin pointer.


Boris-Barboris:
@s-ludwig all passed.
Btw, do you mind such poking?

PR #2136, opened 2018-03-30
Add RestInterfaceClient.requestBodyFilter property.

s-ludwig:
This can be used to modify outgoing request headers based on the content of the request body.

### DIFF
--- a/web/vibe/web/rest.d
+++ b/web/vibe/web/rest.d
@@ -220,7 +220,7 @@
 		To see how to implement the client side in detail, jump to
 		the `RestInterfaceClient` documentation.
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
@Extrawurst Didn't really find the time to think about an optimal solution, so I just went with an `InputStream`.

Extrawurst:
@s-ludwig That is still awesome! Thanks! How about tagging a beta after this is merged? :)

s-ludwig:
Will do, I'll just make it an alpha for now.

s-ludwig:
0.8.4-alpha.1 is now available.

PR #2137, opened 2018-04-02
Make openssl-1.1 the new default

wilzbach:
Always using `--override-config="vibe-d:tls/openssl-1.1"` is really annoying and
OpenSSL 1.1 was released almost two years ago, so it's available on all major Linux distros.

https://www.openssl.org/news/openssl-1.1.0-notes.html

Even Debian (!) transitioned to OpenSSL 1.1:

https://wiki.debian.org/OpenSSL-1.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,15 +210,15 @@ Optionally, run `./setup-freebsd.sh` to create a user/group pair for privilege l
 Switching between OpenSSL versions
 ----------------------------------
 
-By default, vibe.d is built against OpenSSL 1.0.x. On systems that use the newer
-1.1.x branch, this can be overridden on the DUB command line using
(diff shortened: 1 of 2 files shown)


s-ludwig:
The latest Ubuntu LTS AFAIK still uses 1.0.2, which presumably makes up a major portion of the user base. Of course 18.04 will be out shortly, followed by derived distributions a few months later, so this issue should gradually vanish.

wilzbach:
Hmm as Travis still uses Trusty, it looks like we can't default to OpenSSL 1.1 for the foreseeable future.
However, we should be able to auto-detect it: https://github.com/vibe-d/vibe.d/pull/2190

PR #2140, opened 2018-04-05
Install the newest version of dub to avoid random network failures

wilzbach:
Used a forked version of the installer script as the Travis failures get super-annoying and https://github.com/dlang/installer/pull/301 isn't moving anywhere.
Once it's merged, these lines could be dropped and replaced with e.g. `- dmd-2.073.1+dub` (2.073.1 + newest dub compiler)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,10 @@ matrix:
 before_install:
   - pyenv global system 3.6
   - pip3 install 'meson>=0.44.1'
+  # Use the dub-updating fork of the installer script until https://github.com/dlang/installer/pull/301 is merged
+  - wget https://raw.githubusercontent.com/wilzbach/installer-dub/master/script/install.sh -O ~/dlang/install.dub.sh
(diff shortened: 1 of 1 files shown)


wilzbach:
Doesn't this look nice?

image

s-ludwig:
A dream :) It looks almost suspiciously green.

The unrelated AppVeyor failure looks strange:
> Error: MS-Coff object module affix_allocator.obj has magic = 5ac6, should be 8664

These CI virtual machines sometimes really behave awkwardly.

wilzbach:
> A dream :) It looks almost suspiciously green.

Well it definitely gets the newest DUB:

image

Maybe we were just luckily and at night no one was pushing the registry into its restarts...
On Jenkins we had to disable quite a few tests to avoid random failures:

https://github.com/dlang/ci/blob/master/vars/runPipeline.groovy#L161

So I wouldn't have too many hopes, but this should at least eliminate the main source of our CI troubles.
I will do similar PRs for vibe-core + DUB soon.

PR #2145, opened 2018-04-11
MongoDB: adds max pool size setting

denizzzka:


mongodb/vibe/db/mongo/settings.d
```diff
@@ -321,6 +321,7 @@ class MongoClientSettings
 
 	string username;
 	string digest;
+	uint max_connections = uint.max;
```
denizzzka: It is because `uint.max` is default value for ConnectionPool ctor argument `max_concurrent`

denizzzka:
(Isn't know why codecove fails)

PR #2146, opened 2018-04-21
Fix issue with building on dmd 2.079

schveiguy:
The import wasn't properly included. Was fixed in the master branch, but not the 0.7.x branch.

Identical to #1861


schveiguy:
Thx!

PR #2147, opened 2018-04-26
Fix DMD 2.080.0-beta.1 compilation.

s-ludwig:


s-ludwig:
> FYI we had to temporarily remove a lot of projects from the project tester as there were some issues with the dub registry again :/

Oh really? Interesting, I see 100% uptime, with a few seconds (one failure) per day. But I guess the fallbacks in DUB still didn't completely work then?

PR #2149, opened 2018-05-01
Ignore optional undefined Json fields

gedaiu:


s-ludwig:
I was thinking about this and think that fixing this in `readDictionary` would solve a few additional cases that currently won't work as expected. So instead of handling `undefined` in `tryReadNull`, `readDictionary` would simply skip any `undefined` fields.

Similarly, `readArray` *could* also be patched to replace any `undefined` with `null` when setting `m_current`, to match the JavaScript JSON conversion semantics.

s-ludwig:
Thanks!

PR #2151, opened 2018-05-03
Fix serialization of const/mutable and wide character strings

### LINKED ISSUE #2150: std.data.json serializes non-immutable string as array of characters
```D
import std.stdio;
import vibe.data.json;

struct Foo
{
    const(char)[] name;
}

void main()
{
    Foo f = Foo("Hello world");
    writeln(f.serializeToJsonString());
}
```

Result
```
{"name":["H","e","l","l","o"," ","w","o","r","l","d"]}
```

CC @christoph-wasicki-sociomantic

s-ludwig:
Fixes #2150.


Geod24:
Thanks!

PR #2165, opened 2018-05-16
Fix bogus mailto links generated by the Markdown parser.

s-ludwig:
Whenever an "at" character was in an auto URL, the contents were treated as an e-mail address. This was even the case for URLs such as "http://foo/bar@baz".

### DIFF
--- a/textfilter/vibe/textfilter/markdown.d
+++ b/textfilter/vibe/textfilter/markdown.d
@@ -1090,19 +1090,55 @@ pure @safe {
 
 private bool parseAutoLink(ref string str, ref string url)
 pure @safe {
+	import std.algorithm.searching : all;
+	import std.ascii : isAlphaNum;
+
 	string pstr = str;
-	if( pstr.length < 3 ) return false;
-	if( pstr[0] != '<' ) return false;
+	if (pstr.length < 3) return false;
+	if (pstr[0] != '<') return false;
 	pstr = pstr[1 .. $];
 	auto cidx = pstr.indexOf('>');
-	if( cidx < 0 ) return false;
+	if (cidx < 0) return false;
+
 	url = pstr[0 .. cidx];
-	if( anyOf(url, " \t") ) return false;
-	if( !anyOf(url, ":@") ) return false;
+	if (url.anyOf(" \t")) return false;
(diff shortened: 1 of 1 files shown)


textfilter/vibe/textfilter/markdown.d
```diff
@@ -1091,19 +1091,45 @@ pure @safe {
...
-	if( pstr[0] != '<' ) return false;
+	if (pstr.length < 3) return false;
+	if (pstr[0] != '<') return false;
 	pstr = pstr[1 .. $];
 	auto cidx = pstr.indexOf('>');
-	if( cidx < 0 ) return false;
+	if (cidx < 0) return false;
 	url = pstr[0 .. cidx];
-	if( anyOf(url, " \t") ) return false;
-	if( !anyOf(url, ":@") ) return false;
```
wilzbach: ":@" is now valid.

textfilter/vibe/textfilter/markdown.d
```diff
@@ -1091,19 +1091,45 @@ pure @safe {
...
+			assert(outurl == url);
+			assert(strcpy.length == 0);
+		}
+	}
+
+	test(true, "<http://foo/>", "http://foo/");
+	test(false, "<http://foo/", null);
+	test(true, "<mailto:foo@bar>", "mailto:foo@bar");
+	test(true, "<foo@bar>", "mailto:foo@bar");
+	test(true, "<proto:foo@bar>", "proto:foo@bar");
```
wilzbach: Maybe a test that checks for : after @?
e.g. when a custom port is used

textfilter/vibe/textfilter/markdown.d
s-ludwig: I don't think anything changes related to this condition, because it is replicated below as `atidx < 0 && colonidx < 0`. ":@" will not be accepted due to the length check at the top, but indeed this will let a lot of invalid cases go through. True validation has to be done as a separate step

textfilter/vibe/textfilter/markdown.d
s-ludwig: Added, plus allowed `"foo:foo"@bar` style e-mail addresses.

PR #2167, opened 2018-05-28
Fix #1780

### LINKED ISSUE #1780: HTTPClientResponse body without Content-Length
A came across an embedded server (Axis camera) which returns this:

```
HTTP/1.0 200 OK
Keep-Alive: timeout=2, max=100
Connection: Keep-Alive
Allow: OPTIONS, GET, HEAD, POST, PUT, DELETE, TRACE
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: OPTIONS, GET, HEAD, POST, PUT, DELETE, TRACE
Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept
Content-Type: text/xml
Status: 200 OK

<events>
....
</events>
```

And with vibe-d, the response body cannot be read (it creates the stream with limited size set to 0).
It seems to be a problem with combination of Connection: Keep-Alive (which is returned even if not requested) and missing Content-Length - so probably a bug on the server side.

But with std.net.curl it is readed fine, and for example with VSCode REST Client extension too.
Is there some way to read it with vibe?

tchaloupka:
Tried to fix this. It works ok for me with this change.

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -1119,7 +1119,12 @@ final class HTTPClientResponse : HTTPResponse {
 	private @property isKeepAliveResponse()
 	const {
 		string conn;
-		if (this.httpVersion == HTTPVersion.HTTP_1_0) conn = this.headers.get("Connection", "close");
+		if (this.httpVersion == HTTPVersion.HTTP_1_0) {
+			// Workaround for non-standard-conformant servers - for example see #1780
+			auto pcl = "Content-Length" in this.headers;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Looks like a workable solution, I'd just mention in a comment that this is a workaround for non-standard-conformant servers and #1780 in particular.

tchaloupka:
Thx, adjusted the comment

s-ludwig:
Okay, thanks! Set to merge.

PR #2183, opened 2018-07-06
fix Bson.opEquals: order of object keys is not determined

IgorStepanov:
Subj. This PR is needed for [druntime#2243](https://github.com/dlang/druntime/pull/2243)
If it is possible, make a new tag if this will be approved, please.

### DIFF
--- a/data/vibe/data/bson.d
+++ b/data/vibe/data/bson.d
@@ -754,12 +754,28 @@ struct Bson {
 	///
 	bool opEquals(ref const Bson other) const {
 		if( m_type != other.m_type ) return false;
-		return m_data == other.m_data;
+		if (m_type != Type.object)
+			return m_data == other.m_data;
+
+		if (m_data == other.m_data)
+			return true;
+		// Similar objects can have a different key order, but they must have a same length
+		if (m_data.length != other.m_data.length)
+			return false;
+
+		foreach (k, ref v; this.byKeyValue)
+		{
+			if (other[k] != v)
+				return false;
+		}
+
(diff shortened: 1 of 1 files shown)


IgorStepanov:
Looks like that fail unrelated with this PR.

IgorStepanov:
Ping

wilzbach:
Ping @s-ludwig - are you okay with this?
BTW doesn't suffer the Json implementation from the same flaw?

wilzbach:
(going to merge this if there are no objections within the next 72h)

wilzbach:
> I wonder what the correct approach here is, or whether there is one (the spec doesn't seem to be helpful). It could be argued that the order is part of the semantics. I

The spec is pretty clear on that:

Implementation Defined: The built-in associative arrays __do not__ preserve the order of the keys inserted into the array. In particular, in a foreach loop the order in which the elements are iterated is typically unspecified.

https://dlang.org/spec/hash-map.html

Of course, it might make sense for Bson to have a defined order, but you can't rely on built-in arrays for this.

IgorStepanov:
@s-ludwig @wilzbach ping.

@s-ludwig, If we want to create `Bson` implementation  which preserve fields order, we should rewrite `Bson` and don`t use associative array anywhere. This PR fixes univocal trouble of current implementation.

wilzbach:
As releasing a new version is a bit more difficult for Vibe.d, I set the version to be tested at the project tester to master: https://github.com/dlang/ci/pull/241

PR #2187, opened 2018-07-14
In vibe.utils.hashmap fix equals for Object keys and fix special case for Object.toHash

n8sh:
**EDIT: The description is talking about DefaultHashMapTraits.hashOf since that's what the PR was originally about.**

Before the condition was improperly written (`&Object.init.toHash is &T.init.toHash` instead of `&Object.toHash is &T.toHash`) so it never triggered:

https://run.dlang.io/is/yLG7gs
```d
void main()
{
    static assert(&Object.toHash is &Object.toHash); // Passes.
    static assert(&Object.init.toHash is &Object.init.toHash); // Doesn't pass!
}
```

But it's a good thing the condition didn't trigger because it also didn't check that the class was `final`:

https://run.dlang.io/is/BFkjFJ
```d
void main()
{
    static class C1 {}
    static class C2 : C1 { override size_t toHash() { return 4; } }
    static assert(&C2.toHash !is &Object.toHash);
    static assert(&C1.toHash is &Object.toHash); // Uh-oh! Passes!
}
```

This PR fixes both problems so the special case now works as intended. Also the use of `Unqual` was unnecessary so it was removed.

**__EDIT:__ Also fixes `DefaultHashMapTraits.equals` for classes that override `opEquals`.**

### DIFF
--- a/core/vibe/core/drivers/libevent2.d
+++ b/core/vibe/core/drivers/libevent2.d
@@ -693,7 +693,20 @@ struct ThreadSlot {
 	ArraySet!Task tasks;
 }
 /// private
-alias ThreadSlotMap = HashMap!(Thread, ThreadSlot);
+struct ThreadSlotMapTraits {
+	enum Thread clearValue = null;
+	// This relies on no subclass of Thread overriding opEquals or toHash.
+	static bool equals(in Thread a, in Thread b)
+	@nogc nothrow pure @safe {
+		return a is b;
+	}
+	static size_t hashOf(in Thread a)
+	@nogc nothrow pure @trusted {
+		return cast(size_t) cast(const void*) a;
+	}
+}
+/// private
+alias ThreadSlotMap = HashMap!(Thread, ThreadSlot, ThreadSlotMapTraits);
 
 final class Libevent2ManualEvent : Libevent2Object, ManualEvent {
 @safe:
--- a/utils/vibe/utils/hashmap.d
+++ b/utils/vibe/utils/hashmap.d
@@ -15,15 +15,42 @@ import std.traits;
 
 struct DefaultHashMapTraits(Key) {
 	enum clearValue = Key.init;
-	static bool equals(in Key a, in Key b)
+
+	static if (__traits(isFinalClass, Key) && __traits(compiles,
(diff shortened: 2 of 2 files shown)


wilzbach:
> static assert(&C1.toHash is &Object.toHash); // Uh-oh! Passes!

Why is this a problem? If the user upcasts his object to C2, it's expected that he looses the `toHash` specialization too, no?

FWIW after looking at this special case for a bit and the respective druntime implementation (https://github.com/dlang/druntime/blob/v2.081.1/src/object.d#L871), I don't know why the special case even exists.
It looks to me like that's druntime's `Object` default, anyhow? Maybe this special case is a relict from old times and it's better to fully remove it?

n8sh:
>Why is this a problem?

It's a problem because if `C2` is a subclass of `C1` then it is valid to use `C2` objects as keys in a hashmap that expects its keys to be `C1`.

>It looks to me like that's druntime's `Object` default

I think that's the idea. When they know what would be done, they skip the `typeid` song and dance.

n8sh:
Yikes, `DefaultHashMapTraits.equals` is also wrong for any class that override `opEquals`. Here's an example that demonstrates the failure:

```d
/+dub.sdl:
dependency "vibe-d" version="~>0.8.4"
+/
final class String
{
    immutable char[] value;

    this(string s) { value = s; }
    this(scope const char[] s) { value = s.idup; }
    override string toString() const @nogc nothrow pure @safe { return value; }    
    override size_t toHash() const @nogc nothrow pure @safe { return .hashOf(value); }
    override bool opEquals(scope const Object rhs) const @nogc nothrow pure @safe {
        auto s = cast(const String) rhs;
        return s !is null && value[] == s.value[];
    }
}

void main()
{
    import vibe.utils.hashmap : HashMap;
    
    String a = new String("abc");
    String b = new String(a.value);
    assert(a == b); // Keys are equal.
    assert(a.toHash() == b.toHash()); // Keys have same hash code.
    
    auto ht = new HashMap!(String, int)();
    ht.opIndexAssign(5, a);
    assert(ht.get(a) == 5);
    assert(ht.get(b) == 5, "This fails!");
}
```

PR #2197, opened 2018-08-09
Fix RestInterfaceSettings.dup() for errorHandler callback

vitalka200:
Added missing fields copy in RestInterfaceSettings.dup().

Without this, setting `errorHandler` callback will not work because
of this in `module vibe.web.internal.rest.common:L118`
```D
this.settings = settings ? settings.dup : new RestInterfaceSettings;
```


vitalka200:
@s-ludwig @wilzbach When it can be merged?

Issue #2202, opened 2018-08-15
Introduce @wholeBodyParam UDA and support for it in the vibe.d's web framework.


dbankov-vmware:
While working on a web service based on vibe.d targeted at supporting an iPXE client I got into a situation where it would be best for me to be able to map all parameters provided in a POST request body coming from iPXE and encoded using the HTTP Form protocol into a single struct parameter. I saw that this was briefly discussed in #1676 and agree that this shouldn't happen implicitly so I think it should be possible to expose the already existing WebParamAttribute(PrameterKind.wholeBody, ..., ...) instance as a supported UDA for methods registered through the registerWebInterface function.

I'm posting this issue more to start a discussion on this topic but also have an example (rudimentary) implementation that I'll share in a pull request related to this issue shortly.

Please let me know what you think about this proposal so to decide if I should look deeper into implementing support for it.

BTW it seems the @wholeBodyParam UDA will work as expected for the Rest framework without touching anything there.

PR #2203, opened 2018-08-15
Fix for #2202. Introduce support for @wholeBodyParam UDA in vibe.d's web framework.

dbankov-vmware:
This is a preliminary (without tests and documentation) pull request just to illustrate a possible implementation of the fix for #2202.

Please comment if there is a better way to implement this and also if it is worth trying to fix this in the first place.

### DIFF
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -533,6 +533,14 @@ WebParamAttribute bodyParam(string identifier)
 	return WebParamAttribute(ParameterKind.body_, identifier, "");
 }
 
+WebParamAttribute wholeBodyParam(string identifier) @safe
+{
+	import vibe.web.internal.rest.common : ParameterKind;
+	if (!__ctfe)
+		assert(false, onlyAsUda!__FUNCTION__);
+	return WebParamAttribute(ParameterKind.wholeBody, identifier, "");
+}
+
 /**
  * Declare that a parameter will be transmitted to the API through the headers.
  *
@@ -686,7 +694,12 @@ package enum ParamResult {
 }
 
 // NOTE: dst is assumed to be uninitialized
-package ParamResult readFormParamRec(T)(scope HTTPServerRequest req, ref T dst, string fieldname, bool required, NestedNameStyle style, ref ParamError err)
+package ParamResult readFormParamRec(T)(scope HTTPServerRequest req,
+													 ref T dst, string fieldname,
+													 bool required,
+													 NestedNameStyle style,
+													 ref ParamError err,
+													 bool wholeBodyParam)
 {
 	import std.traits;
 	import std.typecons;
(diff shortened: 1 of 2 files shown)


Geod24:
Yeah I think having `@bodyParam` only is better.
But it seems to come up a lot that people don't want the second argument and get confused by it.
One thing we should do is better document it. Another could be to deprecate the second argument, given its probably not widely used (and easy to replace with just a struct containing a struct).

dbankov-vmware:
Hi Wilzbach,

The thing is @bodyParam("params") didn't work for me for a method from the web interface i.e. in this case the web framework looks for params_<member name> in the form parameters and seems there is no support to match the member name directly (without the prefix). And I though this was intentional based on the comments in #1676 and having in mind that @bodyParam had worked this way for a long time it seemed that changing this behaviour is undesired. Because of the latter I came to the  conclusion that the cleanest way to support what I needed was to introduce a new UDA.

If it is better to only have the @bodyParam I think I can easily make @bodyParam("params") to work for the web framework the same way as it works for the Rest framework but as I said above this seemed like a breaking change to me.

Best regards,
Dentcho Bankov

wilzbach:
> The thing is @bodyParam("params") didn't work for me for a method from the web interface 

Ah sorry. I confused that #1723 implemented this in master only for #1723 (and my own vibe.d fork actually does support this - similar to what I proposed in https://github.com/vibe-d/vibe.d/pull/1676).

> Because of the latter I came to the conclusion that the cleanest way to support what I needed was to introduce a new UDA.

I'm not sure introducing yet another UDA is a good idea - especially given that `@bodyParam("foo") already works for the REST part, so it will probably be a bit unintuive for users.

> If it is better to only have the @bodyParam I think I can easily make @bodyParam("params") to work for the web framework the same way as it works for the Rest framework but as I said above this seemed like a breaking change to me.

I don't see how this would be a breaking change, because currently `@bodyParam` isn't supported at all for the web interfaces, so it will only affect new code, no?

> But it seems to come up a lot that people don't want the second argument and get confused by it.
One thing we should do is better document it.

Yeah I saw a few issues that people opened because of this too. Though I think the missing/scarce documentation problem won't be magically be solved by `wholeBodyParam`.

>  Another could be to deprecate the second argument, given its probably not widely used (and easy to replace with just a struct containing a struct).

+1 (but that's something for a different PR).

dbankov-vmware:
Hi Sebastian,

In that case could you please share your implementation so to see if it works for my case and if so I'd like to try to push it through. If indeed it is not a breaking change this is what makes the most sense to me.

Best regards,
Dentcho Bankov

wilzbach:
> In that case could you please share your implementation so to see if it works for my case

Sorry, it's not on GitHub yet, but it looks similar to what https://github.com/vibe-d/vibe.d/pull/1676 did for the REST part.

> If indeed it is not a breaking change this is what makes the most sense to me.

The breaking change proposed in https://github.com/vibe-d/vibe.d/pull/1676 would have been to automatically serialize the body into a parameter without attributes. As long as we still use `@bodyParam` I don't see any problem.

> With 2.082 we will have UDAs for function attributes, so it can be:
> void postFoo(@bodyParam MyStruct foo) {}

FYI:

https://dlang.org/changelog/2.082.0.html#uda-function-arguments

wilzbach:
>>  In that case could you please share your implementation so to see if it works for my case
> Sorry, it's not on GitHub yet, but it looks similar to what #1676 did for the REST part.

I did find an older version of this on GH though. In `hb-web`, Path parameters with start with `_` automatically get injected and everything which has no attribution for PUT/POST requests gets serialized as whole body if it's the first parameter. It's a collection of a few hacks:

https://github.com/teamhackback/hb-web/blob/master/source/hb/web/web.d#L772

PR #2206, opened 2018-08-22
vibe.data.json: switch back to union for possible Json values

FeepingCreature:
Fix Issue 2205? The crashes haven't happened since.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,7 +17,7 @@ environment:
     DVersion: 2.088.1
     arch: x64
   - DC: dmd
-    DVersion: 2.083.1
+    DVersion: 2.084.1
     arch: x86_mscoff
   - DC: ldc
     DVersion: 1.20.1
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -116,7 +116,6 @@ import std.uuid;
 	a JSONException. Additionally, an explicit cast or using get!() or to!() is
 	required to convert a JSON value to the corresponding static D type.
 */
-align(8) // ensures that pointers stay on 64-bit boundaries on x64 so that they get scanned by the GC
 struct Json {
 @safe:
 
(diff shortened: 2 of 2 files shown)


FeepingCreature:
I don't get that error...

FeepingCreature:
skip the assert for json equality with m_object at compiletime, since it confuses the compiler now apparently

FeepingCreature:
ping

FeepingCreature:
I want to emphasize again that this commit is the difference between "hard hang in json code" and "no problems".

**Ping!**

FeepingCreature:
resurrecting this pr as per https://github.com/vibe-d/vibe.d/pull/2525

FeepingCreature:
Ah, turns out having `ubyte[intEnum]` in a union fucks you over when it comes to CTFE. Switched to a `{size_t, size_t, size_t}` struct.

FeepingCreature:
uh.

ping??

s-ludwig (approved):
Okay, so there is one error left for the oldest DMD/x86. If there is no obvious workaround for that, you can try to replace DMD 2.083.x with 2.084.x or 2.085.x, since that still falls within the "latest 10 minor releases" of support. Otherwise LGTM!

BTW, did you get the chance to verify the `zeroFields` workaround in production?

Yeah, and sorry letting this hang in limbo for so long. It looks safe enough now to make the switch without introducing regressions.

FeepingCreature:
updated version to 2.084.1. I didn't test this specific version in production; the version we're deploying is still the one with the void*[] hack. I can switch to this one though, if you want more data.

s-ludwig:
> updated version to 2.084.1. I didn't test this specific version in production; the version we're deploying is still the one with the void*[] hack. I can switch to this one though, if you want data on effectiveness.
> (You'll forgive me for not deliberately leaking ram on a production system for comparison, I hope. :) )

Sure, just validating that the new code really doesn't introduce a leak is definitely enough here ;) I'll also just merge it now, but it would be good if you would get the chance to try it out under controlled circumstances before a new release is tagged. I can do the same with the DUB registry, which is the most loaded service that I currently run. The leaks were showing through there pretty quickly if I remember right.

FeepingCreature:
Ah, my genuine apologies. I was unable to convince folks to switch from a version that we think works to one that is not fully proven over the short term. I'll try to merge the union-based version into mainline and test it as part of a future release, but we will only be able to confirm it fixes the bug without leaking in the medium future, ie. weeks to months.

s-ludwig:
I'll test with the dub-registry after I have collected some before-statistics. From my side, if that succeeds, that will be good enough to risk a release.

s-ludwig:
Okay, it's running for a few hours now and so far no change in memory consumption. I'll keep monitoring it, but it really should have shown by now. So it looks like this is fine!

PR #2213, opened 2018-09-25
trusted methods for openssl alpn

GallaFrancesco:
The current implementation doesn't compile when ALPN is enabled (with openssl version >= 1.0.2 or `version=VibeForceALPN`) since `@safe` methods are not allowed to manipulate `extern(C)` pointers and ubyte arrays, causing compile-time errors.

The changes proposed wrap the affected statements in `@trusted` blocks so that the manipulation is allowed.

### DIFF
--- a/tls/vibe/stream/openssl.d
+++ b/tls/vibe/stream/openssl.d
@@ -398,7 +398,7 @@ final class OpenSSLStream : TLSStream {
 			char[32] data;
 			uint datalen;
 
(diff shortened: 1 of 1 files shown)


GallaFrancesco:
Ok, as for the previous PRs I already rebased the fixup commit into the previous one. Thanks!

s-ludwig:
Thanks! CI failures are unrelated, so merging manually.

Issue #2215, opened 2018-10-01
dmd 2.081.0 build fails


Extrawurst:
looks like currently vibe does not keep up being compatible with current dmd versions anymore. dmd 2.080.1 is the last supported one, already with 2.081 it fails to build - current dmd version is 2.082 already.

```
vibe-d:stream 0.8.4: building configuration "library"...
../../../.dub/packages/vibe-d-0.8.4/vibe-d/stream/vibe/stream/wrapper.d(179,22): Error: no property connected for type const(InterfaceProxy!(ConnectionStream))
../../../.dub/packages/vibe-d-0.8.4/vibe-d/stream/vibe/stream/wrapper.d(187,19): Error: no property connected for type InterfaceProxy!(ConnectionStream)
../../../.dub/packages/vibe-d-0.8.4/vibe-d/stream/vibe/stream/wrapper.d(188,15): Error: no property close for type InterfaceProxy!(ConnectionStream)
../../../.dub/packages/vibe-d-0.8.4/vibe-d/stream/vibe/stream/wrapper.d(198,22): Error: no property waitForData for type InterfaceProxy!(ConnectionStream)
```

Extrawurst:
Just realised how long there was no activity on the project here. Is vibe.d discontinued? see also http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/thread/65001/

s-ludwig:
I'm currently preparing eventcore and vibe-core to be tested on the latest versions. But I was unter the impression that the project tester is still testing vibe.d to make sure that there are no DMD/Phobos/Druntime regressions. Anyway, I'll look into this afterwards.

But no, the development is not discontinued, but I'm currently not really able to contribute. This is also true for DUB and the other projects. Basically the only exceptions are occasional fixes in eventcore/vibe-core that are needed for our internal project and occasional PR reviews. This will change once the project is on the market.

There is also the ongoing HTTP/2 SAoC project, which happens in the separate vibe-http repository, so if everything goes right, there should finally be basic HTTP/2 functionality by the end of the year.

wilzbach:
This is a known regression which was the reason for the quick 2.081.1 follow-up.
See https://github.com/vibe-d/vibe.d/issues/2181

PR #2216, opened 2018-10-02
Update compiler support to 2.076.1-2.080.0.

s-ludwig:
See #2215


s-ludwig:
Does anyone know right off the bat what causes the Meson errors for DMD/libevent builds (e.g. https://travis-ci.org/vibe-d/vibe.d/jobs/436089694)? I don't really know where those command line arguments come from and with what they should be replaced. CC @ximion

PR #2219, opened 2018-10-07
Added minimum requirement section.

Beyarz:
I had to upgrade my server to at least 4096 MB in order to compile without any fatal errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,3 +246,7 @@ used in analogy to the above to build against the OpenSSL 0.9.8 branch.
 
 Finally, there is a "botan" configuration for using the D port of the Botan
 library.
+
+Minimum requirements
+--------------------
+RAM: 4096 MB


Geod24:
@Beyarz : That widely depends on the code you are compiling, and the compiler you are using.
It doesn't really makes sense for Vibe.d to have this information here, as some people will be able to work with less than a quarter of this.

However, if you are finding errors which made it hard to track down the source of the problem, I encourage you to report that to DMD's bugzilla.

Beyarz:
> @Beyarz : That widely depends on the code you are compiling, and the compiler you are using.
> It doesn't really makes sense for Vibe.d to have this information here, as some people will be able to work with less than a quarter of this.
> 
> However, if you are finding errors which made it hard to track down the source of the problem, I encourage you to report that to DMD's bugzilla.

Great point, didn't think of that!

PR #2223, opened 2018-10-20
fix #2222

### LINKED ISSUE #2222: Unable to serveRestJSClient with empty URL
If I want to use js client generation I **must** to specify url for requests. But if I want use my program in different machines I need to get ip at each machine at start of program. If I want use many internet interfaces with different ip it's makes problems. But js client what send requests to self server don't need full url (for example `http://192.168.99.13/api/v1/some`). It's can send requests to relative url (`/api/v1/some`).

deviator:


deviator:
ping @s-ludwig

deviator:
ping @wilzbach

inet/vibe/inet/url.d
```diff
@@ -293,10 +293,13 @@ struct URL {
...
-		dst.put(":");
-		if (isDoubleSlashSchema(schema))
-			dst.put("//");
+		if (schema.length)
+		{
+			dst.put(schema);
+			dst.put(":");
+			if (isDoubleSlashSchema(schema))
+				dst.put("//");
+		}
```
s-ludwig: Without the schema, the generated string is not a URL anymore. Of course it could be argued to extend this class into a full URI, but since this is just a side-effect, I'd like to keep that a separate question. Without having looked at the JS generation code too closely, to me it seems like this optimization should be handled there. If I'm not mistaken, this should also be made an opt-in/out setting, since the JS might be accessed from page hosted on a different server or sub domain.

wilzbach:
Travis failures are very likely related to https://github.com/vibe-d/eventcore/pull/86

s-ludwig:
@wilzbach, I only see `std.digest` related errors and the Meson build failing

s-ludwig:
Merged #2216, which fixes the Meson build error.

web/vibe/web/internal/rest/jsclient.d
```diff
@@ -100,12 +105,24 @@ class JSRestClientSettings
...
+				}
+			} else {
+				if (route.fullPathParts.length)
+				{
+					fmtParam(fout, route.fullPathParts[0]);
+					foreach (p; route.fullPathParts[1..$]) {
+						fout.put(" + ");
+						fmtParam(fout, p);
+					}
+				}
```
s-ludwig: Would be good to have an `assert(false, "Empty route function path");` here in the else case, so that no invalid JavaScript can be produced here. (the `else if` could also be merged to a single line without an extra block)

web/vibe/web/internal/rest/jsclient.d
```diff
@@ -100,12 +105,24 @@ class JSRestClientSettings
...
 			burl.pathString = "/";
-			fout.serializeToJson(burl.toString()[0 .. $-1]);
-			// and then assemble the full path piece-wise
-			foreach (p; route.fullPathParts) {
-				fout.put(" + ");
-				if (!p.isParameter) fout.serializeToJson(p.text);
-				else fout.formattedWrite("toRestString(%s)", p.text);
+
+			// if url not empty
+			if (intf.baseURL != intf.basePath) {
```
s-ludwig: I think using `intf.baseURL != URL.init` would make this a bit more obvious here.

s-ludwig:
Otherwise, apart from the failures looks good to merge now to me too. Would also be good to make a final squash/rebase to get rid of the two merge commits.

wilzbach:
> @wilzbach, I only see std.digest related errors and the Meson build failing

Look at the vibe-core failures:

```
+cd tests/mongodb
+dub --compiler=dmd --override-config=vibe-d:core/vibe-core --build-mode=separate
Performing "debug" build using dmd for x86_64.
diet-ng 1.5.0: target for configuration "library" is up to date.
mir-linux-kernel 1.0.1: target for configuration "library" is up to date.
taggedalgebraic 0.10.12: target for configuration "library" is up to date.
eventcore 0.8.37: target for configuration "epoll" is up to date.
stdx-allocator 2.77.4: target for configuration "library" is up to date.
vibe-core 1.4.3: target for configuration "epoll" is up to date.
vibe-d:utils 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:data 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:crypto 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:stream 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:textfilter 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:inet 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:tls 0.8.4+commit.23.gce16c4a: target for configuration "openssl" is up to date.
vibe-d:http 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
vibe-d:mongodb 0.8.4+commit.23.gce16c4a: target for configuration "library" is up to date.
tests ~master: building configuration "application"...
Linking...
To force a rebuild of up-to-date targets, run again with --force.
Running ./tests 
Program exited with code -11
```

Similar errors for vibe.d with the vibe-core subconfiguration happen on the Project Tester since yesterday, which is why I think it's due to vibe-d/eventcore#86

web/vibe/web/internal/rest/jsclient.d
deviator: `intf.baseURL` is a `string`, not `URL`. `URL.init` is empty, `baseURL == basePath` can be if `URL.init` is passed to `RestInterface.ctor` [here](https://github.com/vibe-d/vibe.d/blob/master/web/vibe/web/internal/rest/common.d#L41)
I don't understand why `RestInterface` use `strings` for `baseURL` and `basePath` instead `URL` for both. Is it principled? Why `URL` can't be initialize with empty string? Why `URL.init` prints not empty string? Maybe it can be changed? I think it can simplify code.

s-ludwig:
The remaining failures were just due to stdx-allocator still failing for the Meson build (I guess it just requires a new version tag).

PR #2232, opened 2018-11-13
[WIP] HTTP middleware

boolangery:
Middlewares are ran before each request handler of a rest or web controller.

The goal is to allow the user to define custom middlewares like IP banning for example.

### Minimal example to show the new feature:
```d
final class IpBanMiddleware : IMiddleware
{
	bool run(HTTPServerRequest req, HTTPServerResponse res) @safe
	{
		if (req.clientAddress.toAddressString() == "127.0.0.1")
			throw new HTTPStatusException(403);

		return true;
	}
}

interface IAPI
{
	string getSomeInfo();
}

class API : WebController, IAPI
{
	this()
	{
		registerMiddleware(new IpBanMiddleware());
	}
	
	string getSomeInfo()
	{
		return "Some Info!";
	}
}

shared static this()
{
	auto router = new URLRouter;
	router.registerRestInterface(new API());
	listenHTTP("127.0.0.1:8080", router);
}
```

With this system, `vibe.web.auth` could be refactored as a middleware.

Feedback appreciated :)

### DIFF
--- a/tests/rest-web-middlewares/dub.json
+++ b/tests/rest-web-middlewares/dub.json
@@ -1,0 +1,7 @@
+﻿{
+	"name": "tests",
+	"dependencies": {
+		"vibe-d": {"version": "~master", "path": "../../"}
+	},
+	"versions": ["VibeDefaultMain"]
+}
--- a/tests/rest-web-middlewares/source/app.d
+++ b/tests/rest-web-middlewares/source/app.d
@@ -1,0 +1,160 @@
+/* This example module consists from several small example REST interfaces.
+ * Features are not grouped by topic but by common their are needed. Each example
+ * introduces few new more advanced features. Sometimes registration code in module constructor
+ * is also important, it is then mentioned in example comment explicitly.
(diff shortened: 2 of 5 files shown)


gedaiu:
I think that you could just add your run method like this:

```   
        auto router = new URLRouter;
        auto middleware = new IpBanMiddleware();

        router.any("", &middleware.run);
	router.registerRestInterface(new API());
	listenHTTP("127.0.0.1:8080", router);
```

why you would need a middleware?

WebFreak001:
I agree with gedaiu, router already implements middleware which is run in order as registered into the router, which covers all use-cases of a middleware.

Also about vibe.web.auth: it is only done on web interfaces anyway, where I don't see how it would be helpful as middleware / explicitly registering it in the router when you already have it as attributes, which you can also mass apply using `@noAuth { ... }`

What use-case would the `router.any("*", func)` method not cover for you?

Otherwise can this be closed? (has been WIP for 1.5 years)

PR #2238, opened 2018-12-06
Fix segmentation fault on SSL_read due to null SSLState

GallaFrancesco:
While testing the HTTP/2 module there were some cases in which reading from a `OpenSSLStream` failed with a segmentation fault. This is due to the fact that the underlying `SSLState` gets freed when receiving a `close_notify` alert from TLS.

The `SIGSEV` was caused by the usage of a `null` `m_tls` state which would be passed to `SSL_read` and `SSL_peek`, resulting in a memory corruption. An easy fix for this is to add checks for `null` in both `OpenSSLStream.read()` and `OpenSSLStream.leastSize()` and since both methods are supposed to return an integral, return `0` instead.

### DIFF
--- a/tls/vibe/stream/openssl.d
+++ b/tls/vibe/stream/openssl.d
@@ -47,7 +47,7 @@ version(VibeForceALPN) enum alpn_forced = true;
 else enum alpn_forced = false;
 enum haveALPN = OPENSSL_VERSION_NUMBER >= 0x10200000 || alpn_forced;
 
-// openssl 1.1.0 hack: provides a 1.0.x API in terms of the 1.1.x API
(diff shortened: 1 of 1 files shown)


tls/vibe/stream/openssl.d
```diff
@@ -336,6 +338,7 @@ final class OpenSSLStream : TLSStream {
 	size_t read(scope ubyte[] dst, IOMode mode)
 	{
 		size_t nbytes = 0;
+		if(m_tls == null) return 0;
```
s-ludwig: This one should throw an exception instead (e.g. `new Exception("Reading from closed TLS stream")`).

PR #2247, opened 2019-01-02
Add handling of arrays in form/query data like 'arr[]'

### LINKED ISSUE #1984: Cannot access form "arrays" in web interface
Are web interface array parameters as in http://forum.rejectedsoftware.com/groups/rejectedsoftware.vibed/post/20928 still intended to work?
```
import vibe.web.web;

class MyWebService {
    // available as POST /form
    void postForm(string[] tags)
    {
        render!("some_template.dt", tags);
    }
}
```

I can't get this to work with the web interface; reading the data from the server request works though.

index.dt:
```
doctype html
html
    head
        title Testing
    body
        form(method="POST", action="form")
            input(name="data", type="text")
            input(name="data", type="text")
            input(name="data", type="text")
            button(type="submit") Submit
```
form.dt:
```
doctype html
html
    head
        title Data
    body
        p Data: #{data}
```
app.d:
```
import vibe.vibe;

void main()
{
    auto r = new URLRouter;
    r.registerWebInterface(new MyWebService);

    settings.port = 8080;
    listenHTTP(settings, r);

    logInfo("Please open http://127.0.0.1:8080/ in your browser.");
    runApplication();
}

class MyWebService {
    void index() {
        render!("index.dt");
    }
    void postForm(string[] data) {
        render!("form.dt", data);
    }
}
```
This outputs: "Data: []" for any input.
I've tried 0.8.1 on Windows and Linux, and 0.7.32 on Linux, with the same behavior.

schveiguy:
Fixes #1984 

A few notes about this:

1. I didn't see any unittests to modify, all the unittests in here just check for compilation issues.
2. This *should* work with arrays nested in structs, but not arrays of structs (unless the struct is completely serialized in one go). This is intentional, as the appending depends completely on serializing the whole thing at once.
3. DictionaryList requires a `@safe` delegate for getAll, but there's a lot of unsafe things within the `webConvTo` function, so I couldn't do it. TBH, I'm not sure its really the right approach to use the `=void`, and use that very unsafe copy mechanism, but I'm assuming there's a good reason.

Please double check the test I wrote for when things can't be serialized. I think I got it right, but I didn't test.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -1012,7 +1012,6 @@ final class HTTPServerRequest : HTTPRequest {
 
 		private void parseFormAndFiles() @safe {
 			_form = FormFields.init;
-			assert(!!bodyReader);
 			parseFormData(_form, _files, headers.get("Content-Type", ""), bodyReader, MaxHTTPHeaderLineLength);
 		}
 
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -691,7 +691,7 @@ private enum MAX_ARR_INDEX = 0xffff;
 // handle the actual data inside the parameter
 private ParamResult processFormParam(T)(scope string data, string fieldname, ref T dst, ref ParamError err)
 {
-	static if(is(T == bool))
+	static if (is(T == bool))
 	{
 		// getting here means the item is present, set to true.
 		dst = true;
@@ -699,7 +699,7 @@ private ParamResult processFormParam(T)(scope string data, string fieldname, ref
 	}
 	else
 	{
-		if (!(*pv).webConvTo(dst, err)) {
+		if (!data.webConvTo(dst, err)) {
 			err.field = fieldname;
 			return ParamResult.error;
 		}
@@ -724,16 +724,16 @@ package ParamResult readFormParamRec(T)(scope HTTPServerRequest req, ref T dst, 
 					!is(typeof(T.fromStringValidate(string.init, null))) &&
 					!is(typeof(T.fromISOExtString(string.init))));
 
-		static assert(!is(EL == bool),
-			"Boolean arrays are not allowed, because their length cannot " ~
-			"be uniquely determined. Use a static array instead.");
-		// array to check for duplicates
-		static if(isStaticArray!T)
+		static if (isStaticArray!T)
 		{
 			bool[T.length] seen;
 		}
 		else
 		{
+			static assert(!is(EL == bool),
+			  "Boolean arrays are not allowed, because their length cannot " ~
+			  "be uniquely determined. Use a static array instead.");
+			// array to check for duplicates
 			dst = T.init;
 			bool[] seen;
 		}
@@ -742,38 +742,38 @@ package ParamResult readFormParamRec(T)(scope HTTPServerRequest req, ref T dst, 
 		const minLength = fieldname.length + style == NestedNameStyle.d ? 2 : 1;
 		const indexTrailer = style == NestedNameStyle.d ? "]" : "";
 
-		void processItems(DL)(DL dlist)
+		ParamResult processItems(DL)(DL dlist)
 		{
-			foreach(k, v; dlist.byKeyValue)
+			foreach (k, v; dlist.byKeyValue)
 			{
-				if(k.length < minLength)
+				if (k.length < minLength)
 					// sanity check to prevent out of bounds
 					continue;
-				if(k.startsWith(fieldname) && k[fieldname.length] == indexSep)
+				if (k.startsWith(fieldname) && k[fieldname.length] == indexSep)
 				{
 					// found a potential match
 					string key = k[fieldname.length + 1 .. $];
 					size_t idx;
-					if(key == indexTrailer)
+					if (key == indexTrailer)
 					{
-						static if(isSimpleElement)
+						static if (isSimpleElement)
 						{
 							// this is a non-indexed array item. Find an empty slot, or expand the array
-							idx = seen.countUntil(false);
-							static if(isStaticArray!T)
+							import std.algorithm : countUntil;
+							idx = seen[].countUntil(false);
+							static if (isStaticArray!T)
 							{
-								if(idx == size_t.max)
+								if (idx == size_t.max)
 								{
 									// ignore extras, and we know there are no more matches to come.
 									break;
 								}
 							}
-							else if(idx == size_t.max)
+							else if (idx == size_t.max)
 							{
+								// append to the end.
 								idx = dst.length;
-								dst.length = idx + 1;
 							}
-							seen[idx] = true;
 						}
 						else
 						{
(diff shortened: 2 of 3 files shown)


schveiguy:
BTW, bikeshedding of `dAppend` attribute welcome, I don't like it either.

schveiguy:
The travis build failure seems unrelated, like a network failure.

schveiguy:
I'm going to start depending on this in my local project since I need it. Please let me know if there are any objections or changes that need to be made here.

schveiguy:
ping @s-ludwig

s-ludwig:
Just an idea that I haven't fully thought through, but I actually wanted to implement this kind of array specification a while ago. My plan was to just detect the style used and keep supporting both ways under the same `NestedNameStyle.d` name. Do you see any issues with this? I would say that since we are just talking about the receiving end that it should be fine.

schveiguy:
So there are decisions to be made with the API the way it currently works and adding in arbitrary appending to the array.

1. The usual mechanism used to find the elements is first to build the variable name (e.g. `arr[0]`) and then search the query parameters for that variable name. This doesn't lend itself well for blended names (i.e. having both `arr[0]` and `arr[]` in the query).
2. The dictionary list gives no way to get multiple identically named items out of it without using the delegate mechanism, or without allocating a separate throw-away array, so you have to make a decision early on whether you want to search for the non-indexed items (you can't just say "give me the 1st instance of this name").
3. Should the underscore mechanism also receive treatment this way? You could just include the name with no `_0` after it, or maybe just with a trailing underscore, and it could be the same thing. Seems like it would be inconsistent not to do this.

What I could do is change this PR to first search for the array using `arr[]`, and if I find that, use the delegate mechanism, and ignore any indexed ones. Another possibility is to search for both, and if I find both, throw an error. Though it may be useful to have blended items (i.e. you have `arr[0]`, `arr[1]` specified anywhere in the query, but then the others are all `arr[]` and are just tacked onto the end). In that case, I can search for indexed items first, and then search for non-indexed ones. There may be a weird inconsistency if you had for instance `arr[0]`, `arr[]`, `arr[]`, `arr[2]`. In that case (as it is today), the `arr[2]` would be ignored, only the first 3 would get in, and the array would actually have a 3rd element, but it wouldn't match the `arr[2]` parameter.

All of this also falls apart if you have an array of structs, as you can't use the non-indexed mechanism to specify an array of structs, I just want to make sure you agree with this.

I'm willing to do all of this work, just let me know what you think is the best way, and will get merged :)

Note, a way to get a range of elements that have the same name from `DictionaryList` would be super-useful, and would avoid the burden of requiring the calling code to be `@safe`. I'm willing to do that as well if you will accept it. There could also be a way to get the nth indexed item of a given name out of the list. So for instance `get("arr[]", 0)` could do that (though this is lesser performing since every lookup is O(n)).

Any other ideas, I'm game. I really need this functionality.

schveiguy:
Actually, looking at the DictionaryList, I can simply do byKeyValue and filter on the key name. So a range possibility already exists. I can switch to using that instead of the delegate.

s-ludwig:
I'd generally say that the mixed case is somthing that is rather "nice to have", but not at all required (ignoring one of the two, or erroring out would be okay if there are gains in terms of performance or implementation complexity). Supporting the append array style for the underscore convention would also be "nice", but mainly if the changed implementation makes that trivial.

Speaking of the implementation, to me it looks like the whole thing should really be turned upside down - iterating over the whole array using `byKeyValue` once and inserting into `dst` on-the-fly. There'd just have to be an additional flag per field of some sort to error out on missing fields. But I wouldn't mind keeping the current implementation and just changing the array part either.

Anyway, following that `byKeyValue` route, the most obvious implementation for handling the mixed case appears to be to let `arr[]` append and `arr[i]=value` would mean `if (i >= arr.length) arr.length = i + 1; arr[i] = value;`. The semantics that would result out if this seem relatively intuitive to me.

schveiguy:
> Speaking of the implementation, to me it looks like the whole thing should really be turned upside down - iterating over the whole array using `byKeyValue` once and inserting into `dst` on-the-fly. There'd just have to be an additional flag per field of some sort to error out on missing fields. But I wouldn't mind keeping the current implementation and just changing the array part either.

Yeah, I think I will try that approach and see how it goes. It's bound to perform MUCH better, no more linear searches for each item, and I can probably avoid building strings to search with as well.

schveiguy:
ping @s-ludwig I have updated the implementation based on the discussion. This was a REALLY hard thing to get right. Please don't merge yet, I haven't tested it. I have to think about how to unittest this. I just wanted your review.

I didn't update the struct-handling branch, it still searches the same way it did before for members.

How it works now:

`arr[]` will now append to the array in question if all slots have been filled. Otherwise, it fills a slot that hasn't been specified. So for instance: `arr[0]=hi&arr[2]=there&arr[]=vibe` turns into `["hi", "vibe", "there"]`

Unspecified slots are left as default values. So `arr[0]=hi&arr[2]=there` will result in `["hi", null, "there"]`.

I also specified a max array length for indexes as 2^16-1. This is to prevent an attack of something like `arr[40000000000]=crash`.

If an array element is specified twice, the second instance is ignored to preserve existing behavior.

If an array element beyond a static array length is specified, it's ignored to preserve existing behavior.

schveiguy:
Ah, one thing I just realized I forgot -- the static array branch should return error if not all elements are filled in...

But that's a minor change, will work on it maybe tomorrow.

schveiguy:
Will work on the compilation failures...

s-ludwig:
I would probably have handled the mixed cases just the way that yields the shortest code, but if mixing the two styles is considered an actual use case, the new semantics are probably more useful, so I think that looks fine, too.

Do you think that you'll be able to tackle fixing the compile error (+ add some tests) in the next few days? I'm planning to tag a new release and it would of course be nice to get this in (since the next stable one is probably a bit further off).

schveiguy:
I will work on this and try to get it done in the next couple days. Sorry, I've been busy with other work.

> I would probably have handled the mixed cases just the way that yields the shortest code, but if mixing the two styles is considered an actual use case, the new semantics are probably more useful

I can't tell what style is going to be used from the enum (I'm assuming you mean both append and indexed style), so I have to look for both. Plus you had said that it would be desirable to have sparse arrays, so I implemented that too.

I suppose the easiest thing to do would be to just try the `arr[]` name first, and if that works, use the range, otherwise use the original code. What I like about this implementation though, is we aren't building tons of intermediate throwaway strings.

Note, the part I wrestled with the most is that when you have a simple type (one that is parsed from a single string), then you don't really want to recurse and search the array again. So I thought it would be good to split out the simple parsing and the nested parsing. But it was not as straightforward as I thought it would be.

Anyway, will work on this soon.

schveiguy:
ping @s-ludwig I've fixed all the compilation issues (wow, it was bad. I should have tried testing before). Will monitor the CI to make sure everything works.

Also added unit tests.

http/vibe/http/server.d
```diff
@@ -1012,7 +1012,6 @@ final class HTTPServerRequest : HTTPRequest {
 
 		private void parseFormAndFiles() @safe {
 			_form = FormFields.init;
-			assert(!!bodyReader);
```
schveiguy: This is moved to `parseFormData`, because if the content type is not a form, `bodyReader` isn't used.

Makes unit testing much easier, since I don't have to invent a null input stream (which is suspiciously missing).

schveiguy:
The travis failure seems environment related. I don't understand the codecov report, I definitely am unit testing the code.

web/vibe/web/common.d
```diff
@@ -685,33 +685,180 @@ package enum ParamResult {
...
+			return ParamResult.error;
+		if(processItems(req.query) == ParamResult.error)
+			return ParamResult.error;
+
+		// make sure all static array items have been seen
+		static if(isStaticArray!T)
+		{
+			import std.algorithm : canFind;
+			if(seen[].canFind(false))
+				return ParamResult.error;
```
schveiguy: Looking at this I realize this is a different behavior. The previous behavior asserted on this, instead of returning an error. Which means non-release code would throw an Error, whereas release code would let it slide. Not sure of the thoughts on this.

s-ludwig: Do you mean line 712 in the original code? That assertion should always be true, because `readFormParamRect` is called with `required==true` and thus should never return `ParamResult.skipped`. In that case, the new behavior would be consistent.

s-ludwig:
I restarted the failing travis job, looks like it should pass then. I think the CodeCov results are looking strange because the branch needs a rebase. It seems to contain all changes on master since (presumably) January.

web/vibe/web/common.d
schveiguy: Yes, you are right! But here is something I forgot. I didn't fill in the error struct as before. I can fix, but I assume this means CI will restart (and it's all green right now).

Maybe better on a quick fixup?

web/vibe/web/common.d
```diff
@@ -753,6 +900,49 @@ package ParamResult readFormParamRec(T)(scope HTTPServerRequest req, ref T dst,
...
+	int[][] arr2;
+	result = req.readFormParamRec(arr2, "arr", false, NestedNameStyle.d, err);
+	assert(result == ParamResult.ok);
+	import std.stdio;
+	assert(arr2 == [[2,4],[3],[1]]);
+
+	int[][3] staticarr2;
+	result = req.readFormParamRec(staticarr2, "arr", false, NestedNameStyle.d, err);
+	assert(result == ParamResult.ok);
+	import std.stdio;
```
schveiguy: ugh, forgot to remove these too.

web/vibe/web/common.d
```diff
@@ -802,6 +992,7 @@ package void setVoid(T, U)(ref T dst, U value)
 unittest {
 	static assert(!__traits(compiles, { bool[] barr; ParamError err;readFormParamRec(null, barr, "f", true, NestedNameStyle.d, err); }));
 	static assert(__traits(compiles, { bool[2] barr; ParamError err;readFormParamRec(null, barr, "f", true, NestedNameStyle.d, err); }));
+	//{ bool[2] barr; ParamError err;readFormParamRec(null, barr, "f", true, NestedNameStyle.d, err); }
```
schveiguy: And this... grrr....

schveiguy:
Also TODO: need to update docs.

schveiguy:
OK, I fixed the issues I had identified. Should be good to go.

web/vibe/web/common.d
schveiguy: I fixed it, good to merge. May want to wait for the CI to finish.

web/vibe/web/common.d
```diff
@@ -753,6 +905,47 @@ package ParamResult readFormParamRec(T)(scope HTTPServerRequest req, ref T dst,
...
+	assert(staticarr == [2,3,1]);
+
+	// try nested arrays
+	req = createTestHTTPServerRequest(URL("http://localhost/route?arr[2][]=1&arr[0][]=2&arr[1][]=3&arr[0][]=4"));
+	int[][] arr2;
+	result = req.readFormParamRec(arr2, "arr", false, NestedNameStyle.d, err);
+	assert(result == ParamResult.ok);
+	assert(arr2 == [[2,4],[3],[1]]);
+
+	int[][2] staticarr2;
```
schveiguy: I modified this to a 2-element array to test that the `arr[2][]` field is ignored (found a bug in the process as well).

schveiguy:
ping @s-ludwig this should be all set.

s-ludwig:
Okay, looks good AFAICS. I'll merge and tag a beta release.

schveiguy:
> TODO: need to update docs.

See #2266

schveiguy:
@s-ludwig I found a bug in this code (stupid operator precedence bug). Will submit a patch soon, don't release yet.

schveiguy:
Followup PR: #2270

PR #2249, opened 2019-01-10
Fix Json array clone

### LINKED ISSUE #2246: Json clone concatenates arrays
Having this json:

```
`{"a": [[[1,2] [3,4]]] }`.parseJsonString.clone();
```

after cloning will be 
```
{"a": [1,2,3,4]}
```

gedaiu:
fix  #2246


gedaiu:
I guess the `~=` operator is confusing for arrays. I often do the same mistake. Do you think that a method like `addObject` would help?

PR #2250, opened 2019-01-10
Made WebParamAttribute public to allow rest introspection

burner:
I'm currently building a angular service generator.
This is done by introspecting rest interfaces.
To get the parameter names right, it needs to look at the UDAs, and more
specifically at WebParamAttribute.

### DIFF
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -482,14 +482,18 @@ package struct PathAttribute
 /// private
 package struct NoRouteAttribute {}
 
-/// Private struct describing the origin of a parameter (Query, Header, Body).
-package struct WebParamAttribute {
+/**
+ * This struct contains a mapping between the name used by HTTP (field)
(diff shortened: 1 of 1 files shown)


web/vibe/web/common.d
```diff
@@ -482,14 +482,18 @@ package struct PathAttribute
...
+ * This struct contains a mapping between the name used by HTTP (field)
+ * and the parameter (identifier) name of the function.
+ */
+public struct WebParamAttribute {
 	import vibe.web.internal.rest.common : ParameterKind;
 
+	/// The type of the WebParamAttribute
 	ParameterKind origin;
-	/// Parameter name
+	/// Parameter name (Server side)
```
s-ludwig: What about "function parameter name" to make this even more clear?

burner: will do

burner:
@s-ludwig thank you

PR #2256, opened 2019-01-22
Fix #2241

### LINKED ISSUE #2241: auth_digest crashes : out of memory
the latest vibe.d auth_digest example crashes due to out of memory.

just compile and run, and after input username and passwd, the example executable crashed.

`core.exception.OutOfMemoryError@core/exception.d(700): Memory allocation failed`
Program exited with code -11

kubo39:
This fixes segmentation fault in auth_digest

### DIFF
--- a/http/vibe/http/auth/digest_auth.d
+++ b/http/vibe/http/auth/digest_auth.d
@@ -132,7 +132,7 @@ HTTPServerRequestDelegate performDigestAuth(DigestAuthInfo info, scope DigestHas
 /// Scheduled for deprecation - use a `@safe` callback instead.
 HTTPServerRequestDelegate performDigestAuth(DigestAuthInfo info, scope string delegate(string, string) @system pwhash)
 @system {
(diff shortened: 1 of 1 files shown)


s-ludwig:
Do you have an idea what causes the crash?  From a coarse look the existing code looks fine, in which case I'd try to get a reduced test case for a DMD bug report.

kubo39:
I'm sorry, you're right. First I thought this is related to https://issues.dlang.org/show_bug.cgi?id=16037, but I investigated and the reason why it crashes is that DMD decrements `tookAddressOf` parameter when scope function was given.
I'm not sure this is DMD's bug and still investigate how it should be, anyway, close this. :bowing_man:

kubo39:
The [function's parameter storage classes definition](https://dlang.org/spec/function.html#param-storage) says that scope parameter storage class: references in the parameter cannot be escaped.
So, at least, we should not expect to generate closure on the heap, IMO.

PR #2257, opened 2019-01-31
Remove DNS cache

marcioapm:
Caching at this layer is not going to work, because we don't know the TTL.
Because of this cache, you might be in a situation where you try to connect to a host that is no longer there, or worse, you connect to a host that is not what you expected.


s-ludwig:
Sounds okay, though. This task should be done by an external caching DNS proxy or to some wrapper around `resolveHost` instead.

PR #2263, opened 2019-02-14
Make PathAttribute public for introspection

burner:
To build a function that can call a rest interfaces and provide a JWT,
I needed to introspect on the PathAttribute.


burner:
thank you

PR #2274, opened 2019-03-02
Fix #1941 - real fix for the workaround

### LINKED ISSUE #1941: vibe-d-0.8.2 tree structure serialization still a problem
Partialy solved in #1855

But if we change the unittest from system to safe:
```D
@safe unittest {
	import vibe.data.json;
	static struct Bar { Bar[] foos; int i; }
	auto b = deserializeJson!Bar(`{"i":1,"foos":[{"foos":[],"i":2}]}`);
	assert(b.i == 1);
	assert(b.foos.length == 1);
	assert(b.foos[0].i == 2);
}
```

We get:
```
data/vibe/data/serialization.d(799,30): Error: @safe function vibe.data.serialization.__unittestL796_160' cannot call @system function 'vibe.data.json.deserializeJson!(Bar, string).deserializeJson'
```

When we also use `@optional` on the array field we get:
```
data/vibe/data/serialization.d(752,30): Error: function vibe.data.json.JsonStringSerializer!(string, false).JsonStringSerializer.readDictionary!(Traits!(Bar, DefaultPolicy, optional)).readDictionary (scope void delegate(string) @safe entry_callback) is not callable using argument types (void) data/vibe/data/serialization.d(567,108): Error: template instance vibe.data.serialization.deserializeValueImpl!(JsonStringSerializer!(string, false), DefaultPolicy).deserializeValueDeduced!(Bar, optional) error instantiating
data/vibe/data/serialization.d(664,32):        instantiated from here: deserializeValue!(Bar, optional)
data/vibe/data/serialization.d(567,108):        instantiated from here: deserializeValueDeduced!(Bar[], optional)
data/vibe/data/serialization.d(767,48):        instantiated from here: deserializeValue!(Bar[], optional)
data/vibe/data/serialization.d(752,31):        ... (4 instantiations, -v to show) ...
data/vibe/data/json.d(1436,48):        instantiated from here: deserialize!(JsonStringSerializer!(string, false), Bar, string)
data/vibe/data/serialization.d(799,30):        instantiated from here: deserializeJson!(Bar, string)
data/vibe/data/serialization.d(752,30): Error: function vibe.data.json.JsonStringSerializer!(string, false).JsonStringSerializer.readDictionary!(Traits!(Bar, DefaultPolicy)).readDictionary (scope void delegate(string) @safe entry_callback) is not callable using argument types (void)
```

tchaloupka:
This bothered me so long that I've finally looked into it and found the issue. :)

Problem was that test for `isSafeSerializer` is using `writeValue` and `readValue` for checking and to pass correctly it requires both to compile.
In unittests, `TestSerializer` is used and works both as a serializer and a deserializer, so the test passes ok.
But that is not the case for ie `JsonStringSerializer` which has only one of those based on the provided range type.

### DIFF
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -1199,11 +1199,15 @@ private template isBuiltinTuple(T, string member)
 }
 
 // heuristically determines @safe'ty of the serializer by testing readValue and writeValue for type int
-private enum isSafeSerializer(S) = __traits(compiles, (S s) @safe {
+private template isSafeSerializer(S)
+{
 	alias T = Traits!(int, DefaultPolicy);
-	s.writeValue!T(42);
-	s.readValue!(T, int)();
-});
+	static if (__traits(hasMember, S, "writeValue"))
+		enum isSafeSerializer = __traits(compiles, (S s) @safe { s.writeValue!T(42); });
+	else static if (__traits(hasMember, S, "readValue"))
+		enum isSafeSerializer = __traits(compiles, (S s) @safe { s.readValue!(T, int)(); });
+	else static assert(0, "Serializer without writeValue or readValue is invalid");
+}
 
 private template hasAttribute(T, alias decl) { enum hasAttribute = findFirstUDA!(T, decl).found; }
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Ah, good catch! I'm wondering whether the two cases should be ANDed together to not miss an unsafe deserializer part when a safe serializer part is detected. Or maybe the best approach would be to just split this up into a `isSafeSerializer` and `isSafeDeserializer`?

tchaloupka:
Right, splitting it up is probably the best. I've made the change.

PR #2275, opened 2019-03-02
change data serialization methods to `auto ref in`

tchaloupka:
Also allow data to have disabled copy constructor.

### DIFF
--- a/data/vibe/data/bson.d
+++ b/data/vibe/data/bson.d
@@ -1780,4 +1780,4 @@ pure @safe {
 }
 
 /// private
-package template isBsonSerializable(T) { enum isBsonSerializable = is(typeof(Unqual!T.init.toBson()) == Bson) && is(typeof(Unqual!T.fromBson(Bson())) == Unqual!T); }
+package template isBsonSerializable(T) { enum isBsonSerializable = is(typeof(T.init.toBson()) : Bson) && is(typeof(T.fromBson(Bson())) : T); }
--- a/data/vibe/data/bson.d
+++ b/data/vibe/data/bson.d
@@ -1165,7 +1165,7 @@ struct BsonRegex {
 
 	See_Also: `deserializeBson`
 */
-Bson serializeToBson(T)(T value, ubyte[] buffer = null)
+Bson serializeToBson(T)(auto ref in T value, ubyte[] buffer = null)
 {
 	return serialize!BsonSerializer(value, buffer);
 }
@@ -1466,10 +1466,9 @@ struct BsonSerializer {
 	void beginWriteArrayEntry(Traits)(size_t idx) { m_entryIndex = idx; }
 	void endWriteArrayEntry(Traits)(size_t idx) {}
 
-	// auto ref does't work for DMD 2.064
-	void writeValue(Traits, T)(/*auto ref const*/ in T value) { writeValueH!(T, true)(value); }
+	void writeValue(Traits, T)(auto ref in T value) { writeValueH!(T, true)(value); }
 
-	private void writeValueH(T, bool write_header)(/*auto ref const*/ in T value)
+	private void writeValueH(T, bool write_header)(auto ref in T value)
 	{
 		static if (write_header) writeCompositeEntryHeader(getBsonTypeID(value));
 
@@ -1604,7 +1603,7 @@ struct BsonSerializer {
 		return false;
 	}
 
-	private static Bson.Type getBsonTypeID(T, bool accept_ao = false)(/*auto ref const*/ in T value)
+	private static Bson.Type getBsonTypeID(T, bool accept_ao = false)(auto ref in T value)
 	@safe {
 		Bson.Type tp;
 		static if (is(T == Bson)) tp = value.type;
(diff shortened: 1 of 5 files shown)


s-ludwig:
Nice that this works now for all supported compilers, this should definitely work in some way. Unfortunately, requiring getters and `toXXX` to be `const`, this is a hard breaking change. Wouldn't just `auto ref` without the `in` work, too (letting `T` be inferred as `const`)?

data/vibe/data/json.d
```diff
@@ -1779,22 +1779,23 @@ struct JsonSerializer {
...
-		} else static if (isSomeString!T && !is(T == string)) {
+		} else static if (isSomeString!T && !is(UT == string)) {
 			writeValue!Traits(value.to!string);
 		} else m_current = Json(value);
 	}
 
-	void writeValue(Traits, T)(Json value) if (is(T == Json)) { m_current = value; }
-	void writeValue(Traits, T)(in Json value) if (is(T == Json)) { m_current = value.clone; }
+	void writeValue(Traits, T)(auto ref T value) if (is(Unqual!T == Json)) { m_current = value; }
+	//void writeValue(Traits, T)(auto ref in T value) if (is(Unqual!T == Json)) { m_current = value.clone; }
```
tchaloupka: I'm not sure what to do about this as all tests passes without it.

tchaloupka:
Agreed would definitely break something.
I've lessened it just to `auto ref` and moved type qualifier awareness to serializers to handle as required.
It seems ok, but I'm not so sure all the cases are handled properly as it's a bit magic sometimes :)

data/vibe/data/bson.d
```diff
@@ -1229,7 +1229,7 @@ unittest {
 @safe unittest {
 	static struct A { int value; static A fromJson(Json val) @safe { return A(val.get!int); } Json toJson() const @safe { return Json(value); } Bson toBson() { return Bson(); } }
 	static assert(!isStringSerializable!A && isJsonSerializable!A && !isBsonSerializable!A);
-	static assert(!isStringSerializable!(const(A)) && !isJsonSerializable!(const(A)) && !isBsonSerializable!(const(A)));
+	static assert(!isStringSerializable!(const(A)) && isJsonSerializable!(const(A)) && !isBsonSerializable!(const(A)));
```
tchaloupka: This test changed as it can be serialized to Json using const struct just fine.

utils/vibe/utils/dictionarylist.d
```diff
@@ -72,7 +72,7 @@ struct DictionaryList(VALUE, bool case_sensitive = true, size_t NUM_STATIC_FIELD
 		return ret;
 	}
 	/// ditto
-	FieldTuple[] toRepresentation() {
+	FieldTuple[] toRepresentation() const {
```
tchaloupka: Probably not needed after `auto ref in` -> `auto ref`

data/vibe/data/bson.d
s-ludwig: The thing here is that the `isXSerializable` traits work bi-directionally (i.e. both, serialization and deserialization must be possible) and . I'd have to think about possibilities how this change might or might not break existing code, but my initial guess would be that this should be fine in all typical cases.

I'd just apply the `Unqual` change to `isStringSerializable` and `isBsonSerializable` as well to keep the behavior consistent.

utils/vibe/utils/dictionarylist.d
s-ludwig: True, but I'd still leave it, since it can't hurt here. Only the added `const` in the unit tests are problematic.

data/vibe/data/json.d
s-ludwig: I think we need both, if serializing `const(Json)` is supposed to work. I'd do it with template constraints in this case instead of `Unqual`:
```
	void writeValue(Traits, T)(auto ref T value) if (is(T == Json)) { m_current = value; }
	void writeValue(Traits, T)(auto ref T value) if (!is(T == Json) && is(T : const(Json))) { m_current = value.clone; }
```
If you add this change, it would be good to also have a unit test that tests something like a `const(C)` with `class C { Json foo; }` to make sure the second case is actually hit.

tchaloupka: fixed

data/vibe/data/bson.d
tchaloupka: changed with some more unittest changes

utils/vibe/utils/dictionarylist.d
tchaloupka: ok, I've reverted added const in unittest I've overooked

s-ludwig:
Thanks, this looks good now. The only thing I'm still wondering is how the correct `isXSerializable` traits should look like. I'm thinking about something like this, because the trait should only yield true for `const(T)` when it actually has a `const` `toX` method:
```
enum isXSerializable(T) = is(typeof(T.init.toX()) : X) && is(T.fromX(X.init) : T);
```

tchaloupka:
I've made the change, hopefully in the expected way :) Tests passed ok for me.

PR #2276, opened 2019-03-05
Ddoc warning fix

denizzzka:
```
json.d(2149,6): Warning: Ddoc: parameter count mismatch, expected 3, got 2
```

How it passed CI?!

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -2143,6 +2143,8 @@ unittest
 	Params:
 		dst   = References the string output range to which the result is written.
 		json  = Specifies the JSON value that is to be stringified.
+		level = Specifies the base amount of indentation for the output. Indentation is always
+				done using tab characters.
 
 	See_Also: Json.toString, writePrettyJsonString
 */


wilzbach:
I don't think that the CI checks for this.

wilzbach:
> I don't think that the CI checks for this.

It would need to use `-w` for this.

denizzzka:
> It would need to use -w for this.

What you mean?

wilzbach:
> > It would need to use -w for this.
> 
> What you mean?

Otherwise the CI wouldn't error, but just print the warning.

denizzzka:
> Otherwise the CI wouldn't error, but just print the warning.

But why Ddoc is not checked by CI in this repo?

(Current version of dub ddoc build causes to check and build docs for ALL depend packages)

ZombineDev:
@wilzbach you can merge manually, as for some reason TravisCI hasn't reported it's status to GitHub.

wilzbach:
I can't. One needs to be admin for this.

PR #2289, opened 2019-03-28
consider 304 Not Modified as a success code

tchaloupka:
We use a proxy mechanism in our API and have log full of exceptions due to 304 Not Modified handled as error code.

Of 3xx codes this one should probably be handled as a success call?

### DIFF
--- a/http/vibe/http/status.d
+++ b/http/vibe/http/status.d
@@ -185,10 +185,10 @@ bool justifiesConnectionClose(int status)
 }
 
 /**
-	Determines if status code is generally successful (>= 200 && < 300)
+	Determines if status code is generally successful ((>= 200 && < 300) || 304)
 */
 bool isSuccessCode(HTTPStatus status)
 {
-	return status >= 200 && status < 300;
+	return (status >= 200 && status < 300) || status == 304;
 }
 


s-ludwig:
Not sure about this one. I'm a bit worried that including codes from the 3xx range would make it wrong to assume that the reponse body contains the expected resource contents. Maybe some kind of `enum HTTPStatusClass { informational, success, redirect, clientError, serverError }` would be better, which could then be combined in a `final switch` depending on the situation - or reverse it and add an `isErrorStatus`.

tchaloupka:
What I'm trying to workaround is error handling in REST client here: https://github.com/vibe-d/vibe.d/blob/master/web/vibe/web/rest.d#L1895
Which throws new exception that is than handled here: https://github.com/vibe-d/vibe.d/blob/master/http/vibe/http/client.d#L513

So logged as error (in debug log level) with complete backtrace.
Combined with this: https://issues.dlang.org/show_bug.cgi?id=19322
It leads to a lot of memory used by vibe-d process in debug builds.

So what I would like to achieve here is either change this behavior for this status code or be able to set the different behavior externally.

I get that altering the `isSuccessCode` is a bit hackish, but was the easiest change ;-)

So what would be the best approach for this?
* add special case handling in `HttpClient` catch handler to log backtrace only for some `HTTPStatusClass` (for `HTTPStatusException`)
* add some custom error handler to `HTTPClientSettings` and change the exception handler to use it instead if set
* ?

Thanks

WebFreak001:
Looking at other web frameworks like ASP.NET core, success status codes are only defined as 200-299.

I think adding "not modified" as success status code would cause unnecessary confusion and could introduce silent bugs. Status code 304 may only be sent if the request contains cache check headers like `If-None-Match` or `If-Modified-Since`.

Only then if the requester chooses to include these cache headers, 304 may be considered a valid success status code. In this case the requester should check for it explicitly instead of using isSuccessCode.

Geod24:
Closing as there is enough evidence to support we do not want to change 3xx to success codes

PR #2292, opened 2019-04-08
Fix HashMap when keys are objects that may override toHash and/or opEquals

n8sh:
Includes a unittest demonstrating code that fails without this patch.

### DIFF
--- a/utils/vibe/utils/hashmap.d
+++ b/utils/vibe/utils/hashmap.d
@@ -17,12 +17,16 @@ struct DefaultHashMapTraits(Key) {
 	enum clearValue = Key.init;
 	static bool equals(in Key a, in Key b)
 	{
-		static if (is(Key == class)) return a is b;
+		static if (__traits(isFinalClass, Key) && &Unqual!Key.init.opEquals is &Object.init.opEquals)
+			return a is b;
+		else static if (is(Key == class))
+			// BUG: improperly casting away const
+			return () @trusted { return a is b ? true : (a !is null && (cast(Object) a).opEquals(cast(Object) b)); }();
 		else return a == b;
 	}
 	static size_t hashOf(in ref Key k)
 	@safe {
-		static if (is(Key == class) && &Unqual!Key.init.toHash is &Object.init.toHash)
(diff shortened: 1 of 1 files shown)


n8sh:
The current test failure appears to be spurious so I'll just force-push to make the tests repeat until all pass.

wilzbach:
Just ignore the spurious test failures. No point in wasting time with fighting against them.

utils/vibe/utils/hashmap.d
```diff
@@ -17,12 +17,16 @@ struct DefaultHashMapTraits(Key) {
 	enum clearValue = Key.init;
 	static bool equals(in Key a, in Key b)
 	{
-		static if (is(Key == class)) return a is b;
+		static if (__traits(isFinalClass, Key) && &Unqual!Key.init.opEquals is &Object.init.opEquals)
+			return a is b;
+		else static if (is(Key == class))
+			// BUG: improperly casting away const
+			return () @trusted { return a is b ? true : (a !is null && (cast(Object) a).opEquals(cast(Object) b)); }();
```
wilzbach: Why not?

```suggestion
			return () @trusted { return a is b ? true :  a == b; }();
```

n8sh: I wanted to avoid it because it does a bunch of work solely to support improperly designed class hierarchies that can have asymmetric opEquals:

https://github.com/dlang/druntime/blob/60d992da13027948b590e7804238e60bcefafde2/src/object.d#L1095-L1115
```d
bool opEquals(Object lhs, Object rhs)
{
    // If aliased to the same object or both null => equal
    if (lhs is rhs) return true;


    // If either is null => non-equal
    if (lhs is null || rhs is null) return false;


    // If same exact type => one call to method opEquals
    if (typeid(lhs) is typeid(rhs) ||
        !__ctfe && typeid(lhs).opEquals(typeid(rhs)))
            /* CTFE doesn't like typeid much. 'is' works, but opEquals doesn't
            (issue 7147). But CTFE also guarantees that equal TypeInfos are
            always identical. So, no opEquals needed during CTFE. */
    {
        return lhs.opEquals(rhs);
    }


    // General case => symmetric calls to method opEquals
    return lhs.opEquals(rhs) && rhs.opEquals(lhs);
}
```

PR #2297, opened 2019-04-11
Add new cookie option: sameSiteStrict

wilzbach:
See also:
- https://www.owasp.org/index.php/SameSite
- https://medium.com/compass-security/samesite-cookie-attribute-33b3bfeaeb95

### DIFF
--- a/http/vibe/http/common.d
+++ b/http/vibe/http/common.d
@@ -710,12 +710,19 @@ final class Cookie {
 		long m_maxAge;
 		bool m_secure;
 		bool m_httpOnly;
+		SameSite m_sameSite;
 	}
 
 	enum Encoding {
 		url,
 		raw,
 		none = raw
+	}
+
+	enum SameSite {
+		default_,
+		lax,
+		strict,
 	}
 
 	/// Cookie payload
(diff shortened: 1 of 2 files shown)


wilzbach:
Tested this and it can be used with https://github.com/vibe-d/vibe.d/pull/2299

```d
settings.sessionOptions |= SessionOption.sameSiteStrict;
```

image

image

http/vibe/http/common.d
```diff
@@ -768,6 +769,12 @@ final class Cookie {
 	/// ditto
 	@property bool httpOnly() const { return m_httpOnly; }
 
+	/** Prevent cross-request forgery.
```
s-ludwig: typo: "cross-site request forgery"

http/vibe/http/common.d
```diff
@@ -768,6 +769,12 @@ final class Cookie {
...
+	*/
+	@property void sameSiteStrict(bool value) { m_sameSiteStrict = value; }
```
s-ludwig: Since this is the lowest API level, it would be nice to add a `Cookie.SameSite` enum with `default`, `lax` and `strict`, and a `@property void sameSite(SameSite)`.

http/vibe/http/server.d
```diff
@@ -837,7 +837,17 @@ enum SessionOption {
...
+
+	/**
+    Prevents the browser from sending this cookie along with cross-site requests.
+
+    By default, the protection is `lax`. This flag allows to set it to `strict`.
+    The strict value will prevent the cookie from being sent by the browser
+    to the target site in all cross-site browsing context,
+    even when following a regular link.
+	*/
+	sameSiteStrict = 1<<3,
```
s-ludwig: I'd even argue to make this the default and provide an optional flag to switch back to the default behavior. Should of course be noted prominently in the change log, but it would be bad to leave the safe-by-default track here.

wilzbach:
> Since this is the lowest API level, it would be nice to add a Cookie.SameSite enum with default, lax and strict, and a @property void sameSite(SameSite).

Done.

> I'd even argue to make this the default and provide an optional flag to switch back to the default behavior. 

Done :+1:

PR #2298, opened 2019-04-11
Allow to configure options of SessionVar

wilzbach:
Currently, it's not possible to configure the session initialization when `SessionVar` is used.


wilzbach:
Closing in favor of https://github.com/vibe-d/vibe.d/pull/2299 as the other PR is more generic and this should really be done in HTTPServerSettings.

PR #2303, opened 2019-04-16
Add way to disable port and address reuse

rracariu:
This integrates with recent event core `reuseAddress` flag.
It allows one to selectively disable those flags.
See https://github.com/vibe-d/vibe-core/pull/150

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -531,7 +531,9 @@ private enum HTTPServerOptionImpl {
 	none                      = 0,
 	errorStackTraces          = 1<<7,
 	reusePort                 = 1<<8,
-	distribute                = 1<<9 // deprecated
+	distribute                = 1<<9, // deprecated
+	reuseAddress              = 1<<10,
+	defaults                  = reuseAddress
 }
 
 // TODO: Should be turned back into an enum once the deprecated symbols can be removed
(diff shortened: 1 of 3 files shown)


wilzbach:
If it's the default, why not rename the options accordingly then: `doNotReuseAddress` (or sth. like that)

http/vibe/http/server.d
```diff
@@ -585,13 +586,15 @@ struct HTTPServerOption {
...
+	/// Enable address reuse in `listenTCP()`
+	static enum reuseAddress              = HTTPServerOptionImpl.reuseAddress;
 
 	/** The default set of options.
 
 		Includes all parsing options, as well as the `errorStackTraces`
 		option if the code is compiled in debug mode.
 	*/
-	static enum defaults = () { debug return HTTPServerOptionImpl.errorStackTraces; else return HTTPServerOptionImpl.none; } ().HTTPServerOption;
+	static enum defaults = () { debug return HTTPServerOptionImpl.errorStackTraces; else return HTTPServerOptionImpl.reuseAddress; } ().HTTPServerOption;
```
wilzbach: Uhm, now Vibe.d wouldn't set `reuseAddress` for debug builds ...

rracariu: Oops, fixed.

rracariu:
> If it's the default, why not rename the options accordingly then: `doNotReuseAddress` (or sth. like that)

Was looking to be consistent with `reusePort` and the `TCP` and `Stream` defines for the `reuseAddress` option.

s-ludwig:
I would propose to add a new `defaults` enum member, analoguous to `TCPListenOptions` and `StreamListenOptions`. It makes sense to have this as an additive option in terms of consistency, but switching default values in multiple places doesn't seem to be a good long-term idea either.

PR #2312, opened 2019-05-28
Update Travis configuration

Geod24:
This does a couple of cleanup to the travis configuration.
See the linked blog posts for the non-D-specific parts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: d
-dist: trusty
+dist: xenial
 
 services:
   - mongodb
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: d
-sudo: false
 dist: trusty
 
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,9 @@ d:
   # this way the overall test time gets cut down (GDC/LDC are a lot
   # slower tham DMD, so they should be started early), while still
   # catching most DMD version related build failures early
-  - dmd-2.085.0
-  - dmd-2.076.1
+  - dmd-2.086.0
+  - dmd-2.077.1
+  - ldc-1.15.0
   - ldc-1.14.0
   - ldc-1.13.0
   - ldc-1.12.0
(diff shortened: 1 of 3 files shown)


Geod24:
Hum, there's a lot of duplication in the readme as to what's supported.
Shouldn't we state the goal and point to the travis file instead ?

Geod24:
Well I give up on GDC for the time being...

.travis.yml
```diff
@@ -34,15 +34,14 @@ d:
   - ldc-1.9.0
   - ldc-1.8.0
   - ldc-1.7.0
-  - ldc-1.6.0
+  - dmd-2.085
```
s-ludwig:
```suggestion
  - dmd-2.085.1
```

s-ludwig:
There is also appveyor.yml with an additional set of duplicate entries, but since updates are relatively rare, it didn't really bother me so far.

s-ludwig:
Looks like the DMD 2.085.1/vibe-core failure is genuine, unfortunately, I'd suggest to simply allow it to fail for now:
```
matrix:
    allow_failures:
        - d: 2.085.1
          env: VIBED_DRIVER=vibe-core
```
(hopefully got the syntax right there)

s-ludwig:
Looks like the DMD 2.085.1/vibe-core failure is genuine, unfortunately, I'd suggest to simply allow it to fail for now:
```
matrix:
    allow_failures:
        - d: 2.085.1
          env: VIBED_DRIVER=vibe-core
```
(hopefully got the syntax right there)

Geod24:
@s-ludwig : That didn't work. Did a bit of digging and all parts of the build need to be matched, so I added `PARTS` too but it didn't work either. Why don't we just stick to v2.085.0 ? v2.085.1 is not going to be fixed anyway.

s-ludwig:
Okay, not worth the trouble then, lets stick with 2.085.0.

s-ludwig:
Finally green 🎉 That looked like a lot more effort than it should be, thanks for finishing this!

PR #2313, opened 2019-05-28
Deprecate indexOfCT and use indexOf instead

Geod24:
This seems to be a relic from the time when indexOf was not CTFEable.

I can adjust the time of removal to your needs - I didn't knew if `v0.8.11` was a good idea :)

### DIFF
--- a/inet/vibe/inet/url.d
+++ b/inet/vibe/inet/url.d
@@ -116,7 +116,7 @@ struct URL {
 		auto str = url_string;
 		enforce(str.length > 0, "Empty URL.");
 		if( str[0] != '/' ){
-			auto idx = str.indexOfCT(':');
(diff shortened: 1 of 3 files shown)


Geod24:
@wilzbach : What's the hold out on this ? If it's the deprecation, I'd rather remove the exact date and move forward. There is a *lot* of string processing code that needs some cleanup, and I'd rather get started before we get to the next release (E.g. next on my list is the [always allocating, Exception-happy sanitizeUTF8](https://github.com/vibe-d/vibe.d/blob/c38658769d761bf73e340c056181440548a1c9a7/utils/vibe/utils/string.d#L24-L46)).

wilzbach:
> What's the hold out on this ? If it's the deprecation

No, it's time. Things are currently very busy.

PR #2315, opened 2019-05-28
Remove usage of deprecated features in tests/examples

Geod24:


Geod24:
Note: This is only for things deprecated inside of Vibe.d
There is one usage of stopwatch that needs to be changed, but I didn't want to deal with the CI...

PR #2327, opened 2019-06-24
Add createIndex, getIndexes and improve docstrings

Member1221:
This PR adds createIndex, getIndex and improves some of the index related documentation strings.

The strings are taken (and slightly modified) info from the mongodb documentation.

This is a part of the way to getting https://github.com/vibe-d/vibe.d/issues/2200 done

### DIFF
--- a/mongodb/vibe/db/mongo/collection.d
+++ b/mongodb/vibe/db/mongo/collection.d
@@ -469,6 +469,7 @@ struct MongoCollection {
 		if (flags & IndexFlags.expireAfterSeconds) doc["expireAfterSeconds"] = expire_time.total!"seconds";
 		database["system.indexes"].insert(doc);
 	}
+
 	/// ditto
 	deprecated("Use the overload taking an array of field_orders instead.")
 	void ensureIndex(int[string] field_orders, IndexFlags flags = IndexFlags.none, ulong expireAfterSeconds = 0)
@@ -479,6 +480,9 @@ struct MongoCollection {
 		ensureIndex(orders, flags, expireAfterSeconds.seconds);
 	}
 
+	/**
+		Drops or removes the specified index from the collection.
+	*/
 	void dropIndex(string name)
 	@safe {
 		static struct CMD {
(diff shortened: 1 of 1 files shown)


mongodb/vibe/db/mongo/collection.d
```diff
@@ -493,6 +497,46 @@ struct MongoCollection {
...
+	@safe {
+		static struct CMD {
+			string listIndexes;
+		}
+
+		CMD cmd;
+		listIndexes = m_name;
+
+		auto reply = database.runCommand(cmd);
+		return reply;
```
WebFreak001: I think this returns the raw protocol response, it should probably return the data value it returns instead? The raw response Bson is probably not interesting for the user if they use a high level method for this.

https://docs.mongodb.com/manual/reference/command/listIndexes/#listIndexes.cursor

The docs say this returns a cursor so the return type of the function should be a cursor too.

mongodb/vibe/db/mongo/collection.d
```diff
@@ -493,6 +497,46 @@ struct MongoCollection {
...
+		Returns an array that holds a list of documents that identify and describe the existing indexes on the collection. 
+	*/
+	R getIndexes(T = Bson, R = Bson)() 
+	@safe {
+		static struct CMD {
+			string listIndexes;
+		}
+
+		CMD cmd;
+		listIndexes = m_name;
```
WebFreak001: isn't this missing `cmd.`?

mongodb/vibe/db/mongo/collection.d
```diff
@@ -493,6 +497,46 @@ struct MongoCollection {
...
+
+		auto reply = database.runCommand(cmd);
```
WebFreak001: the "ok" field should be checked and a mongo exception should be thrown on failure (see the other methods)

mongodb/vibe/db/mongo/collection.d
```diff
@@ -493,6 +497,46 @@ struct MongoCollection {
 		enforce(reply["ok"].get!double == 1, "dropIndex command failed: "~reply["errmsg"].opt!string);
 	}
 
+	/**
+		Creates indexes on the collection.
+	*/
+	void createIndex(T)(T query)
```
WebFreak001: ensureIndex already exists, which is basically this command but using the old API. Is there a reason for this new API? I think there are new index types which we need to support but I don't really like the current API you implemented for this (accepting literally anything as argument, which is called unintuitively "query")

Member1221: https://docs.mongodb.com/manual/reference/method/db.collection.ensureIndex/#db.collection.ensureIndex ensureIndex is deprecated

WebFreak001: the current ensureIndex implementation actually directly writes into the indices document in mongodb, so it's really just what it's called in vibe.d currently. I don't really like this approach but maybe extending and improving ensureIndex would be a better idea than creating this new API.

Also both methods should add tests!

Member1221: imo it should be high priority to reflect mongodb's standard API layout within vibe-d's mongodb implementation so that it's easy for people to understand it.

WebFreak001:
the API should be implemented according to the standard index API: https://github.com/mongodb/specifications/blob/edb7e12e12ec5cea6536d4c17fbea3a0704d798d/source/index-management.rst#common-api-components

This means that for createIndexes:
```
   * For MongoDB 2.6 and higher this method MUST execute a createIndexes command.
   *
   * For MongoDB 2.4 this method MUST insert the index specifications directly into
   * the system.indexes collection. The write concern provided provided to the server
   * MUST be { w: 1 }.
```

which means this PR is dependent on #2201

mongodb/vibe/db/mongo/collection.d
tirithen: @WebFreak001 I can no longer use the ensureIndex method with my mongodb server (version 4.2.1) as the indexes are no longer positioned at the place ensureIndex is trying to write to. Please make the standard createIndex available as there is no reason to re-invent the wheel here.

s-ludwig:
Considering that current MongoDB versions do not work with the direct insertion approach anymore, I'll merge this now as an immediate fix (should have been earlier, but I just stumbled over this now), but will try to look at improving the API before the next version gets tagged.

WebFreak001:
@s-ludwig I started the API conforming to the spec a while ago, I can PR what I have later today. Actually made that `@embedNullable` serialization PR mainly to make the implementation of the API simpler with patching structs

s-ludwig:
Sounds good. I'll try to review ASAP.

PR #2332, opened 2019-07-02
Change ChunkedInputStream to not block right after chunk is read

tchaloupka:
This is probably not an ideal solution but it at least solves my problem - that I need to process chunked stream data right when it arrives and not when the next chunk is available.


http/vibe/http/common.d
```diff
@@ -398,11 +399,17 @@ final class ChunkedInputStream : InputStream
...
 	}
 
-	@property bool empty() const { return m_bytesInCurrentChunk == 0; }
+	@property bool empty() { return bytesInCurrentChunk == 0; }
 
-	@property ulong leastSize() const { return m_bytesInCurrentChunk; }
+	@property ulong leastSize() { return bytesInCurrentChunk; }
 
-	@property bool dataAvailableForRead() { return m_bytesInCurrentChunk > 0 && m_in.dataAvailableForRead; }
+	@property bool dataAvailableForRead() { return bytesInCurrentChunk > 0 && m_in.dataAvailableForRead; }
```
s-ludwig: This must stay as `m_bytesInCurrentChunk`, so that `dataAvailableForRead` does not block (its purpose is to be able to tell whether a read will block).

http/vibe/http/common.d
```diff
@@ -398,11 +399,17 @@ final class ChunkedInputStream : InputStream
 		readChunk();
```
s-ludwig: It probably makes sense to take the opportunity and remove this as well, so that the stream is completely lazy.

http/vibe/http/common.d
s-ludwig:
```suggestion
	@property bool dataAvailableForRead() { return m_bytesInCurrentChunk > 0 && m_in.dataAvailableForRead; }
```

http/vibe/http/common.d
s-ludwig:
```suggestion
```

Geod24:
@tchaloupka Ping

tchaloupka:
I've made the proposed changes.
Sorry for the delay - time flies..

PR #2334, opened 2019-07-07
Collection of code cleanup

Geod24:
I was originally intending to clean up some `utils` code, and most importantly remove `sanitizeUTF8`, which `std.utf` can do quite well.
However I ended up finding a compiler bug and some unused fields, and found out that the approach I'd like to use wouldn't really work without a few bug fixes...
Best reviewed commit-by-commit. Each commit can be taken separately and compiled.


Geod24:
@wilzbach ?

PR #2337, opened 2019-07-20
WebSocket.close: Silently ignore exceptions for close message

v1ne:
See commit message for details.

### DIFF
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -701,10 +701,11 @@ final class WebSocket {
 		if (connected) {
 			send((scope msg) {
 				m_sentCloseFrame = true;
-				if (code != 0)
+				if (code != 0) {
 					msg.write(std.bitmanip.nativeToBigEndian(code));
 					msg.write(cast(const ubyte[])reason);
-				}, FrameOpcode.close);
+				}
+			}, FrameOpcode.close);
 		}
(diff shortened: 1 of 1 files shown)


v1ne:
@s-ludwig: Do you think this change is at the right place? This issue itself is quite annoying, even though I can only reproduce it stochasticly.

http/vibe/http/websockets.d
```diff
@@ -699,12 +699,14 @@ final class WebSocket {
...
-					msg.write(cast(const ubyte[])reason);
-				}, FrameOpcode.close);
+			try {
+				send((scope msg) {
+					m_sentCloseFrame = true;
+					if (code != 0)
+						msg.write(std.bitmanip.nativeToBigEndian(code));
+						msg.write(cast(const ubyte[])reason);
+					}, FrameOpcode.close);
+			} catch (Exception) {}
```
s-ludwig: Would be good to add a `logDiagnostic("Failed to send active web socket close frame: %s", e.msg);` here.

v1ne: I'd consider this normal behaviour, but I added the log message, sure.
Also, I added a commit where I fix the close reply (the indentation has been misleading…).

v1ne:
@s-ludwig: What do you think?

s-ludwig:
Sorry, I lost track of this, but it looks good now!

PR #2338, opened 2019-07-21
Fix deprecation warnings in vibe.data.json

ghost91-:


ghost91-:
This can probably not be merged at the moment because it breaks building with DMD < 2.082.0, the version where the new identifiers have been added and the old ones have been deprecated.

As soon as we can drop support for DMD < 2.082.0, this should be merged.

wilzbach:
Why not use static if and alias the old declarations to the new ones?

ghost91-:
@wilzbach Do you mean something like this:
```d
static if(__VERSION__ >= 2082)
{
	import std.json : JSONValue, JSONType;
}
else
{
	import std.json : JSONValue, JSON_TYPE;

	enum JSONType : byte
	{
		null_ = JSON_TYPE.NULL,
		string = JSON_TYPE.STRING,
		integer = JSON_TYPE.INTEGER,
		uinteger = JSON_TYPE.UINTEGER,
		float_ = JSON_TYPE.FLOAT,
		array = JSON_TYPE.ARRAY,
		object = JSON_TYPE.OBJECT,
		true_ = JSON_TYPE.TRUE,
		false_ = JSON_TYPE.FALSE,
	}
}
```
?

This would actually be OK for the current use case (there is only a final switch statement using `JSONType`), but it would not work if there any Phobos (or other library) functions are called which return a `JSON_TYPE` or expect it as a parameter.

The problem is that you cannot inject new aliases into the existing `enum`. Another solution would be to define our own `alias`es and only use those. Something along the lines of this:
```d
static if(__VERSION__ >= 2082)
{
	import std.json : JSONValue, JSONType;
	alias jsonTypeNull = JSONType.null_;
	/* ... */
}
else
{
	import std.json : JSONValue, JSONType = JSON_TYPE;
	alias jsonTypeNull = JSONType.NULL;
	/* ... */
}
/* ... */
void foo(JSONType type)
{
	final switch(type)
	{
		case jsonTypeNull: /* ... */
	}
}
```
However, this would be quite a lot of additional code and effort with only small benefit... What do you think about this?

wilzbach:
> @wilzbach Do you mean something like this:

Yup. :+1: 

> but it would not work if there any Phobos (or other library) functions are called which return a JSON_TYPE or expect it as a parameter.

It's only used once here to provide compatibility between std.json and vibe.data.json, so I think we should be fine here.

ghost91-:
force pushed to retrigger CI (one of the jobs failed because of a dub registry 503, all other jobs were successful).

PR #2342, opened 2019-08-05
Close TCPConnection on failure

gedaiu:
I'm not sure if this is the best fix, but it looks like the TCPConnection state is connected after it fails to connect here: `m_conn = connectTCP(proxyAddr, m_settings.networkInterface);`

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -569,7 +569,8 @@ final class HTTPClient {
 	{
 		assert(!m_requesting, "Interleaved HTTP client requests detected!");
 		assert(!m_responding, "Interleaved HTTP client request/response detected!");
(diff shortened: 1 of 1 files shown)


s-ludwig:
There is a `try`-`catch`  around the TLS stream construction, which seems to be the only place in the upper part of the function that can throw. It should be sufficient to put the additions into the closure at the bottom.  `scope (failure)` is a bit problematic, though, because unfortunately it catches any `Throwable` as opposed to just `Exception`, so using an eplicit `try`-`catch` would be better here.

gedaiu:
I found this issue on http connections, where there was no tls context. As I remember, the exception that I got was from vibe-core.

PR #2343, opened 2019-08-06
add support to rtsp url

tchaloupka:


tchaloupka:
damn, one test fails with timeout :)

PR #2344, opened 2019-08-06
Adds HTTP client timeout setting

denizzzka:
Fixes https://github.com/vibe-d/vibe.d/issues/1375

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -352,12 +352,6 @@ unittest {
 		sw.stop;
 		isThrown = true;
 
-		version(VibeLibasyncDriver)
-		{
-			if(!e.msg.canFind("timeout"))
-				throw e;
-		}
-
 		if(sw.peek > 1.seconds) // request should fail ~immediately
 			throw e;
 	}
@@ -758,7 +752,7 @@ private auto connectTCPWithTimeout(NetworkAddress addr, NetworkAddress bind_addr
 {
 	auto conn = vibe.core.net.connectTCP(addr, bind_address);
 
-	if(timeout != Duration.max)
+	if(isTcpConnectionTimeoutEnabled && timeout != Duration.max)
 		conn.readTimeout = timeout;
 
 	return conn;
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -673,7 +673,7 @@ final class HTTPClient {
 
 				NetworkAddress proxyAddr = resolveHost(m_settings.proxyURL.host, m_settings.dnsAddressFamily, use_dns);
 				proxyAddr.port = m_settings.proxyURL.port;
-				m_conn = connectTCP(proxyAddr, m_settings.networkInterface, m_settings.tcpConnectionTimeout);
+				m_conn = connectTCPWithTimeout(proxyAddr, m_settings.networkInterface, m_settings.tcpConnectionTimeout);
 			}
 			else {
 				version(UnixSocket)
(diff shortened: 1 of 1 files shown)


http/vibe/http/client.d
```diff
@@ -363,6 +380,7 @@ final class HTTPClient {
 	*/
 	void connect(string server, ushort port = 80, bool use_tls = false, const(HTTPClientSettings) settings = defaultSettings)
 	{
+		//FIXME: actually this function is not connects, but destroys existing connection
```
s-ludwig: This should be removed. Maybe the method name should eventually be changed to `setConnectionTarget` or something along those lines, but the implementation is working as intended (and it has nothing to do with the other changes). BTW, although it doesn't physically open a TCP connection, "connect" is in line with the BSD socket API when using datagram based sockets, which are in a way similar to a request based protocol like HTTP.

denizzzka: Ok, sure

 It just cost me a bit of time and I decided to pay attention to this fact. I’ll delete it now.

denizzzka:
@s-ludwig I am sorry for distracting you. What do you think about the last commit?

Currently, I’m not ready to go deep into fix of the libasync driver itself. I just want to use vibe-core.

Maybe there are better ideas?

denizzzka:
Wrote everything so that driver users which have timeouts unsupported see that they have such ability - just need to slightly fix this drivers.

denizzzka:
CI tests passed, with the exception of 3 pcs due to network problems on the CI server side

denizzzka:
Still one win32 test failed due to CI side. Can it be restarted?

wilzbach:
Yes. just tried to do so, but it looks like only the whole job can be restarted :/

s-ludwig:
It should be okay to simply document that only vibe-core currently supports connection timeouts and then use `version (Have_vibe_core)` to handle the `tcpConnectionTimeout` setting (leaving it as a normal field and possibly emitting a runtime warning if it is not `Duration.max`). 

BTW, what exactly is the intent of this change, to define the `connect` timeout, a read timeout, or both? I was assuming the first, but currently the second is implemented.

s-ludwig:
@wilzbach They have finally added a `re-run incomplete` action a while ago (which really is `re-run failed`)

denizzzka:
Please restart one failed test if you can?

s-ludwig:
@denizzzka: Even though this has already been merged, can you clarify my last question? This doesn't seem to do what the API states.

Geod24:
The intent is to define a read timeout, so that it can be used with `RestInterfaceClient`

s-ludwig:
Okay, then I'll rename it to "read timeout" to make that clear.

The initial implementation that passed the timeout to `connectTCP` was defining a connect timeout and no read timeout, which added to the confusion.

denizzzka:
> It should be okay to simply document that only vibe-core currently supports connection timeouts and then use version (Have_vibe_core) to handle the tcpConnectionTimeout setting (leaving it as a normal field and possibly emitting a runtime warning if it is not Duration.max).

I do not agree with this proposal, it complicates. My thoughts:
Really I do not know how many drivers are implemented in vibe.d, but I know for sure that only two of them are broken and these two are highlighted by version directives.

If fix will be delayed, but someone wants to add a new driver, then it is better to let him immediately see that he still needs to implement timeouts in order to pass the tests.

> BTW, what exactly is the intent of this change, to define the connect timeout, a read timeout, or both? I was assuming the first, but currently the second is implemented.

Yes, at first I also thought of doing the first, but the second was implemented only. That's right, you can rename it.

denizzzka:
> The intent is to define a read timeout, so that it can be used with RestInterfaceClient

Could there be necessary timeout when write too? Maybe it is need to implement read+write timeout on sockets?

For example, some kind of stuck with write buffers at busied system will not forces more and more new connections to be created and then wait endlessly for write to it?

s-ludwig:
I've opened a PR that addresses by concerns and also introduces a "connect" timeout: #2347

W.r.t. making the timeout a constant on buggy driver implementation vs. emitting a runtime warning - making it a constant would require application developers to make very unobvious compile time switches if they still want to be able to compile against one of those drivers. This would go against the basic idea of providing a uniform API across platforms. We are also talking about a feature that is usually non-critical, and the runtime warning that gets emitted for every connection should also be very noticable.

Agreed about adding a configurable write timeout. Both, read and write timeouts should also be reviewed in terms of their semantics (does the full operation time out, or the time until *some* data arrives or gets sent).

Issue #2209, opened 2018-09-11
Deserialization by value is not checked against enum values


cspencer51:
The following code fails with: `core.exception.SwitchError@source/app.d(17): No appropriate switch clause found` because the deserialization code casts the value to the Enum type but does not check the cast value against the possible enum values (https://github.com/vibe-d/vibe.d/blob/master/data/vibe/data/serialization.d#L601).

```
import vibe.d;

enum A : string
{
    Type1 = "one"
}

void main()
{
    struct MyStruct
    {
        A thing;
    }

    MyStruct x = deserializeJson!MyStruct(`{"thing": "nonsense"}`);

    final switch(x.thing)
    {
        case A.Type1:
            break;
    }
}

```

PR #2345, opened 2019-08-06
fix enum handling

madmax-inc:
Hey there!

Seems the enum values handling in (de)serialization is inaccurate right now. One should probably throw an exception if input isn't a valid enum member.

Relates: #2209

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1771,7 +1771,7 @@ unittest {
 	See_Also: vibe.data.serialization.serialize, vibe.data.serialization.deserialize, serializeToJson, deserializeJson
 */
 struct JsonSerializer {
-	template isJsonBasicType(T) { enum isJsonBasicType = std.traits.isNumeric!T || isBoolean!T || isSomeString!T || is(T == enum) || is(T == typeof(null)) || is(Unqual!T == UUID) || isJsonSerializable!T; }
+	template isJsonBasicType(T) { enum isJsonBasicType = std.traits.isNumeric!T || isBoolean!T || isSomeString!T || is(T == typeof(null)) || is(Unqual!T == UUID) || isJsonSerializable!T; }
 
 	template isSupportedValueType(T) { enum isSupportedValueType = isJsonBasicType!T || is(Unqual!T == Json) || is(Unqual!T == JSONValue); }
 
@@ -1867,22 +1867,6 @@ struct JsonSerializer {
 				case Json.Type.undefined: return T.nan;
 				case Json.Type.float_: return cast(T)m_current.get!double;
 				case Json.Type.bigInt: return cast(T)m_current.bigIntToLong();
-			}
-		} else static if (is(T == enum)) {
-			const T value = () {
-				static if (hasPolicyAttributeL!(ByNameAttribute, Traits.Policy, Traits.Attributes)) {
-					return readValue!(Traits, string)().to!T();
-				} else {
-					return cast(T)readValue!(Traits, OriginalType!T)();
-				}
-			}();
-
-			switch (value) {
-				static foreach (enumvalue; EnumMembers!T) {
-					case enumvalue: return value;
-				}
-				default:
-					throw new JSONException("Unexpected enum value " ~ value.to!string());
 			}
 		} else static if (is(T == const(char)[])) {
 			return readValue!(Traits, string);
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1771,7 +1771,7 @@ unittest {
 	See_Also: vibe.data.serialization.serialize, vibe.data.serialization.deserialize, serializeToJson, deserializeJson
 */
 struct JsonSerializer {
-	template isJsonBasicType(T) { enum isJsonBasicType = std.traits.isNumeric!T || isBoolean!T || isSomeString!T || is(T == typeof(null)) || is(Unqual!T == UUID) || isJsonSerializable!T; }
+	template isJsonBasicType(T) { enum isJsonBasicType = std.traits.isNumeric!T || isBoolean!T || isSomeString!T || is(T == enum) || is(T == typeof(null)) || is(Unqual!T == UUID) || isJsonSerializable!T; }
 
 	template isSupportedValueType(T) { enum isSupportedValueType = isJsonBasicType!T || is(Unqual!T == Json) || is(Unqual!T == JSONValue); }
 
@@ -1867,6 +1867,22 @@ struct JsonSerializer {
 				case Json.Type.undefined: return T.nan;
 				case Json.Type.float_: return cast(T)m_current.get!double;
 				case Json.Type.bigInt: return cast(T)m_current.bigIntToLong();
+			}
+		} else static if (is(T == enum)) {
+			const T value = () {
+				static if (hasPolicyAttributeL!(ByNameAttribute, Traits.Policy, Traits.Attributes)) {
+					return readValue!(Traits, string)().to!T();
+				} else {
+					return cast(T)readValue!(Traits, OriginalType!T)();
+				}
+			}();
+
+			switch (value) {
+				static foreach (enumvalue; EnumMembers!T) {
+					case enumvalue: return value;
+				}
+				default:
+					throw new JSONException("Unexpected enum value " ~ value.to!string());
 			}
 		} else static if (is(T == const(char)[])) {
 			return readValue!(Traits, string);
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -285,6 +285,26 @@ unittest {
 	assert(test.text == "Hello");
 }
 
+unittest {
+	import vibe.data.json;
+	import std.exception;
+
+	enum Testable : string {
+		foo = "foo",
+		bar = "bar"
+	}
+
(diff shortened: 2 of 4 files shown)


web/vibe/web/rest.d
```diff
@@ -1444,12 +1445,16 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
 						catch (JSONException e) enforceBadRequest(false, e.msg);
+						catch (SerializationLogicError e) enforceBadRequest(false, e.msg);
 					} else static if (sparam.kind == ParameterKind.body_) {
 						try {
 							if (auto pv = fieldname in req.json)
 								v = deserializeJson!PT(*pv);
-						} catch (JSONException e)
+						} catch (JSONException e) {
 							enforceBadRequest(false, e.msg);
+						} catch (SerializationLogicError e) {
```
Geod24: What's the point of introducing a new exception type if the handling is exactly the same ?
Also, it's an `Exception`, not an `Error`. In D, `XXXError` are classes derived from `Error` by convention.

madmax-inc: > What's the point of introducing a new exception type if the handling is exactly the same ?

Well a custom exception class is here because it's being thrown from a generic deserialization routine (not a JSON one). The same logic works well on a `BsonSerializer` or any other custom serializer.
So the exception definitely should not be a `JSONException`.

madmax-inc: > Also, it's an Exception, not an Error. In D, XXXError are classes derived from Error by convention.

Ok then. I'll rename it into `SerializationLogicException`. Thanks :)

Geod24: > So the exception definitely should not be a `JSONException`.

Except, any client will not want to catch both `JSONException` and `SerializationLogicException`.  So we're forcing callers to handle 2 exception type for an internal detail. That's a pretty bad API IMO.

madmax-inc: > > So the exception definitely should not be a `JSONException`.
> 
> Except, any client will not want to catch both `JSONException` and `SerializationLogicException`. So we're forcing callers to handle 2 exception type for an internal detail. That's a pretty bad API IMO.

Agree. Do you think it'll be a good decision to wrap the underlying exception everywhere earlier?
```d
...
try {
  return ser.deserializeValue!(T)(...);
} catch (Exception e) {
  throw new SerializationLogicException(...);
}
```

Or maybe we should introduce the same enum-handling logic in every serializer implementation?

Geod24: I'd say the logic should go in the serializer.
It's not a great solution, but it's the best option I can think of.

Bear in mind that someone could very well be calling `std.conv.to` and get a `ConvException` so you can never really rely on the exception type anyway.

By the way catch-and-rethrow is quite inefficient, not something we'd want in a high performance web server.

madmax-inc: Ok then let's place it into serializer and throw well-known `JSONException` manually. I agree it's an acceptable solution which will allow us not to check the whole codebase against my changes.

I'll return with my changes soon. Thanks )

Geod24: Thanks for taking care of this. Let me know when this is ready for another review.

madmax-inc: @Geod24 Hey, I copied the logic into a serializer. Now the `JSONSerializer` is responsible for parsing enum values while any other implementation will use the old generic logic.
The bad news is that there is also a `@byName` handling logic which was also copied =(

Have you got an idea how to make the solution better without bloating the serializer "interface"?

data/vibe/data/serialization.d
```diff
@@ -1262,7 +1282,7 @@ unittest {
 	static assert(!hasAttributeL!(AsArrayAttribute!DefaultPolicy, byName));
 }
 
-private template hasPolicyAttributeL(alias T, alias POLICY, ATTRIBUTES...)
+template hasPolicyAttributeL(alias T, alias POLICY, ATTRIBUTES...)
```
s-ludwig: I'd keep this `package` for now to avoid public dependencies.

data/vibe/data/json.d
```diff
@@ -1868,6 +1868,22 @@ struct JsonSerializer {
 				case Json.Type.float_: return cast(T)m_current.get!double;
 				case Json.Type.bigInt: return cast(T)m_current.bigIntToLong();
 			}
+		} else static if (is(T == enum)) {
+			const T value = () {
```
s-ludwig: Since `static if` doesn't create a scope, this could be shortened to just
```
			static if (hasPolicyAttributeL!(ByNameAttribute, Traits.Policy, Traits.Attributes)) {
				const T value = readValue!(Traits, string)().to!T();
			} else {
				const T value = cast(T)readValue!(Traits, OriginalType!T)();
			}
```

Maybe I need to look closer, but shoudn't the `to!T` throw a `ConvException` in the `string` case, circumventing the `JSONException` in the bottom?

data/vibe/data/serialization.d
```diff
@@ -285,6 +285,26 @@ unittest {
...
+		deserialize!(JsonSerializer, Testable)(Json(value));
+	}
+
+	foreach (string val; ["foo", "bar"]) {
+		assertNotThrown(deserializeString(val));
+	}
+
+	assertThrown!JSONException(deserializeString("foobar"));
+}
+
```
s-ludwig: Looks like the test should be in the `vibe.data.json` module.

data/vibe/data/serialization.d
```diff
@@ -597,14 +617,14 @@ private template deserializeValueImpl(Serializer, alias Policy) {
 		static if (isPolicySerializable!(Policy, T)) {
 			alias CustomType = typeof(Policy!T.toRepresentation(T.init));
 			return Policy!T.fromRepresentation(ser.deserializeValue!(CustomType, ATTRIBUTES));
+		} else static if (Serializer.isSupportedValueType!T) {
```
s-ludwig: I wonder whether `isSupportedValueType!T` wrongly returns true for some of the serializers (e.g. due to `is(T : long)` or a similar condition).

s-ludwig:
W.r.t. exceptions, the current situation is pertty messy. I never really added an exception hierarchy to vibe.d so far, as I was hoping that one would emerge in Phobos over time, and merging two exception hierarchies seems impossible without either silent breakage or multiple inheritance. Of course that never really happened and my ambitions of contributing this kind of high level change to Phobos are simply far too small these days.

Since `JSONException` complicates the general solution here, I would suggest to stick to just `ConvException`, to be in line with the `to!T`, and put a proper exception specification for the serialization module on the TODO list. An idea would be to generally let the serialzation functions only throw a `SerializationException`, or a sub type thereof, and chain the original exception internally, should it be of a different type.

data/vibe/data/json.d
madmax-inc: You're right, it should throw. I wonder if it is used now =) It should generate a 500 response I suppose.

madmax-inc:
> W.r.t. exceptions, the current situation is pertty messy. I never really added an exception hierarchy to vibe.d so far, as I was hoping that one would emerge in Phobos over time, and merging two exception hierarchies seems impossible without either silent breakage or multiple inheritance. Of course that never really happened and my ambitions of contributing this kind of high level change to Phobos are simply far too small these days.
> 
> Since `JSONException` complicates the general solution here, I would suggest to stick to just `ConvException`, to be in line with the `to!T`, and put a proper exception specification for the serialization module on the TODO list. An idea would be to generally let the serialzation functions only throw a `SerializationException`, or a sub type thereof, and chain the original exception internally, should it be of a different type.

Thanks for your review. Do you think we should catch both `JSONException` and `ConvException` (while throwing `ConvException` manually in the second case) like it was in the previous revision (https://github.com/vibe-d/vibe.d/pull/2345/commits/17f15e31467e58c3b57ca51aa337584135d82b7c#diff-473d2bce6de3149a10a70497fc010d05R1447) ?

Btw @Geod24 mentioned that it's extremely inefficient to catch-and-rethrow an exception.

data/vibe/data/serialization.d
madmax-inc: I suppose it should stay here if we decide to move the logic back into a `vibe.data.serialization` module ) Let's discuss below.

madmax-inc: I left it here since the serialization logic is also in the same module.

madmax-inc:
ping @Geod24 , @s-ludwig could you please review once again? Thx =)

data/vibe/data/serialization.d
```diff
@@ -601,7 +645,15 @@ private template deserializeValueImpl(Serializer, alias Policy) {
 			static if (hasPolicyAttributeL!(ByNameAttribute, Policy, ATTRIBUTES)) {
 				return ser.deserializeValue!(string, ATTRIBUTES).to!T();
 			} else {
-				return cast(T)ser.deserializeValue!(OriginalType!T);
+				const auto value = cast(T)ser.deserializeValue!(OriginalType!T);
+
+				switch (value) {
+					static foreach (enumvalue; EnumMembers!T) {
+						case enumvalue: return value;
```
schveiguy: This is going to have a problem with enums that have repeated members. This doesn't mean the value is invalid, just ambiguous.

PR #2346, opened 2019-08-10
Code deduplication + IPv6 HTTP client fix

denizzzka:
* Removes multiple copy-and-pasted lines from HTTP client module.

* Fixes ommited support of IPv6 addresses in two of four requestHTTP fuctions

* Adds module vibe.http.auth.basic_auth_client to avoid cyclic dependency of ctors/dtors between HTTP client and server code.

* @safe coverage slightly increased

### DIFF
--- a/http/vibe/http/auth/basic_auth.d
+++ b/http/vibe/http/auth/basic_auth.d
@@ -105,9 +105,8 @@ bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pw
 	return false;
 }
 
-static import vibe.http.auth.basic_auth_client;
+static import vibe.http.internal.basic_auth_client;
 
-deprecated("It will be removed in 0.9. Use vibe.http.auth.basic_auth_client.addBasicAuth instead.")
-alias addBasicAuth = vibe.http.auth.basic_auth_client.addBasicAuth;
+alias addBasicAuth = vibe.http.internal.basic_auth_client.addBasicAuth;
 
 alias PasswordVerifyCallback = bool delegate(string user, string password);
--- a/http/vibe/http/auth/basic_auth.d
+++ b/http/vibe/http/auth/basic_auth.d
@@ -105,4 +105,9 @@ bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pw
 	return false;
 }
 
+static import vibe.http.auth.basic_auth_client;
+
+deprecated("It will be removed in 0.9. Use vibe.http.auth.basic_auth_client.addBasicAuth instead.")
+alias addBasicAuth = vibe.http.auth.basic_auth_client.addBasicAuth;
+
 alias PasswordVerifyCallback = bool delegate(string user, string password);
--- a/http/vibe/http/auth/basic_auth.d
+++ b/http/vibe/http/auth/basic_auth.d
@@ -105,15 +105,4 @@ bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pw
 	return false;
 }
 
-
-/**
-	Augments the given HTTP request with an HTTP Basic Auth header.
-*/
-void addBasicAuth(scope HTTPRequest req, string user, string password)
-{
-	string pwstr = user ~ ":" ~ password;
-	string authstr = () @trusted { return cast(string)Base64.encode(cast(ubyte[])pwstr); } ();
-	req.headers["Authorization"] = "Basic " ~ authstr;
-}
-
 alias PasswordVerifyCallback = bool delegate(string user, string password);
--- a/http/vibe/http/auth/basic_auth_client.d
+++ b/http/vibe/http/auth/basic_auth_client.d
@@ -1,0 +1,21 @@
+/**
+	Implements HTTP Basic Auth for client.
+
+	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
+*/
+module vibe.http.auth.basic_auth_client;
+
+import vibe.http.common;
+import std.base64;
+
+@safe:
+
+/**
+	Augments the given HTTP request with an HTTP Basic Auth header.
+*/
+void addBasicAuth(scope HTTPRequest req, string user, string password)
+{
+	string pwstr = user ~ ":" ~ password;
+	string authstr = () @trusted { return cast(string)Base64.encode(cast(ubyte[])pwstr); } ();
+	req.headers["Authorization"] = "Basic " ~ authstr;
+}
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -232,6 +232,13 @@ static ~this()
 private struct ConnInfo { string host; ushort port; bool useTLS; string proxyIP; ushort proxyPort; NetworkAddress bind_addr; }
 private static vibe.utils.array.FixedRingBuffer!(Tuple!(ConnInfo, ConnectionPool!HTTPClient), 16) s_connections;
 
+// TODO: Connection timeout argument is accepted by libasync and win32 drivers but not affects to anything by some reason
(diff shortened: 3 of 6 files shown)


http/vibe/http/client.d
```diff
@@ -182,6 +150,29 @@ void requestHTTP(URL url, scope void delegate(scope HTTPClientRequest req) reque
...
 
+private bool isTlsNeed(in URL url, in HTTPClientSettings settings)
+{
+	version(UnixSocket) {
+		enforce(url.schema == "http" || url.schema == "https" || url.schema == "http+unix" || url.schema == "https+unix", "URL schema must be http(s) or http(s)+unix.");
+	} else {
+		enforce(url.schema == "http" || url.schema == "https", "URL schema must be http(s).");
+	}
+	enforce(url.host.length > 0, "URL must contain a host name.");
+	bool use_tls;
```
Geod24: Instead of creating a variable, you can just `return` when it is set, because the variable is only ever assigned to once

http/vibe/http/client.d
```diff
@@ -173,6 +134,38 @@ private bool isTlsNeed(in URL url, in HTTPClientSettings settings)
 	return use_tls;
 }
 
+//TODO: remove @trusted
+private void httpRequesterDg(scope HTTPClientRequest req, in URL url, in HTTPClientSettings settings, scope void delegate(scope HTTPClientRequest req) requester) @trusted
```
Geod24: If the `delegate` aren't `@safe`, this can't be marked `@trusted`, ever

http/vibe/http/auth/basic_auth.d
```diff
@@ -105,15 +105,4 @@ bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pw
 	return false;
 }
 
-
-/**
-	Augments the given HTTP request with an HTTP Basic Auth header.
-*/
-void addBasicAuth(scope HTTPRequest req, string user, string password)
```
Geod24: This is a public function. You can't just remove it. But you can introduce a `deprecated alias`

http/vibe/http/auth/basic_auth_client.d
```diff
@@ -0,0 +1,21 @@
+/**
+	Implements HTTP Basic Auth for client.
+
+	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
+*/
+module vibe.http.auth.basic_auth_client;
```
Geod24: Why the new module ? If it's only for one function I am not sure it's worth it.

http/vibe/http/client.d
```diff
@@ -156,14 +156,10 @@ private void httpRequesterDg(scope HTTPClientRequest req, in URL url, in HTTPCli
...
-		import std.base64;
-		string pwstr = url.username ~ ":" ~ url.password;
-		req.headers["Authorization"] = "Basic " ~
-			cast(string)Base64.encode(cast(ubyte[])pwstr);
-	}
+	if ("authorization" !in req.headers && url.username != "")
+		req.addBasicAuth(url.username, url.password);
 
-	if (requester) requester(req);
+	if (requester) () @trusted{ requester(req); } ();
```
Geod24: There's no point in using `@safe` if we trust a random external delegate.

http/vibe/http/client.d
denizzzka: This is already tagged as "Outdated".

I used @trusted temporary to not to make a mistake by making a big change right away

http/vibe/http/auth/basic_auth_client.d
denizzzka: It is need to avoid cyclic dependency of ctors/dtors between HTTP client and server modules what both use `addBasicAuth`

http/vibe/http/client.d
denizzzka: In fact, if look carefully, initially all of this code block was unsafe. Now only one line of it marked as trusted.

http/vibe/http/client.d
denizzzka: I want to avoid changes as much as possible so as not to add any error here. It's okay if such a variable stays here for a while. And personally I do not like many exit points.

http/vibe/http/auth/basic_auth.d
denizzzka: Fixed

http/vibe/http/auth/basic_auth.d
```diff
@@ -105,4 +105,12 @@ bool checkBasicAuth(scope HTTPServerRequest req, scope PasswordVerifyCallback pw
 	return false;
 }
 
+deprecated("It will be removed in 0.9. Use vibe.http.auth.basic_auth_client.addBasicAuth instead.")
+void addBasicAuth(scope HTTPRequest req, string user, string password)
```
Geod24: No need for a full fledged symbol, `deprecated alias` will do (and then you don't need to do the symbol rename in imported places).

denizzzka: Wow, I didn't know about such an opportunity. Fixed.

http/vibe/http/client.d
denizzzka: In other words, we previously trusted him too and nothing changed.

denizzzka:
One test failed on CI side again

http/vibe/http/client.d
```diff
@@ -136,6 +99,19 @@ void requestHTTP(string url, scope void delegate(scope HTTPClientRequest req) re
...
+	auto cli = connectHTTP(url.getFilteredHost, url.port, use_tls, settings);
+	cli.request(
+		(scope req){ httpRequesterDg(req, url, settings, requester); },
+		responder
+	);
+	assert(!cli.m_requesting, "HTTP client still requesting after return!?");
+	assert(!cli.m_responding, "HTTP client still responding after return!?");
+}
+
+private bool isTlsNeed(in URL url, in HTTPClientSettings settings)
```
s-ludwig: Naming: `isTLSNeeded` or `isTLSRequired`

http/vibe/http/client.d
s-ludwig: In that case you could add `@system` to `httpRequesterDg` instead. `@trusted` is bad, because it communicates that the contained code has been manually verified to be safe, which is of course not the case with a user supplied callback.

Of course the next step at some point will be to templatize the callback parameter and let the compiler infer the safety of the function (which currently requires putting it in a different scope or replacing the `@safe:` at the beginning of the file).

denizzzka:
> I'm not sure about the name of that internal module, since it will only contain one almost trivial function

Maybe some another functions will be moved to it too? I just moved one because compilation issue.
(But this change can be left for the future.)

s-ludwig:
> Maybe some another functions will be moved to it too? I just moved one because compilation issue.
(But this change can be left for the future.)

Agreed, but we can freely change the module name later as required, so it really just shouldn't become a reason to worry about now.

denizzzka:
> A more suitable solution may be to create a new vibe.http.internal.*

This solution may lead to same cyclic dependency in the future when some another functions will be moved to it. (Because internal.* sounds like that it should contain both client and server code)

s-ludwig:
I'm mostly concerned with the public API. The internal module is a little silly for the amount of code it contains, but that's fine if the alternative is to have a public module with the same issue, plus making a pointless breaking change.

denizzzka:
What if I am just remove `deprecated` from `alias addBasicAuth = vibe.http.auth.basic_auth_client.addBasicAuth;`? Will documentation also be rendered for `vibe.http.auth.basic_auth` module? In this case no one will notice anything and nothing will need to be changed in the user code.

s-ludwig:
Yes, it's still a normal public module in that case. The convention is that any `*.internal.*` modules will be hidden from the documentation.

PR #2347, opened 2019-08-11
Fixup HTTP client timeout

s-ludwig:
See #2344

### DIFF
--- a/core/vibe/core/drivers/libevent2_tcp.d
+++ b/core/vibe/core/drivers/libevent2_tcp.d
@@ -122,7 +122,7 @@ package final class Libevent2TCPConnection : TCPConnection {
 	@property void readTimeout(Duration v)
 	{
 		m_readTimeout = v;
-		if( v == dur!"seconds"(0) ){
+		if (v == dur!"seconds"(0) || v == Duration.max) {
 			() @trusted { bufferevent_set_timeouts(m_ctx.event, null, null); } ();
 		} else {
 			assert(v.total!"seconds" <= int.max);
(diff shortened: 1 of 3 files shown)


s-ludwig:
Note that the wrong timeout handling in the libasync driver didn't seem to be fixable without making changes in the libasync API, which is currently out of my scope.

Geod24:
Thanks! Do you have plans for a release any time soon ?

s-ludwig:
In fact I wanted to have one out weeks ago, but always get distracted by internal stuff. I would try to make a quick pass over the list of open PRs to see if anything is easy to merge and can then tag an RC.

denizzzka:
>  I would try to make a quick pass over the list of open PRs to see if anything is easy to merge and can then tag an RC.

#2346

AndrejMitrovic:
@s-ludwig isn't the timeout supposed to be used with the call to `resolveHost` too? Here:

https://github.com/bpfkorea/vibe.d/blob/35079d4cf348e36785783480535ff186f2e4ef02/http/vibe/http/client.d#L703-L706

`resolveHost` itself doesn't take any timeout parameters - and it calls `asyncAwaitAny` without a timeout (https://github.com/vibe-d/vibe-core/blob/b417214e325a7fb684817b4ae7b218eec46c11c1/source/vibe/core/net.d#L69).

We're having an issue right now where a call to `resolveHost` blocks. I think maybe `resolveHost` should also get a `timeout` parameter and pass it on to `asyncAwaitAny`. 

What do you think?

s-ludwig:
@AndrejMitrovic Fully agreed. Cancellation is supported by the eventcore API, too, so there is really no reason not do do that.

PR #2351, opened 2019-08-23
URL: pure ctors added

### LINKED ISSUE #2322: vibe.inet.url.URL constructor not pure
Can it be made pure?

denizzzka:
Fix #2322


denizzzka:
`nothrow` adding stucks into that fact what canFind is can throw

PR #2365, opened 2019-09-09
[WIP] Mark vibe.crypto.passwordhash as deprecated

kubo39:
~~Remove deprecated passwordhash module, it's insecure.~~
vibe.crypto.passwordhash is not secure, so mark this as deprecated.
And all functions in this module were already marked as deprecated, so just remove.


Geod24:
Hum, why was the module not deprecated ? I agree with removing the functions, but it could break packages with lingering imports.

kubo39:
> why was the module not deprecated?

I'm not sure, I saw [this issue](https://github.com/vibe-d/vibe.d/issues/794) but didn't mention for it.

Would it be better to mark the passwordhash module as deprecated and remove all functions in this module?

Geod24:
> Would it be better to mark the passwordhash module as deprecated and remove all functions in this module?

Yes

crypto/vibe/crypto/passwordhash.d
```diff
@@ -5,89 +5,5 @@
 	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
 	Authors: Sönke Ludwig
 */
+deprecated("This module will be removed(see #794). Phobos's std.digest.hmac or the DUB packages dauth or scrypt may be suitable alternatives.")
```
kubo39: I just added std.digest.hmac module. Is there any suggestion for?

kubo39: I was wrong, hmac is not suitable for.

PR #2368, opened 2019-09-14
Contributing an example on using forms with vibe-d,

pdenapo:
This is an adaptation of example in Python from

https://github.com/mdn/learning-area/blob/master/html/forms/sending-form-data/python-example.py

See also

https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Sending_and_retrieving_form_data

for a full explanation.

### DIFF
--- a/examples/form/dub.json
+++ b/examples/form/dub.json
@@ -1,0 +1,12 @@
+{
+	"name": "form_example",
+	"authors": [
+		"Pablo De Napoli"
+	],
+	"dependencies": {
+		"vibe-d": "~>0.8.2"
+	},
+	"description": "An example of how to use forms with vibe",
+	"copyright": "Copyright © 2019, Pablo De Napoli",
+	"license": "MIT"
+}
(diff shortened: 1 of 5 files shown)


examples/form/dub.json
```diff
@@ -0,0 +1,12 @@
+{
+	"name": "form_example",
+	"authors": [
+		"Pablo De Napoli"
+	],
+	"dependencies": {
+		"vibe-d": "~>0.8.2"
```
s-ludwig:
```suggestion
		"vibe-d:web": {"path": "../../"}
```

This line caused some CI builds to get stuck for some reason. It makes sense to to restrict the import to `vibe.web` here (would require to adjust the `import vibe.vibe;`, too), but it should in any case reference vibe.d by path, so that it picks up the version that is actually checked out.

From a pure example point-of-view this is a bit unfortunate, but once I get around to including the examples on the website, I'd make sure to add a simple replacement rule to show version based dependencies there.

PR #2371, opened 2019-09-24
Link to RestException from RestInterfaceClient

Panke:


web/vibe/web/rest.d
```diff
@@ -586,6 +586,10 @@ unittest {
 	the matching method style for this. The RestInterfaceClient class will derive from the
 	interface that is passed as a template argument. It can be used as a drop-in replacement
 	of the real implementation of the API this way.
+
+	Non-success:
+		If an request failed or the server returned an non-success status code,
+		an `web.common.RestException` will be thrown.
```
Geod24:
```suggestion
		a `vibe.web.common.RestException` will be thrown.
```

web/vibe/web/rest.d
```diff
@@ -586,6 +586,10 @@ unittest {
 	the matching method style for this. The RestInterfaceClient class will derive from the
 	interface that is passed as a template argument. It can be used as a drop-in replacement
 	of the real implementation of the API this way.
+
+	Non-success:
+		If an request failed or the server returned an non-success status code,
```
Geod24:
```suggestion
		If a request failed or the server returned an non-success status code,
```

You could also mention timeout here.

Panke:
Thanks for the review, will come back to this in the evening.

Panke:
Can we re-run the travis-ci check? Seems to be a fluke, since the error is "could not download diet-ng".

s-ludwig:
Hm, DUB failures have become more frequent in the last months, I think this is related to the HTTP timeout that was added at some point - needs some investigation. I've restarted the check now.

Panke:
Much appreciated

PR #2372, opened 2019-09-27
Allow cleanup of redis sockets to prevent event core leaking.

schveiguy:
See issue #2245. This allows user code to clean up a `RedisClient` or `RedisSessionStore` before exiting the program to prevent messages of leaked eventcore drivers. Not attached to the function name, so if that needs to change, let me know.

### DIFF
--- a/redis/vibe/db/redis/redis.d
+++ b/redis/vibe/db/redis/redis.d
@@ -130,6 +130,16 @@ final class RedisClient {
 		});
 	}
 
+	/** Release all connections that are not in use. Call before exiting the
+	 * program to avoid relying on the GC to clean up the sockets.
+	 */
+	void releaseUnusedConnections() @safe
+	{
+		m_connections.removeUnused((conn) @safe {
+                 conn.m_conn.close();
+		});
+	}
+
 	/// Returns Redis version
 	@property string redisVersion()
 	{
--- a/redis/vibe/db/redis/redis.d
+++ b/redis/vibe/db/redis/redis.d
@@ -135,8 +135,12 @@ final class RedisClient {
 	 */
 	void releaseUnusedConnections() @safe
 	{
-		m_connections.removeUnused((conn) @safe {
+		// the try/catch in this is for the old vibe.d:core library, not
+		// necessary in vibe-core, but we have to work with both.
+		m_connections.removeUnused((conn) @safe nothrow {
+			try {
                  conn.m_conn.close();
+			 } catch(Exception e) {}
 		});
 	}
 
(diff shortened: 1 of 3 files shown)


Geod24:
Any way this could be tested ?

schveiguy:
I'm trying to think of how to test this without having an actual Redis server...

Indeed, there's only 2 unittests in this file, and they both do not work with the actual Redis server.

I'm open to ideas, but I don't know if it can be done.

s-ludwig:
> I'm trying to think of how to test this without having an actual Redis server...
>
> Indeed, there's only 2 unittests in this file, and they both do not work with the actual Redis server.

There is a Redis integration test in the "tests" directory where this could be added. The only way to verify this that I can see is to manually count all open file descriptors/SOCKET handles, before and after. I'd not make this mandatory, but it would be nice to add a simple call in the existing test, so that ideally the current warnings go away ([CI output](https://travis-ci.org/vibe-d/vibe.d/jobs/591144252#L1806-L1842)).

redis/vibe/db/redis/redis.d
```diff
@@ -130,6 +130,20 @@ final class RedisClient {
...
+	 * program to avoid relying on the GC to clean up the sockets.
+	 */
+	void releaseUnusedConnections() @safe
+	{
+		// the try/catch in this is for the old vibe.d:core library, not
+		// necessary in vibe-core, but we have to work with both.
+		m_connections.removeUnused((conn) @safe nothrow {
+			try {
+                 conn.m_conn.close();
+			 } catch(Exception e) {}
```
s-ludwig:
```suggestion
			 } catch(Exception e) logDebug("Failed to close unused Redis connection: %s", e.msg);
```

Not sure whether this can actually occur in the implementation, but at least that way there is a chance to notice unexpected behavior.

schveiguy: Admittedly, this was simply to shut up the compiler. It only requires this for the `vibe-d:core` version, as the interface does not require `nothrow` on the `close` method.

I will update to log the exception as requested.

schveiguy:
Updated

schveiguy:
Ugh, I keep screwing up my redis test. I couldn't run the test locally because I don't have a redis server that I want to muck with. In any case, I had to force push again. If it turns green, I'll ping you to reapply the auto merge.

schveiguy:
OK, ping @s-ludwig. The failure seems unrelated to the patch.

schveiguy:
ping @s-ludwig again.

s-ludwig:
Sorry, missed the first time! Restarted and toggling auto-merge.

PR #2374, opened 2019-09-29
Add support for user added headers, like authentication tokens.

CryptoAhash:
The user can now create a websocket client with custom added req.headers.

### DIFF
--- a/http/vibe/http/websockets.d
+++ b/http/vibe/http/websockets.d
@@ -83,7 +83,7 @@ class WebSocketException: Exception
 /**
 	Returns a WebSocket client object that is connected to the specified host.
 */
-WebSocket connectWebSocket(URL url, const(HTTPClientSettings) settings = defaultSettings)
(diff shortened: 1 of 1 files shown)


CryptoAhash:
It should not break, I did tests locally. I guess it has to do with the spaces..

> This is a breaking change: You can't add a parameter in the middle like this.
Why not?

> Also it only affects one of the two `connectWebSocket` overloads.
Isn't that what overloads are for?

Geod24:
> Why not?

Because your tests did not include a custom `HTTPClientSettings`. If someone calls it with `connectWebSocket(myURL, myHTTPClientSettings)`, it will break.

> Isn't that what overloads are for?

The overload I mentioned [does not call this overload](https://github.com/vibe-d/vibe.d/blob/d4e60c4b8c23259e8c8b85ca5dec2d9da8659648/http/vibe/http/websockets.d#L112). They should probably be unified though, as it looks like code duplication.

PR #2376, opened 2019-10-01
Nullable: use .get explicitly

Panke:


Panke:
> Fetching botan-math 1.0.3 (getting selected version)...
>Fetching eventcore 0.8.46 (getting selected version)...
> ZipException: no end record

I don't think the failures are connected to this PR.

PR #2379, opened 2019-10-08
Add libssl search path for the latest Homebrew/macOS versions.

s-ludwig:


Geod24:
Why isn't `libs` used here ? With the detection based on `pkg-config`, it would *just work*.

PR #2381, opened 2019-10-18
Vibed broken OpenSSL dependency

rracariu:
See https://github.com/vibe-d/vibe.d/issues/2284

This allows to use Vibe.d out of the box with another project that uses `OpenSSL 1.1.0h`


rracariu:
Couldn't make it work...

PR #2384, opened 2019-10-27
Minor adjustment

Beyarz:
The type attribute is only useful when it's not javascript

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -382,7 +382,7 @@ struct DefaultDietFilters {
 		string indent_string = "\n";
 		while (indent-- > 0) indent_string ~= '\t';
 
-		string ret = indent_string~"<script type=\"application/javascript\">";
+		string ret = indent_string~"<script>";
 		ret ~= indent_string~'\t' ~ "//<![CDATA[";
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'm not entirely sure how to go about this. It *is* technically required for HTML 3.x/4.x, but I have no idea how many people still declare their HTML documents like that these days (and additionally using this filter). Unfortunately it is not possible to know which version is used from within the filter.

Beyarz:
> I'm not entirely sure how to go about this. It _is_ technically required for HTML 3.x/4.x, but I have no idea how many people still declare their HTML documents like that these days (and additionally using this filter). Unfortunately it is not possible to know which version is used from within the filter.

Should we just leave it as it is?

Beyarz:
While I am at it, the type attribute is optional too

Beyarz:
Of course if you want to support the older browsers then I can close my PR.

s-ludwig:
Okay, I just tested it and indeed it doesn't matter in practice whether the attribute is there or not for older HTML doctypes on any of the current browser engines. So let's just go ahead with this.

Beyarz:
Sweet!

PR #2385, opened 2019-10-30
Allow tuning of retry times

NotSpooky:
Added them as HTTPClientSettings fields.
Also adjusted ConnInfo so that the pool selection algorithm don't use a connection with different settings.


NotSpooky:
It seems that this doesn't actually solve anything.
If it's useful please reopen.

PR #2399, opened 2019-12-10
fix: use current UUID type

tirithen:
std.uuid.UUID as being converted to the legacy BSON UUID type.

BREAKING CHANGE: type change from BinData 0x03 to 0x04 for std.uuid.UUID

### DIFF
--- a/data/vibe/data/bson.d
+++ b/data/vibe/data/bson.d
@@ -798,7 +798,7 @@ struct BsonBinData {
 		generic = 0x00,
 		function_ = 0x01,
 		binaryOld = 0x02,
-		uuid = 0x03,
+		uuid = 0x04,
 		md5 = 0x05,
 		userDefined = 0x80,
 


s-ludwig:
We'll have to go through a deprecation phase here to avoid the immediate breakage. I'd suggest to add two new definitions (`uuidOld = 0x03` and `uuidNew = 0x04`) and then mark `uuid` as deprecated, but keep it as 0x03. Then, `Bson.get` would need to be adjusted to allow both, 0x03 and 0x04, and the remaining uses would be replaced by `uuidNew`.

Finally, in the next major release, `uuid` would then be changed to alias to `uuidNew` instead of `uuidOld`.

tirithen:
@s-ludwig I see, make sense deprecating first, I'll stay on my fork for now. It's such a nice design in that it can take the base structs and automatically map the members from and to BSON without annotations. Thanks for a nice library.

thaven:
> feat(SysTime): add support to parse SysTime    17f897b
> feat: serialize SysTime   b41b7ef
> Merge branch 'master' of github.com:tirithen/vibe.d  4adcc91

@tirithen I suppose you did not mean to push these commits to this PR?

tirithen:
No that was not intended, I'm experimenting on my fork for needs I have in my project. I'll see if I can clean it up.

6 feb. 2020 06:42:24 thaven <notifications@github.com>:

PR #2407, opened 2020-01-27
Redis pubsub fixes

yazd:
This fixes two issues in redis pubsub implementation.

* The first issue (solved in 1st commit) is an infinite loop (that causes 100% CPU) that happens when the redis *server* exits while listening to pubsub messages.
The following code shows a reproduction case:
```
import vibe.db.redis.redis;
import vibe.core.core;
import vibe.core.log;

void main() {
	runTask(&listenRedis);
	runApplication();
}

void listenRedis() {
	auto redis = connectRedis("127.0.0.1");
	auto subs = redis.createSubscriber();
	subs.subscribe("test");

	logInfo("subscribed");

	auto task = subs.listen((string channel, string message) {
		logInfo("received message on %s: %s", channel, message);
	});
	task.join();

	logInfo("done listening");
}
```
Redis server must be stopped to face the issue.

* Second issue (solved in 2nd commit) is that the task doesn't join back or terminate because the code in `teardown` was waiting for `Action.STOP` message, but that will never arrive, because it was already consumed by the `handler`. I don't think there are cases when that is not the case.

### DIFF
--- a/tests/vibe.db.redis.redis.pr2407/source/app.d
+++ b/tests/vibe.db.redis.redis.pr2407/source/app.d
@@ -1,29 +1,36 @@
 import vibe.db.redis.redis;
 import vibe.core.core;
 import vibe.core.log;
-
-enum redisPort = 16379;
+import vibe.core.core;
+import vibe.core.concurrency;
 
 void main()
 {
-	runTask(&redisServerMock);
-	runTask(&listenRedis);
+	auto task = runTask(&listenRedis);
+	runTask(&redisServerMock, task);
 	runApplication();
 }
 
-void redisServerMock()
+void redisServerMock(Task task)
 {
-	listenTCP(redisPort, (conn) {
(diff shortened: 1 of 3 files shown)


tests/vibe.db.redis.redis.pr2407/source/app.d
```diff
@@ -0,0 +1,40 @@
...
+void main()
+{
+	runTask(&redisServerMock);
+	runTask(&listenRedis);
+	runApplication();
+}
+
+void redisServerMock()
+{
+	listenTCP(redisPort, (conn) {
```
s-ludwig: I'd always recommend to use port 0 and use `.bindAddress.port` of the returned `TCPListener`, since it's possible that multiple CI jobs run in parallel on the same machine.

yazd: I updated the test to use `.bindAddress.port`.
If you'd like me to squash the commits, just let me know.

s-ludwig: Looks good as-is!

PR #2412, opened 2020-01-29
Do not use deprecated function in http client test 1426

Geod24:
Also do not use VibeDefaultMain and simply use a main instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ services:
 addons:
   apt:
     packages:
+      - gdb
       - pkg-config
       - zlib1g-dev
       - libevent-dev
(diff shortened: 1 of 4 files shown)


Geod24:
Note: The reason I'm looking into this test is because it always segv on Travis with a recent compiler.
On my machine it passes but triggers a wild "socket error 54!".
```console
To force a rebuild of up-to-date targets, run again with --force.
Running ./tests
sock error 54!
Warning (thread: main): leaking eventcore driver because there are still active handles
   FD 9 (streamListen)
Warning (thread: main): leaking eventcore driver because there are still active handles
```

Which originates from eventcore, either from [send](https://github.com/vibe-d/eventcore/blob/4813cba33823bd5b96e980e1ecb337d8cf3d7e15/source/eventcore/drivers/posix/sockets.d#L812) or [read](https://github.com/vibe-d/eventcore/blob/4813cba33823bd5b96e980e1ecb337d8cf3d7e15/source/eventcore/drivers/posix/sockets.d#L381).

yazd:
On my linux machine, it prints "sock error 104".
I solved it by making the server read the request (```c.readUntil("\r\n\r\n".representation);```) before writing the response.
Also, not to leak fds, you can stop the listener before calling `exitEventLoop`.

Anyway, I wasn't able to reproduce the segfault on my machine as well.

Geod24:
Turns out it was, once again, the parallel GC which was to blame...

Geod24:
> On my linux machine, it prints "sock error 104".

On Travis too. I have OSX which is probably the cause of the difference.
@wilzbach : Could you re-review ? I added a bit of gdb magic so that SEGV are shown even if they aren't in unittest.

PR #2413, opened 2020-01-30
Remove some deprecated items scheduled in 0.9

kubo39:
Could we remove more deprecations?


Geod24:
Depends on @s-ludwig release plans

s-ludwig:
It probably makes sense to remove all deprecated functions - I used to do that every other release, but 0.8.x->0.9.0 would be a natural time. I'd also mark all symbols that are currently documents as "scheduled for deprecation" as deprecated.

PR #2425, opened 2020-03-24
meson: remove deleted win32.d from vite_utils_src

Cogitri:
Required after 4f9bbc7f8125ad48dab4f6c74ccd08e61d9aaee6

### DIFF
--- a/utils/meson.build
+++ b/utils/meson.build
@@ -13,7 +13,6 @@ vibe_utils_src = [
     'vibe/internal/meta/uda.d',
     'vibe/internal/rangeutil.d',
     'vibe/internal/utilallocator.d',
-    'vibe/internal/win32.d',
     'vibe/utils/array.d',
     'vibe/utils/dictionarylist.d',
     'vibe/utils/hashmap.d',


Cogitri:
It seems like the meson build system is in very bad shape right now anyway. Are you interested in keeping it? If so, I'd love to bring it back into shape and add CI for it. Then we could introduce vibe-d to the Alpine Linux repos.

s-ludwig:
The problem is that there was constant breakage due to behavior changes in Meson, so, not using or being familiar with it, I eventually had to remove it from the CI checks. I wouldn't mind adding it back, it's just that someone else needs to make sure it stays working.

Cogitri:
> The problem is that there was constant breakage due to behavior changes in Meson, so, not using or being familiar with it, I eventually had to remove it from the CI checks. I wouldn't mind adding it back, it's just that someone else needs to make sure it stays working.

Ah yes, meson had some problems with D up until now. With the 0.54 release that should be released soon this is hopefully a thing of the past. I'll get meson support for Vibe-D back to speed so we can compile it in Alpine Linux.

Geod24:
Blocked by https://github.com/rejectedsoftware/diet-ng/issues/81

Geod24:
@Cogitri : Can you amend & force push ? CI is stuck.

Cogitri:
I think I'll close this instead and open a new PR soon which fixes all of meson for Vibe since this commit alone doesn't do much. Thanks for the heads up though :)

PR #2427, opened 2020-03-26
add some nothrow to json

benjones:
I suspect you'll want some tweaks to the formatting, so just LMK what you want to see.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -123,7 +123,7 @@ struct Json {
 	static assert(!hasElaborateDestructor!BigInt && !hasElaborateCopyConstructor!BigInt,
 		"struct Json is missing required ~this and/or this(this) members for BigInt.");
 
-	private nothrow{
+	private {
 		// putting all fields in a union results in many false pointers leading to
 		// memory leaks and, worse, std.algorithm.swap triggering an assertion
 		// because of internal pointers. This crude workaround seems to fix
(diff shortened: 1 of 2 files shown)


data/vibe/data/json.d
```diff
@@ -181,19 +181,20 @@ struct Json {
...
 	/// New JSON value of Type.Array
-	static @property Json emptyArray() { return Json(cast(Json[])null); }
+	static @property Json emptyArray() nothrow { return Json(cast(Json[])null); }
 
 	version(JsonLineNumbers) int line;
 
 	/**
 		Constructor for a JSON object.
 	*/
+    nothrow {
```
s-ludwig: I'd repeat this for every constructor for now to be in line with the rest. In the current form I think it may also break the generated documentation.

data/vibe/data/serialization.d
```diff
@@ -2028,6 +2028,7 @@ unittest {
 
 unittest { // issue 1991 - @system property getters/setters does not compile
 	static class A {
+        @safe:
```
s-ludwig: Spaces vs. tabs

benjones: I think I fixed this, I copy-pasted a tab character from the next line because I'm bad at emacs :)  Worth you double checking though

s-ludwig:
Thanks!

PR #2439, opened 2020-05-01
Fix #2438 - Error on registerWebInterface with a parameterized constructor class

### LINKED ISSUE #2438: Error on registerWebInterface with a parameterized constructor class
When I register a class with a parameterized constructor. I get the following error:

```
Error: static assert:  "Instances may only be returned from parameter-less functions (__ctor)!"
```
This happens in all  projects that were successfully built before.
For example:

```
$ git clone git@github.com:rejectedsoftware/diskuto.git
$ cd diskuto/examples/embed-raw
$ dub build

/home/o3o/.dub/packages/vibe-d-0.8.6/vibe-d/web/vibe/web/web.d(200,6): Error: static assert:  "Instances may only be returned from parameter-less functions (__ctor)!"
../../source/diskuto/web.d(33,29):        instantiated from here: registerWebInterface!(DiskutoWebInterface, cast(MethodStyle)5)
dmd failed with exit code 1.
```

My environment
```
$ dmd --version
DMD64 D Compiler v2.091.0

$ dub --version
DUB version 1.16.0-beta.1, built on Jul 22 2019
```
Thank you

Geod24:


Geod24:
CC @s-ludwig you might want to consider a new release.

web/vibe/web/web.d
```diff
@@ -183,7 +183,16 @@ URLRouter registerWebInterface(C : Object, MethodStyle method_style = MethodStyl
...
+
+		// Ignore special members, such as ctor, dtors, postblit, and opAssign
+		// See https://github.com/vibe-d/vibe.d/issues/2438
+		static if (M == "__ctor" || M == "__dtor" || M == "__xdtor" ||
+				   M == "__postblit" || M == "__xdtor" || M == "__xpostblit" ||
+				   M == "opAssign")
+		{ /* Ignore */ }
+
+		// exclude Object's default methods and field
+		else static if (!is(typeof(__traits(getMember, Object, M)))) {
```
schveiguy, line 195: This seems backwards. What about combining both static ifs?

```suggestion

		// Ignore special members, such as ctor, dtors, postblit, and opAssign
		// See https://github.com/vibe-d/vibe.d/issues/2438
		// exclude Object's default methods and field
		static if (M != "__ctor" && M != "__dtor" && M != "__xdtor" &&
				   M != "__postblit" && M != "__xdtor" && M != "__xpostblit" &&
				   M != "opAssign" &&
				   !is(typeof(__traits(getMember, Object, M)))) {
```

Geod24: IMO it only seems backward if you don't see it as a `continue` ;)
Personally I prefer to "get things out of the way", using `continue`, early return, etc, and this is why I originally wrote it this way. But I don't mind either way, so suggestion applied.

Geod24:
Ping @s-ludwig , this is blocking quite a few project (including the dlang tour)

Geod24:
Ping @s-ludwig

Geod24:
Ping @s-ludwig

o3o:
Ping @s-ludwig

web/vibe/web/web.d
```diff
@@ -183,7 +183,14 @@ URLRouter registerWebInterface(C : Object, MethodStyle method_style = MethodStyl
 		/*static if (isInstanceOf!(SessionVar, __traits(getMember, instance, M))) {
 			__traits(getMember, instance, M).m_getContext = toDelegate({ return s_requestContext; });
 		}*/
-		static if (!is(typeof(__traits(getMember, Object, M)))) { // exclude Object's default methods and field
+
+		// Ignore special members, such as ctor, dtors, postblit, and opAssign,
+		// and object default methods and fields.
+		// See https://github.com/vibe-d/vibe.d/issues/2438
+		static if (M != "__ctor" && M != "__dtor" && M != "__xdtor" &&
+				   M != "__postblit" && M != "__xdtor" && M != "__xpostblit" &&
```
drug007, line 191: "__xdtor" is repeated twice. I guess there should be "__xctor"? 

Offtopic: does special members list exist somewhere out?

schveiguy: > Offtopic: does special members list exist somewhere out?

good idea, druntime should define `isCompilerGeneratedMember(string)`, which would make this much more pleasant.

Geod24: Oooooor the language could be sane and not have magic names. Just a thought.

schveiguy: Hm... another thought here, could we just do:

```D
static if (M.length >= 2 && M[0 .. 2] == "__")
```

Since double underscores are reserved for the language?

UplinkCoder: does this come from an `__traits(allMembers)` ? If so it's already filtered.

schveiguy: > does this come from an `__traits(allMembers)` ?

Yes.

> If so it's already filtered.

No, it's not.

Geod24: There's no such thing as `__xctor`, it was just a copy-paste error.
Reorganized things now and removed the duplication.

@schveiguy : We *could* in a future improvement, but that might affect code that shouldn't be affected, so I'd like to keep this a regression fix.

drug007: > There's no such thing as __xctor, it was just a copy-paste error.

That's why I asked if there is the special members list. Thanks!

schveiguy: I mean technically postblit isn’t ever going to show up for a class either. But I think it’s reasonable to include all magic functions in the list. It’s just that using double underscore checking would be future proof and a decent compromise.

s-ludwig:
It's just unfortunate that this apparently triggers an Optlink crash. I couldn't reproduce it locally, so finding a workaround may be a bit involved.

If this weren't the default compilation mode for DMD, I'd just completely remove support for it, instead, but I'm reluctant to raise the bar for beginners. Not yet sure how to go about it.

wilzbach:
Dub defaults to mscoff since about a year, so I would recommend to just drop support for Optlink:

https://dlang.org/changelog/2.086.0.html#dmd-mscoff-default

s-ludwig:
Oh, right, I totally forgot about that. Okay, I'll merge then and make the Optlink builds "allowed_failures".

PR #2443, opened 2020-05-20
Replace rdmd dependency with single dub package

andre2007:
`import std;` is not used by purpose to not cause issues for older D versions.

### DIFF
--- a/tls/dub.sdl
+++ b/tls/dub.sdl
@@ -26,44 +26,61 @@ configuration "openssl" {
 	sourceFiles "../lib/win-i386/libssl.lib" "../lib/win-i386/libcrypto.lib" platform="windows-x86-dmd"
 
 	// Posix
-	preGenerateCommands `rdmd --eval='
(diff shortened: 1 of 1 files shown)


tls/dub.sdl
```diff
@@ -26,44 +26,60 @@ configuration "openssl" {
...
+	import std.process;
+	import std.stdio;
+	import std.string;
+	import std.range;
+	import std.uni;
+	import std.file;
+	import std.conv;
+	import std.algorithm;
+	import std.path;
+    import std.functional;
```
Geod24, line 42: Indent is off. Also can you order those ?

Geod24:
Better seen with whitespace diff off. LGTM minus a style nit.

tls/dub.sdl
andre2007: Fixed

andre2007:
Is there currently an issue with AppVeyor? It seems it crashes on linking a test on windows x86 environments.

PR #2445, opened 2020-05-24
Remove all Optlink/OMF based CI configurations.

s-ludwig:
These cause the a crash in Optlink and make the build time out.


Geod24:
Thanks! Two questions:
- Do you have an ETA for a release (because it's blocking a few things right now) ?
- Should the fact that optlink is not supported anymore be documented ?

PR #2458, opened 2020-07-10
Fix buildkite/dmd for https://github.com/dlang/dmd/pull/11320

### LINKED ISSUE #2455: Heads up: dmd fix for alias parameter missing qualifier breaks unittest
Hi,
My pull for dmd:
https://github.com/dlang/dmd/pull/11320

breaks a unittest in `utils/vibe/internal/meta/codegen.d`:
https://github.com/vibe-d/vibe.d/blob/master/utils/vibe/internal/meta/codegen.d#L95

My dmd pull needs that line:

> static assert (getSymbols!Type.stringof == TypeTuple!(A, B).stringof);

Changed to:

> static assert (getSymbols!Type.stringof == TypeTuple!(A, const(B), const(A)).stringof);

Does that look correct?

ntrel:
This fixes #2455.


Geod24:
A comment would be nice

ntrel:
@geod24 I've edited the description and added links, hope that's ok

utils/vibe/internal/meta/codegen.d
```diff
@@ -78,7 +78,7 @@ template getSymbols(T)
 			alias Implementation = TypeTuple!();
 	}
 
-	alias getSymbols = NoDuplicates!(Implementation!T);
+	alias getSymbols = NoDuplicates!(staticMap!(Unqual, Implementation!T));
```
UplinkCoder, line 81: This has compile-time performance implications.

schveiguy: Isn't this a breaking change? Wouldn't the original say `AliasSeq!(int, const(int))` for `int[const(int)]`? alias parameters seem to allow `const(int)` to go through, whereas `Unqual!(const(int))` turns into `int`

ntrel: @schveiguy Oops you're right

ntrel: Actually I think basic types are ignored anyway

schveiguy: Yep, you are right, I didn't realize that.

schveiguy:
looking at the usage (figure out which imports need to happen to include these symbols), I think this change is correct in terms of what this thing is meant to do. I would suggest a document comment that clarifies that attributes will be stripped from types.

utils/vibe/internal/meta/codegen.d
schveiguy: Instead of this can you apply Unqual at the leaf instantiations?  That would prevent the need for static map.

Issue #1043, opened 2015-03-28
[REST](14369) Default parameters won't compile if not CTFE-able


Geod24:
https://issues.dlang.org/show_bug.cgi?id=14369

Saddly this can't be fixed in Vibe.d, we have to wait on a DMD / Phobos bugfix.

s-ludwig:
Reminds me of the annoying related bug that accessing a default initialized `SysTime` causes an access violation.

s-ludwig:
-> https://github.com/D-Programming-Language/phobos/pull/2088

PR #2464, opened 2020-08-12
docs: function parameter documentation including `id`

jblachly:
This addresses https://github.com/vibe-d/vibe.d/issues/2118

Documentation for vibe.web.rest was enhanced by:
* Being explicit about how function parameters are populated, and that in special cases they may affect the route path
* Reorder GET/PUT before POST/PATCH
* **Explicitly call out `id` as a magic function/method parameter name** that affects the route (not documented elsewhere)
* Describe the DMD bug #14369 (Vibe.d: #1043) as a CTFE problem

### DIFF
--- a/web/vibe/web/rest.d
+++ b/web/vibe/web/rest.d
@@ -207,11 +207,15 @@
 		}
 		---
 
-	Parameter_passing:
-		By default, parameter are passed via different methods depending on the
-		type of request. For POST and PATCH requests, they are passed via the
-		body as a JSON object, while for GET and PUT they are passed via the
-		query string.
+	Parameters:
+		Function parameters may be populated from the route, query string,
+		request body, or headers. They may optionally affect the route URL itself.
+
+		By default, parameters are passed differently depending on the type of
+		request (i.e., HTTP method). For GET and PUT, parameters are passed
+		via the query string (`<route>?paramA=valueA[?paramB=...]`),
+		while for POST and PATCH, they are passed via the request body
+		as a JSON object.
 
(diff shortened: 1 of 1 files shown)


web/vibe/web/rest.d
```diff
@@ -186,14 +190,34 @@
...
+        
+		Further, how function parameters are named may affect the route:
+
+		$(UL
+			$(LI If the first parameter is named `id`, this is interpreted as a
+				leading route component. For example, `getName(int id)` becomes
+				`/:id/name`.)
+				$(B Note that this style of parameter-based URL routing is
+				different than in many other web frameworks, ) where instead
+				ths example would be routed as `/name:id`.
```
Geod24, line 202: Shouldn't this be `/name/:id` ?

Geod24 (commented):
Thanks for improving the documentation, and sorry for taking so long to review this!
As you noted, "Manual endpoint specification" is a more generic version of "id", and AFAIR the plan was to actually deprecate `id` in favor of `@path`. @s-ludwig ?

web/vibe/web/rest.d
jblachly: Yes, and I also misspelled "this" in that same line. Typing too quickly. Thank you for catching, will update.

Geod24 (approved):

s-ludwig:
> AFAIR the plan was to actually deprecate id in favor of @path. @s-ludwig ?

True, either `_id` should be used in conjunction with `@path`, or, usually the better choice from the D API point-of-view, `Collection!I`.

PR #2465, opened 2020-08-18
Reduce supported versions to the last 10 releases

Geod24:
```
Vibe.d officially supports only the last 5 releases,
but in practice support extends much further than that.
Re-align it to 10 releases to avoid too much hassle
on developers.

Also remove the non-maintained README.md support list,
and replace it with a policy.
```


README.md
```diff
@@ -48,42 +48,18 @@ Alternatively, you can quickstart with examples directly on [![Open on run.dlang
...
+then v2.089.x, v2.088.x, ... v2.080.x are supported.
+Note that support for patch release is desireable,
+but only support for the last patch in a minor is guaranteed.
+
+Additionally, Vibe.d supports all LDC versions that implement
+the version of a supported frontend (e.g. by the previous rule
+[LDC v1.20.0](https://github.com/ldc-developers/ldc/releases/tag/v1.20.0)
+implements v2.090.1 and would be supported).
+
+Finally, Vibe.d aims to support the latest version of GDC.
```
wilzbach, line 62: This is not the case.

wilzbach (changes requested):

README.md
Geod24: Well, I remember support for an older frontend staying just to ensure GDC compilation.
Also I've phrased this differently: The rest are "does" while this is "aims", as in, it's desire-able as long as it doesn't cause unreasonable burden.

wilzbach: Yes, but at some point the workarounds to support 2.068 got out of hands and GDC support was dropped.

I just fear that people understand this as GDC is supported, which it is not.

Geod24: Alright I'll just remove it.

Geod24: Removed

Geod24:
Ping @wilzbach

PR #2473, opened 2020-09-01
Replace randomUUID by securely generated random data for digest authentication.

s-ludwig:
randomUUID does not use a secure RNG and exposes potential security issues, namely if sequences of randomly generated UUIDs (or other random data based on the same RNG) in other places of an application are exposed to an attacker, it becomes possible to predict the digest auth secret.

See https://forum.dlang.org/thread/zjmyxudvscfwrkrdabmt@forum.dlang.org

### DIFF
--- a/http/vibe/http/auth/digest_auth.d
+++ b/http/vibe/http/auth/digest_auth.d
@@ -9,8 +9,9 @@
 */
 module vibe.http.auth.digest_auth;
 
-import vibe.http.server;
 import vibe.core.log;
+import vibe.crypto.cryptorand;
+import vibe.http.server;
 import vibe.inet.url;
 
 import std.base64;
@@ -18,7 +19,6 @@ import std.datetime;
 import std.digest.md;
 import std.exception;
 import std.string;
-import std.uuid;
 
 @safe:
 
@@ -29,12 +29,12 @@ class DigestAuthInfo
 	@safe:
 
 	string realm;
-	ubyte[] secret;
+	ubyte[32] secret;
 	ulong timeout;
 
 	this()
 	{
-		secret = randomUUID().data.dup;
+		secureRNG.read(secret[]);
 		timeout = 300;
 	}
 


Geod24:
1) Why is the class `final` ?
2) Could you target `v0.9.x`, so we can make a patch release soon ?

s-ludwig:
> Why is the class `final` ?

Not sure about what to do with the `final` (independently of whether it belongs in this commit), it shouldn't have been non-final in the first place and it's very unlikely that someone has derived a class from it, but technically it's a breaking change. I'm inclined to just change it to `final` as a separate change and see if there is any complaint, but in any case, I'll remove it from this commit.

> Could you target `v0.9.x`, so we can make a patch release soon ?

What was the reason for introducing of the 0.9.x branch, do you plan to do any breaking changes on master? Would it be okay to instead do breaking changes in a separate branch and keep master as the "0..9.x" branch? I'd like to avoid adding complexity to the development process, unless it has compelling benefits.

s-ludwig:
PS: Just to avoid confusion with the numbering scheme, so far `0.x.y` means a new `x` contains breaking changes and `y` contains only non-breaking changes (new features and bug fixes alike). Taking into account the currently reduced development speed (and also the fact that this is not really a pre-release library anymore), the full `x.y.z` scheme would kind of make sense, it's just that my idea a few years back was to basically separate this whole repository into individually versioned sub libraries and use that as the opportunity to go `1.0.0+`.

Geod24:
> What was the reason for introducing of the 0.9.x branch

I wanted to quickly make small release that wouldn't break any code.

>  I'd like to avoid adding complexity to the development process

If you change the branch, I will simply take care of the release / merging, so the complication is on me :)

> PS: Just to avoid confusion with the numbering scheme, so far `0.x.y` means a new `x` contains breaking changes and `y` contains only non-breaking changes (new features and bug fixes alike).

To be a bit pedantic about this, SemVer says that under v1.x.x, anything goes. So having this set of rule is similar to re-inventing SemVer without really following it, which is quite confusing.

> Taking into account the currently reduced development speed (and also the fact that this is not really a pre-release library anymore), the full `x.y.z` scheme would kind of make sense, it's just that my idea a few years back was to basically separate this whole repository into individually versioned sub libraries and use that as the opportunity to go `1.0.0+`.

There's a few points I want to note:

- First, the separation between `eventcore`, `vibe-core`, and `vibe.d` has been a pain, since a fix to `eventcore` usually means a fix to `vibe-core` and `vibe.d` (e.g. the latest row of DNS fixes).
But conceptually, I understand the reason for the split, and tend to agree with it. However real-life experience hasn't been that pleasant.

- Second, if you follow SemVer, nothing prevents you from breaking down the library later (in a major release). You kind-of did that already with the long `vibe-core` transition.

- Last but not least, I spent my three years at Sociomantic dealing with versioning, and the scheme we used added a bit of development overhead, but was working very, very well:
  - We had two kinds of branches: major branches (`v\d+\.x\.x`, e.g. `v0.x.x`) and minor branches (`v\d+\.\d+\.x`, e.g. `v1.2.x`);
  - All branches were supported versions. We'd  delete a branch when support period ended.
  - There was no master branch. The default branch was  the *oldest  supported major*.
  - Bugfix would go to the oldest supported minor that had this bug.
  - Features would go to the oldest supported major, if possible.
  - When tagging a new feature, we would tag it on the major branch (IOW a commit on `v5.x.x` would become `v5.4.0`), and then create a minor branch out of it (in this case, `v5.4.x`).
  - When tagging a patch release, it would go into the minor branch, then get merged back into the next supported minor. E.g. a  bugfix would target `v5.3.x`, get tagged as `v5.3.1`, then the tag merged into `v5.4.x`, tagged as `v5.4.1`, then merged into `v5.x.x`).
  - When tagging a minor release, we would merge into the next major. So if `v5.4.0` was just tagged, we would merge it into `v6.x.x` if it existed.
  - We'd have a support period and schedule for each release: E.g. minors release would be supported for 2 months from the release of the next minor (we'd have a branch for it for two months), and major for 6 months after the release of the next major.

That was pretty much the set of rules we were using. There's some tiny details, I might have forgotten to write down, but essentially this gave us the ability to:
- Provide a stable experience for users;
- Provide regular releases with new features;
- Break code when breakage was needed;

Vibe.d is currently in a weird state where it's sub `v1.0.0`, so anything could go, but it's a mature library with one of the best support for versioning in the D language. So my `v0.9.x` was an attempt to provide some quick fixes to users who need it (we have tons of issues ATM), without disrupting your workflow (you can just PR to master and I'll treat it as `v0.x.x`) and allowing those to still follow the quality requirement of Vibe.d.

Geod24:
@s-ludwig : On an almost unrelated note, I sent you an email (to see if we could get `diet-ng` moved to this org so we can also move faster).

wilzbach:
We have already tried Sociomantic's versioning with D-Scanner et.al.
It wasn't a great experience as I think it only works well for full-time teams. Otherwise it's a lot of hassle.

However, I do see the advantages in being able to release bug fixes more often. Why can't we just use the easy master/stable approach? It works well and is easy for contributors and familiar for anyone who has ever contributed to DMD.

wilzbach (approved):

s-ludwig:
I still haven't had time to think through the whole picture, but what I can say is that the whole idea of a bugfix branch is messing quite a bit my current personal development mode. One example is that I quite frequently need a little feature here or there, for which I then create an `0.9.x-alpha.y` tag to make it available. Making this work with any kind of multi-branch scenario requires switching to the release versioning scheme (1.x.y), which I did not intend to do for this repository so far.

I will get back to this, but for now please let's just merge this into master, as I need to make such an alpha release.

PS: Just to mention the main thing that motivates the current 0.9.x approach and the related plan to split off parts into their own repositories. There are parts of the code base that clearly not have found their final API so far. Evolving them in a combined repository means frequent major version bumps that affect the whole library. There are basically two solutions to that - splitting the immature code off, or using a different versioning scheme, for example removing deprecated features after n releases. The latter is what vibe.d has been doing so far, similar to DMD.

Geod24:
> One example is that I quite frequently need a little feature here or there, for which I then create [...]

This works well if you're a single user, but there's quite a few Vibe.d users.
I personally had a terrible experience with `v0.9.0` - Since the OpenSSL version was broken, I had to fork Vibe.d and use that for a while in my project. That wouldn't have been *such* a problem if it didn't show up with basically every other projects using Vibe.d: when I tried to contribute to `dlang-bot`, I couldn't run the testsuite because I was hitting the same problem. I ended up modifying the package locally to just make it work, but you can see how user-unfriendly it can be.
Now that I have write access the problem will go away for me, but someone else might just have some similar experience.

> I will get back to this, but for now please let's just merge this into master, as I need to make such an alpha release.

No problem on my side, just wanted to share my experience on a scheme that worked without much downside.

> There are basically two solutions to that - splitting the immature code off, or using a different versioning scheme, for example removing deprecated features after n releases. The latter is what vibe.d has been doing so far, similar to DMD.

Right, but without a schedule for releases, that doesn't quite work. The `alpha` scheme only works if you deal with one project. I wanted the fixes I made for `-preview=in` to be picked up by all projects that depend on Vibe.d, which wouldn't work with an `alpha` release.

PR #2475, opened 2020-09-06
Remove some deprecated code

Geod24:
The `alias this` deprecation in particular is quite painful.

### DIFF
--- a/examples/http_info/views/info.dt
+++ b/examples/http_info/views/info.dt
@@ -4,19 +4,19 @@ html(lang="en")
 		title Info
 	body
 		h1 #{req.method} #{req.requestURL}
-		
+
 		h2 Headers
 		dl
-			- foreach( name, value; req.headers )
+			- foreach( name, value; req.headers.byKeyValue)
 				dt=name
 				dd=value
-				
+
 		h2 Cookies
 		dl
-			- foreach( name, value; req.cookies )
+			- foreach(name, value; req.cookies)
 				dt=name
 				dd=value
-				
+
 		h2 Session
 		- if (req.session)
 			ul
@@ -24,20 +24,20 @@ html(lang="en")
 					li=name
 		- else
 			p No session.
-			
+
 		h2 Query fields
 		p Query string: #{req.queryString}
 		dl
-			- foreach( name, value; req.query )
+			- foreach(name, value; req.query.byKeyValue)
 				dt=name
 				dd=value
-				
+
 		h2 Form Data
 		dl
-			- foreach( name, value; req.form )
+			- foreach( name, value; req.form.byKeyValue)
 				dt=name
 				dd=value
-				
+
 		form( action="/", method="POST" )
 			input(type="text", name="message")
 			input(type="submit")
--- a/examples/http_request/source/app.d
+++ b/examples/http_request/source/app.d
@@ -9,7 +9,7 @@ void main()
 		},
 		(scope res) {
 			logInfo("Response: %d", res.statusCode);
-			foreach (k, v; res.headers)
+			foreach (k, v; res.headers.byKeyValue)
 				logInfo("Header: %s: %s", k, v);
 		}
 	);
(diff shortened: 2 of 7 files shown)


wilzbach (approved):
Though this shouldn't be merged until we decide on whether there will be a stable branch (in some form) or releases will continue from master.

s-ludwig:
Needs a `.byKeyValue`:

`http/vibe/http/client.d(297,4): Error: invalid foreach aggregate res.headers, define opApply(), range primitives, or use .tupleof`

I'm not quite there yet with having a strict stable branch policy (one example being the slightly breaking change in #2487), so this is fine to merge (but it would be anyway, since this is master). We/I should prepare a wiki page with parts of the API that need to be reworked. Once the low-hanging fruits are done and vibe-http is properly migrated to the new repository, I'd be open to switch to a more formal scheme, even jumping to 1.0.0 could be an option then, deferring the rest of the breaking changes to 2.0.0.

Geod24:
Would love to see this!

Regarding this PR, I'm not so sure removing the `alias this` and expecting people to use `byKeyValue` was a good idea.
Perhaps providing range primitives would be better suited here, wouldn't it ?

s-ludwig:
The really bad thing is that there is no cue in the deprecation warning showing where in the code `alias this` is being used, so that although the deprecation has been there for a long time now, some people may not yet have fixed the issue. Maybe letting `alias this` point to a template method with a `static assert(false)` and a migration note could mitigate that.

How would you provide range primitives other than with an `alias this`, though?

s-ludwig:
I've created a wiki page and added all actionable high-level items that I could find from a quick look: https://github.com/vibe-d/vibe.d/wiki/API-Improvements -- there may be some low level issues that I missed now, but most of them are supposedly in the modules that should be moved out of this repository anyway.

Geod24:
> The really bad thing is that there is no cue in the deprecation warning showing where in the code alias this is being used, so that although the deprecation has been there for a long time now, some people may not yet have fixed the issue. Maybe letting alias this point to a template method with a static assert(false) and a migration note could mitigate that.

Funny you would mention this: https://github.com/dlang/dmd/pull/11839
It will be part of v2.095.0, but you can already use it to find the origin of those deprecation messages.

> How would you provide range primitives other than with an `alias this`, though?

```D
struct Bar { int v; }
void main ()
{
    Bar b;
    foreach (x; b) {}
}
```

```console
$ dmd foo.d
foo.d(5): Error: invalid foreach aggregate b, define opApply(), range primitives, or use .tupleof
```

In this case, I think range primitives would make most sense (`front` / `empty` / `popFront`).

s-ludwig:
Okay, but that's what `.byKeyValue` provides and the `alias this` makes it available for `DictionaryList` itself. Having it on `DIctionaryList` directly wouldn't make sense obviously.

But the problem with `alias this` and range primitives is that it restricts the iteration to a single definition (key+value in this case) and to me it feels much cleaner to always provide explicit range properties instead. `opApply` was more flexible in that regard, but the complete inability to handle attributes (e.g. `nothrow` and `@safe`) properly makes that a non-option.

Geod24:
> a single definition (key+value in this case)

Using range primitives you can actually expose only value and rely on `enumerate`, but it's extra work.

>  but the complete inability to handle attributes (e.g. `nothrow` and `@safe`) properly makes that a non-option.

You might be interested in my DConf talk then ;)

I don't think I'll be able to come back to this PR this week. I have some deadlines at work, plus the DConf talk which limit my time quite greatly ATM.

s-ludwig:
> Using range primitives you can actually expose only value and rely on enumerate, but it's extra work.

Not for a key-value-map, like in this case, though ;-)

> You might be interested in my DConf talk then ;)

Indeed!

s-ludwig:
There is still another use of `alias this` in one of the examples:
```
+cd examples/http_info
+dub build --compiler=dmd --build-mode=separate
(...)
info.dt(10,2): Error: invalid foreach aggregate req.headers, define opApply(), range primitives, or use .tupleof
info.dt(31,2): Error: invalid foreach aggregate req.query(), define opApply(), range primitives, or use .tupleof
info.dt(37,2): Error: invalid foreach aggregate req.form(), define opApply(), range primitives, or use .tupleof
```

Geod24:
Should be (finally) G2G

PR #2481, opened 2020-10-09
Fix HTTP range boundary condition handling

s-ludwig:
Previously this would happen:
 - Request:
   ```
   Range: 10-10
   ```
 - Response:
   ```
   Content-Length: 0
   Content-Range: 9-9/11
   ```

now responds with
 - Response:
   ```
   Content-Length: 1
   Content-Range: 10-10/11
   ```

Ranges that cannot be satisfied are now also reported as an error (range start greater than the last byte or range start greater than range end).

### DIFF
--- a/http/vibe/http/fileserver.d
+++ b/http/vibe/http/fileserver.d
@@ -331,29 +331,27 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse 
 		try {
 			if (s[0].length) {
 				rangeStart = s[0].to!ulong;
-				rangeEnd = s[1].length ? s[1].to!ulong : dirent.size;
+				rangeEnd = s[1].length ? s[1].to!ulong : dirent.size-1;
 			} else if (s[1].length) {
-				rangeEnd = dirent.size;
-				auto len = s[1].to!ulong;
-				if (len >= rangeEnd)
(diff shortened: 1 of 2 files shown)


s-ludwig:
@WebFreak001: Just FYI, since you wrote the original code. This was causing Safari to hang when loading an embedded video file.

l-kramer (approved):

http/vibe/http/fileserver.d
```diff
@@ -331,29 +331,27 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse
 		try {
 			if (s[0].length) {
 				rangeStart = s[0].to!ulong;
-				rangeEnd = s[1].length ? s[1].to!ulong : dirent.size;
+				rangeEnd = s[1].length ? s[1].to!ulong : dirent.size-1;
```
WebFreak001, line 334: could this overflow with dirent.size == 0?

s-ludwig: You are right, I had singed types in mind, when writing this. I actually first changed `rangeEnd` to be exclusive, which didn't have this issue, but decided against it to keep the diff smaller, but I guess that's nicer than changing everything to signed.

PR #2482, opened 2020-10-14
Refactor Range header parsing into a separate function and fix handling of zero-sized files

s-ludwig:
Changes the internal representation of a range to be end-exclusive (consistent with D slices) to avoid unsigned integer wrap-around for zero sized files.

See https://github.com/vibe-d/vibe.d/pull/2481#discussion_r504012709.

### DIFF
--- a/http/vibe/http/fileserver.d
+++ b/http/vibe/http/fileserver.d
@@ -315,46 +315,15 @@ private void sendFileImpl(scope HTTPServerRequest req, scope HTTPServerResponse 
 		res.headers["Content-Type"] = mimetype;
 
 	res.headers.addField("Accept-Ranges", "bytes");
-	ulong rangeStart = 0;
-	ulong rangeEnd = 0;
-	auto prange = "Range" in req.headers;
-
-	if (prange) {
-		auto range = (*prange).chompPrefix("bytes=");
-		if (range.canFind(','))
-			throw new HTTPStatusException(HTTPStatus.notImplemented);
-		auto s = range.split("-");
-		if (s.length != 2)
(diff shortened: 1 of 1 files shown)


http/vibe/http/fileserver.d
```diff
@@ -709,3 +678,55 @@ unittest
...
+		throw new HTTPStatusException(HTTPStatus.rangeNotSatisfiable);
+	}
+
+	return ret;
+}
+
+unittest {
+	auto res = createTestHTTPServerResponse();
+	assertThrown(parseRangeHeader("bytes=2-1", 10, res));
+	assertThrown(parseRangeHeader("bytes=10-10", 10, res));
```
WebFreak001, line 722: I believe this is a valid range (10-10 would simply mean, just the one byte at index `10`) and thus should not throw.

See 2.1 in the RFC which only says:

```
   A byte-range-spec is invalid if the last-byte-pos value is present
   and less than the first-byte-pos.
```

plus it's actually used as an example here:

```
   o  The first and last bytes only (bytes 0 and 9999):
        bytes=0-0,-1
```

WebFreak001 (changes requested):

http/vibe/http/fileserver.d
s-ludwig: Wait, but index 10 is outside of 10 bytes isn't it?

s-ludwig: Or do you mean that it should get moved into the range as `9-9`?

WebFreak001: oh sorry you are right about that, I wasn't looking at the size there

PR #2492, opened 2020-11-09
Implemented support for custom serialization

ferencdg:
Added @ResultSerializer UDA to control serialization

### DIFF
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -10,10 +10,12 @@ module vibe.web.common;
 import vibe.http.common;
 import vibe.http.server : HTTPServerRequest;
 import vibe.data.json;
-import vibe.internal.meta.uda : onlyAsUda;
+import vibe.internal.meta.uda : onlyAsUda, UDATuple;
 
+import std.meta : AliasSeq;
 static import std.utf;
 static import std.string;
+import std.traits : getUDAs, ReturnType;
 import std.typecons : Nullable;
 
 
@@ -446,10 +448,18 @@ class RestException : HTTPStatusException {
 		Json m_jsonResult;
 	}
 
-	@safe:
+    ///
+	this (int status, string result, string file = __FILE__, int line = __LINE__,
+		Throwable next = null) @safe
+	{
+		Json jsonResult = Json.emptyObject;
+		jsonResult["message"] = result;
+		this(status, jsonResult, file, line);
+	}
 
 	///
-	this(int status, Json jsonResult, string file = __FILE__, int line = __LINE__, Throwable next = null)
+	this (int status, Json jsonResult, string file = __FILE__, int line = __LINE__,
+		Throwable next = null) @safe
 	{
 		if (jsonResult.type == Json.Type.Object && jsonResult["statusMessage"].type == Json.Type.String) {
 			super(status, jsonResult["statusMessage"].get!string, file, line, next);
@@ -461,8 +471,11 @@ class RestException : HTTPStatusException {
 		m_jsonResult = jsonResult;
 	}
 
-	/// The HTTP status code
-	@property const(Json) jsonResult() const { return m_jsonResult; }
+	/// The result text reported to the client
+	@property inout(Json) jsonResult () inout nothrow pure @safe @nogc
+	{
+		return m_jsonResult;
+	}
 }
 
 /// private
@@ -478,6 +491,39 @@ package struct MethodAttribute
 	HTTPMethod data;
 	alias data this;
 }
+
+///
+public struct ResultSerializer (alias ST, alias DT, string ContentType)
+{
+	enum contentType = ContentType;
+	alias serialize = ST;
+	alias deserialize = DT;
+}
+
+alias resultSerializer(Args...) = ResultSerializer!(Args);
+
+///
+package alias DefaultSerializerT (FuncRetT) =
+	ResultSerializer!(
+		function (ref output_range, ref value) {
+			 serializeToJson(output_range, value);
+		},
+		function (input_stream) {
+			import vibe.stream.operations;
+			return deserializeJson!FuncRetT(input_stream.readAllUTF8);
+		},
+		"application/json"
+	);
+
+/// Convenience template to get all the ResultSerializers for a function
+package template ResultSerializersT (alias FuncT)
+{
+	alias DefinedSerializers = getUDAs!(FuncT, ResultSerializer);
+	static if (getUDAs!(FuncT, ResultSerializer).length)
+		alias ResultSerializersT = DefinedSerializers;
+	else
+		alias ResultSerializersT = AliasSeq!(DefaultSerializerT!(ReturnType!FuncT)());
+};
 
 /**
  * This struct contains the name of a route specified by the `path` function.
--- a/web/vibe/web/rest.d
+++ b/web/vibe/web/rest.d
@@ -148,6 +148,50 @@
 		`vibe.web.common.MethodStyle` for more information about the available
 		styles.
 
+	Serialization:
+		By default the return values of the interface methods are serialized
+		as a JSON text and sent back to the REST client. To override this, you
+		can use the @resultSerializer attribute
+
+		---
+		struct TestStruct {int i;}
+
+		interface IService {
+		@safe:
+			@resultSerializer!(
+				// output_stream implements OutputRange
+				function (output_stream, test_struct) {
+					output_stream ~= serializeToJsonString(test_struct);
+				},
+				// input_stream implements InputStream
+				function (input_stream) {
+					return deserializeJson!TestStruct(input_stream.readAllUTF8());
+				},
+				"application/json")()
+			@resultSerializer!(
+				// output_stream implements OutputRange
+				function (output_stream, test_struct) {
+					output_stream ~= test_struct.i.to!string();
+				},
+				// input_stream implements InputStream
+				function (input_stream) {
+					TestStruct test_struct;
+					test_struct.i = input_stream.readAllUTF8().to!int();
+					return test_struct;
+				},
+				"plain/text")()
+			TestStruct getTest();
+		}
+
+		class Service : IService {
+		@safe:
+			TestStruct getTest() {
+				TestStruct test_struct = {42};
+				return test_struct;
+			}
+		}
+		---
+
 	Parameter_passing:
 		By default, parameter are passed via different methods depending on the
 		type of request. For POST and PATCH requests, they are passed via the
@@ -231,7 +275,7 @@ public import vibe.web.common;
 import vibe.core.log;
 import vibe.core.stream : InputStream;
 import vibe.http.router : URLRouter;
-import vibe.http.client : HTTPClientSettings;
+import vibe.http.client : HTTPClientResponse, HTTPClientSettings;
 import vibe.http.common : HTTPMethod;
 import vibe.http.server : HTTPServerRequestDelegate, HTTPServerRequest, HTTPServerResponse;
 import vibe.http.status : HTTPStatus, isSuccessCode;
@@ -242,8 +286,12 @@ import vibe.inet.message : InetHeaderMap;
 import vibe.web.internal.rest.common : RestInterface, Route, SubInterfaceType;
 import vibe.web.auth : AuthInfo, handleAuthentication, handleAuthorization, isAuthenticated;
 
-import std.algorithm : startsWith, endsWith;
+import std.algorithm : startsWith, endsWith, sort;
+import std.algorithm.searching : count;
+import std.array : appender, split;
+import std.meta : AliasSeq;
 import std.range : isOutputRange;
+import std.string : strip, indexOf, toLower;
 import std.typecons : No, Nullable, Yes;
 import std.typetuple : anySatisfy, Filter;
 import std.traits;
@@ -730,9 +778,12 @@ class RestInterfaceClient(I) : I
 
 			auto httpsettings = m_intf.settings.httpClientSettings;
 
-			return .request(URL(m_intf.baseURL), m_requestFilter,
+			auto http_resp = .request(URL(m_intf.baseURL), m_requestFilter,
 				m_requestBodyFilter, verb, path,
 				hdrs, query, body_, reqReturnHdrs, optReturnHdrs, httpsettings);
+			scope(exit) http_resp.dropBody();
+
+			return http_resp.readJson();
 		}
 	}
 }
@@ -793,6 +844,10 @@ class RestInterfaceSettings {
 	*/
 	MethodStyle methodStyle = MethodStyle.lowerUnderscored;
 
+    /** The content type the client would like to receive the data back
+	*/
+	string content_type = "application/json";
+
 	/** Ignores a trailing underscore in method and function names.
 
 		With this setting set to $(D true), it's possible to use names in the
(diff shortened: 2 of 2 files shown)


ferencdg:
We needed this functionality to interact with non-vibe.d clients that expected the serialized return value to be in a specific format. For example Prometheus server would call into our vibe.d REST service and we need to return the result in a specific format.

s-ludwig (changes requested):
Thanks for this effort! I wanted to do this for a long time, too, but so far didn't have an important enough use case to actually do it. There are a few details that should be changed, but in general the changes look good.

web/vibe/web/common.d
```diff
@@ -479,6 +475,36 @@ package struct MethodAttribute
 	alias data this;
 }
 
+///
+public struct ResultSerializer(alias ST, alias DT, string ContentType)
```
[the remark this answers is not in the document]
ferencdg: > To conform with the general UDA convention, this should be documented as `/// private` and a function `auto resultSerializer(alias ST, alias DT, string content_type)() { return ResultSerializer!(ST, DT, content_type).init; }` should be used as the public API.

I tried this before, and ran into this compiler bug:
https://issues.dlang.org/show_bug.cgi?id=21379

web/vibe/web/common.d
```diff
@@ -479,6 +475,36 @@ package struct MethodAttribute
...
+	alias content_type = ContentType;
+	alias serialize = ST;
+	alias deserialize = DT;
+}
+
+///
+package alias DefaultSerializerT (FuncRetT) =
+	ResultSerializer!(
+		function (FuncRetT rev_val){
+			return rev_val.serializeToJsonString();
```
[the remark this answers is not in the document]
ferencdg: That is a good idea, and I will change the interface to accept a new output parameter, however I don't plan to restrict the interface to only accept OutputRanges. The client of the serializiation code might decide to use a custom output interface other than the OutputRange and the same client might also provide serialization UDAs for that specific custom output interface. Would there be any benefit restricting the type?

ferencdg:
thanks for the comments @s-ludwig, I changed most of them, please see my replies.

I also have an other serialization related pull request:
https://github.com/vibe-d/vibe.d/pull/2493

and today or tomorrow I will have a third one for 
https://github.com/vibe-d/vibe.d/issues/2415

web/vibe/web/common.d
```diff
@@ -440,16 +442,23 @@ unittest {
 
 /**
  	Respresents a Rest error response
-*/
-class RestException : HTTPStatusException {
+*/class RestException : HTTPStatusException {
```
Geod24, line 445: ?

web/vibe/web/common.d
```diff
@@ -440,16 +442,23 @@ unittest {
...
 	private {
 		Json m_jsonResult;
 	}
 
 	@safe:
 
+    ///
+	this (int status, string result, string file = __FILE__, int line = __LINE__, Throwable next = null)
```
Geod24, line 453: Any attributes that can go here, such as`@safe` and `nothrow` ?

web/vibe/web/common.d
```diff
@@ -461,10 +470,11 @@ class RestException : HTTPStatusException {
 		m_jsonResult = jsonResult;
 	}
 
-	/// The HTTP status code
-	@property const(Json) jsonResult() const { return m_jsonResult; }
+	/// The result text reported to the client
+	@property const(Json) jsonResult () const { return m_jsonResult; }
 }
 
+
```
Geod24, line 477: Extra newline ?

web/vibe/web/rest.d
```diff
@@ -1987,6 +2068,159 @@ private string generateModuleImports(I)()
...
+		req_content_types_str = list of allowed content types for example:
+		text/*;q=0.3, text/html;q=0.7, text/html;level=1
+
+	Returns:
+		index of the result serializers in the T... AliasSeq if matching found,
+		-1 otherwise
+
+***************************************************************************/
+
+package int get_matching_content_type (T...)(string req_content_types_str) pure @safe
```
Geod24, line 2114: This whole `Content-Type` handling stuff would be better suited in `vibw.inet`, wouldn't it @s-ludwig ?
E.g. https://github.com/vibe-d/vibe.d/blob/master/inet/vibe/inet/mimetypes.d

Geod24: Of course, it would mean to be made more generic, to allow matching content types with a list of "allow".

web/vibe/web/common.d
ferencdg: @safe already applied (@safe:) and other cannot go, but I will change the attributes on jsonResult in a separate commit

web/vibe/web/rest.d
s-ludwig: It would indeed fit there better, but in general I'm okay with keeping such `private`/`package` things locally if their API is not yet suitable for a public interface. Otherwise it increases the PR review complexity and time-to-merge, which usually is the biggest issue for overall progress.

But this is definitely something to keep in mind/as a follow-up ticket.

web/vibe/web/common.d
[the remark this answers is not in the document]
s-ludwig: Since the web/common.d code provides an object that needs to work for all user serializers, don't really see how to make this work without some form of common interface, and an output range is just the most simple form of that. But of course, since this would be a template parameter, the operations that the concrete type supports are not restricted, and user code could still make use of that.

web/vibe/web/rest.d
```diff
@@ -711,7 +755,7 @@ class RestInterfaceClient(I) : I
 		 * Returns:
 		 *     The Json object returned by the request
 		 */
-		Json request(HTTPMethod verb, string name,
+		string request(HTTPMethod verb, string name,
```
s-ludwig, line 758: I didn't catch that this also is a breaking change due to the `protected` atttribute. I'm also not sure why it is `protected` in the first place, though, so we might just deprecate it and add the new signature as a private method (and go for a string output range instead of a `string` return value to prepare for dropping `appender` in favor of a `StreamOutputRange` or similar).

The old deprecated method would then only be called if no custom serialization is in effect.

web/vibe/web/common.d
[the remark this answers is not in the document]
ferencdg: I could check those UDA's in the getInterfaceValidationError method, but I would need to iterate through all the functions and all the UDA's and then check a certain parameter. Would you prefer that?

web/vibe/web/common.d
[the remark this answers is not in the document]
s-ludwig: Okay, then let's just go with a template alias: `alias resultSerializer(alias ST, ...) = ResultSerializer!(ST, ...)` - should be equivalent for all practical intents.

web/vibe/web/common.d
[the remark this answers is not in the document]
s-ludwig: I don't even think that we need any explicit checks, just documenting it for `@resultSerializer` should be enough (would also be nice to have very a brief `unittest` example there).

web/vibe/web/rest.d
ferencdg: okay, just let quickly ask, if I understood correctly:
1. we lwill eave the original protected 'request' method in place, and we will implement it by using a call into our new private 'request' method
2. the new private 'request' method will no longer return a string, but it will accept a new parameter of type StreamOutputRange

is this correct?

s-ludwig: For number two I was thinking of a generic output range argument, but other than that that sounds correct.

The only issue is that the `protected request()` will have to re-parse the result to `Json` if it just calls the private one. So maybe we can move the serialization independent code to a separate method that is called by both and then let the old request just do its serialization the old way, instead of calling the new `private` one.

ferencdg: I was just thinking and I think we should pass an InputRange(or similar) to the deserializer. So shouldn't we should return HTTPClientResponse from the private 'request' method? and later we can directly pass the InputStream from the HTTPClientResponse to the deserializer.

ferencdg: although I am not sure if InterfaceProxy!InputStream can be converted to InputRange

web/vibe/web/common.d
```diff
@@ -479,6 +492,39 @@ package struct MethodAttribute
...
+alias resultSerializer(Args...) = ResultSerializer!(Args);
+
+///
+package alias DefaultSerializerT (FuncRetT) =
+	ResultSerializer!(
+		function (ref output_stream, ref value){
+			output_stream ~= serializeToJsonString(value);
+		},
+		function (input_stream){
```
ferencdg, line 511: I was unable to extract the actual stream object from the ProxyInterface as it currently requires knowing the type of the stream object compile time, but it is only available in runtime. As a result of that currently the type of the 'input stream' is ProxyInterface!InputStream. 

StreamInputRange also didn't work thanks to the ProxyInterface

ferencdg:
I think I made all the changes @Geod24 @s-ludwig, please let me know if I should change anything else

web/vibe/web/common.d
```diff
@@ -479,6 +492,39 @@ package struct MethodAttribute
...
+	alias deserialize = DT;
+}
+
+alias resultSerializer(Args...) = ResultSerializer!(Args);
+
+///
+package alias DefaultSerializerT (FuncRetT) =
+	ResultSerializer!(
+		function (ref output_stream, ref value){
+			output_stream ~= serializeToJsonString(value);
```
s-ludwig, line 509:
```suggestion
			serializeToJson(output_stream, value);
```

This can now use the range overload of `serializeToJson` to avoid the temporary copy. The parameter should be named `output_range`, tough, as this relies on a range interface rather than a stream one (same for `input_stream` below). It might make sense to go for a stream interface for certain serializers and large responses, but for now I think the range interface is more practical.

web/vibe/web/rest.d
```diff
@@ -1788,12 +1864,25 @@ private auto executeClientMethod(I, size_t ridx, ARGS...)
...
+
+	static if (!is(RT == void)) {
+		string content_type = "";
+		if (const hdr = "Content-Type" in reqhdrs)
+			content_type = *hdr;
+		enum result_serializers = ResultSerializersT!(Func);
+		immutable serializer_ind = get_matching_content_type!(result_serializers)(content_type);
+		foreach (i, serializer; result_serializers)
+			if (serializer_ind == i)
+    			return serializer.deserialize(ret.bodyReader);
```
s-ludwig, line 1882:
```suggestion
    			{
				auto rng = streamInputRange(ret.bodyReader);
				return serializer.deserialize(rng);
			}
```
This should provide a range interface to the serializer to be consistent with `serialize`.

web/vibe/web/rest.d
```diff
@@ -1859,47 +1948,46 @@ private Json request(URL base_url,
...
-
-		if (!isSuccessCode(cast(HTTPStatus)res.statusCode))
-			throw new RestException(res.statusCode, ret);
-	};
-
-	if (http_settings) requestHTTP(url, reqdg, resdg, http_settings);
-	else requestHTTP(url, reqdg, resdg);
+	HTTPClientResponse client_res;
+	if (http_settings) client_res = requestHTTP(url, reqdg, http_settings);
+	else client_res = requestHTTP(url, reqdg);
```
s-ludwig, line 1953: Just as a note, this degrades performance somewhat, because it currently causes the response object to be allocated on the heap instead of the stack, but this this will be a non-issue once the implementation is changed to vibe-http.

web/vibe/web/common.d
ferencdg: thanks makes sense for the input, however I couldn't get rid of using streams for the output, please see:
https://github.com/vibe-d/vibe.d/pull/2492/files#r526105742

web/vibe/web/rest.d
ferencdg: > This should provide a range interface to the serializer to be consistent with `serialize`.

please see: https://github.com/vibe-d/vibe.d/pull/2492/files#r526105742

web/vibe/web/common.d
```diff
@@ -461,8 +471,11 @@ class RestException : HTTPStatusException {
 		m_jsonResult = jsonResult;
 	}
 
-	/// The HTTP status code
-	@property const(Json) jsonResult() const { return m_jsonResult; }
+	/// The result text reported to the client
+	@property const(Json) jsonResult () const nothrow pure @safe @nogc
```
Geod24, line 475:
```suggestion
	@property inout(Json) jsonResult () inout nothrow pure @safe @nogc
```

web/vibe/web/common.d
```diff
@@ -479,6 +492,39 @@ package struct MethodAttribute
...
+
+alias resultSerializer(Args...) = ResultSerializer!(Args);
+
+///
+package alias DefaultSerializerT (FuncRetT) =
+	ResultSerializer!(
+		function (ref output_range, ref value){
+			 serializeToJson(output_range, value);
+		},
+		function (input_stream){
```
Geod24, line 511:
```suggestion
		function (ref output_range, ref value) {
			 serializeToJson(output_range, value);
		},
		function (input_stream) {
```

Style nits

web/vibe/web/rest.d
```diff
@@ -148,6 +148,50 @@
...
+			@resultSerializer!(
+				// output_stream implements OutputRange
+				function (output_stream, test_struct){
+					output_stream ~= serializeToJsonString(test_struct);
+				},
+				// input_stream implements InputStream
+				function (input_stream){
+					return deserializeJson!TestStruct(input_stream.readAllUTF8());
+				},
+				"application/json")()
```
Geod24, line 170: Same style nits

web/vibe/web/rest.d
```diff
@@ -1859,47 +1948,46 @@ private Json request(URL base_url,
...
+	// Get optional headers
+	foreach (k, ref v; optReturnHdrs.byKeyValue)
+		if (auto ptr = k in client_res.headers)
+			v = (*ptr).idup;
+		else
+			v = null;
+	if (missingKeys.length)
+		throw new Exception(
+			"REST interface mismatch: Missing required header field(s): "
+			~ missingKeys.to!string);
```
Geod24, line 1982: Note for the future: I'd really like us to improve this part so it doesn't allocate this much...

Geod24 (approved):
LGTM

s-ludwig:
Now dlang-bot has merged without a merge commit again ;)

Also, my comments w.r.t input stream vs. input range were not yet resolved. I'll try to prepare a follow-up PR for the recent changes ASAP.

Geod24:
> Now dlang-bot has merged without a merge commit again ;)

My bad... We probably shouldn't use `dlang-bot` for single-commit PRs then. Or perhaps make it configurable.

web/vibe/web/common.d
ferencdg: > thanks makes sense for the output, however I couldn't get rid of using streams for the input, please see:
> https://github.com/vibe-d/vibe.d/pull/2492/files#r526105742
> 
> I could create an input range from the stream by copying it, but this would be inefficient

@s-ludwig for the follow up PR, are you okay with my suggestion? 
"I could create an input range from the stream by copying it, but this would be inefficient"

web/vibe/web/rest.d
ferencdg: > > This should provide a range interface to the serializer to be consistent with `serialize`.
> 
> please see: https://github.com/vibe-d/vibe.d/pull/2492/files#r526105742

@s-ludwig for the follow up PR, are you okay with my suggestion?
"I could create an input range from the stream by copying it, but this would be inefficient"

web/vibe/web/common.d
s-ludwig: I didn't realize that both of those were the case - `StreamInputRange` only accepting `InputRange` and the JSON parser code requiring a forward range (for `skipNumber` only as it seems). I'll see what I can do there.

PR #2493, opened 2020-11-12
Added sink type toString to the serialization

ferencdg:
This change should reduce the number of memory allocations when using toString custom serialization

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -2009,6 +2009,25 @@ struct JsonStringSerializer(R, bool pretty = false)
 			m_range.put('\n');
 			foreach (i; 0 .. m_level) m_range.put('\t');
 		}
+
+		private void sink (scope const(char)[] data)
+		{
+			static if (isOutputRange!(R, char))
+			{
+				static if (__traits(compiles, m_range ~= data))
+						m_range ~= data;
+					else
+						foreach (const c; data)
+							m_range.put(c);
+			}
+		}
+
+		package void serializeSinkType (T) (scope auto ref T record)
+		{
+			m_range.put('"');
+			record.toString(&this.sink);
+			m_range.put('"');
+		}
 	}
 
 	//
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -39,6 +39,11 @@
 			$(LI Types satisfying the `isStringSerializable` trait will be
 				serialized as a string, as returned by their `toString`
 				method.)
+			$(LI Types satisfying the `isSinkSerializable` trait will be
+				serialized as a string using the toString(`SerializeSinkDg`)
+				function. toString(`SerializeSinkDg`) has precedence above
+				toString() if both are present. The serializer has to support the
+				`isSerializerSupportSinkType` trait)
 			$(LI Struct and class types by default will be serialized as
 				associative arrays, where the key is the name of the
 				corresponding field (can be overridden using the `@name`
@@ -475,7 +480,10 @@ private template serializeValueImpl(Serializer, alias Policy) {
 			ser.serializeValue!(CustomType, ATTRIBUTES)(value.toRepresentation());
 		} else static if (isISOExtStringSerializable!TU) {
 			ser.serializeValue!(string, ATTRIBUTES)(value.toISOExtString());
-		} else static if (isStringSerializable!TU) {
+		} else static if (isSinkSerializable!TU && isSerializerSupportSinkType!(Serializer, TU)) {
+			ser.serializeSinkType!TU(value);
+		}
+		else static if (isStringSerializable!TU) {
 			ser.serializeValue!(string, ATTRIBUTES)(value.toString());
 		} else static if (is(TU == struct) || is(TU == class)) {
 			static if (!hasSerializableFields!(TU, Policy))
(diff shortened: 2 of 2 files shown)


data/vibe/data/serialization.d
```diff
@@ -475,7 +475,10 @@ private template serializeValueImpl(Serializer, alias Policy) {
 			ser.serializeValue!(CustomType, ATTRIBUTES)(value.toRepresentation());
 		} else static if (isISOExtStringSerializable!TU) {
 			ser.serializeValue!(string, ATTRIBUTES)(value.toISOExtString());
-		} else static if (isStringSerializable!TU) {
+		} else static if (ObjectSupportSinkType!TU && SerializerSupportSinkType!(Serializer, TU)) {
```
ferencdg, line 478: ObjectSupportSinkType is not technically needed as SerializerSupportSinkType implicitly checks for the same

data/vibe/data/json.d
```diff
@@ -1925,6 +1926,17 @@ struct JsonStringSerializer(R, bool pretty = false)
 	this(R range)
 	{
 		m_range = range;
+		static if (isOutputRange!(R, char))
+		{
+			dg = (scope const(char)[] data)
+			{
+				static if (__traits(compiles, m_range ~= data))
+					m_range ~= data;
```
ferencdg, line 1934: in case the OutputRange supports more than just the put method(like with Appender) and operator~ is more optimized

Geod24: It would be better to have this as a comment in the code ;)

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
+///
+package alias SerializeSinkDg = void delegate(scope const(char)[]) @safe;
+
+version(unittest)
+{
+	import std.array : split;
+	import std.format : formattedWrite;
+	import vibe.data.json;
+
+	struct X (alias T)
```
Geod24, line 604:
```suggestion
	struct X (bool hasSink)
```

Would be more descriptive, wouldn't it ?

data/vibe/data/json.d
```diff
@@ -1925,6 +1926,17 @@ struct JsonStringSerializer(R, bool pretty = false)
...
+				else
+					foreach (const c; data)
+						m_range.put(c);
+			};
+		}
```
Geod24, line 1939: Having this delegate in the ctor has the downside of creating an extra GC allocation and an extra reference to `m_range` through the context.
It would be much better to move this delegate to the function that pass it (`serializeSinkType`) and use `scope`.
But even better, you could just make it a private method and pass `&this.sink` at L2028 instead of `dg`.

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
 }
 
+///
+package template isSerializerSupportSinkType (SerT, ObjT)
+{
+	enum isSerializerSupportSinkType = is(typeof(SerT.serializeSinkType!(ObjT)(ObjT.init)) == void);
+}
+
+///
+package template isObjectSupportSinkType (ObjectT)
```
Geod24, line 589:
```suggestion
package template isSinkSerializable (ObjectT)
```

I think the name is shorter and still fits with the existing style, what do you think ?

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
+{
+	enum isObjectSupportSinkType = is(typeof(ObjectT.toString(cast(SerializeSinkDg) null)) == void) &&
```
Geod24, line 591:
```suggestion
	enum isObjectSupportSinkType = is(typeof(ObjectT.toString(SerializeSinkDg.init))) &&
```

`X.init` is a neat way to get a default value typed as `X`. Remember, every cast is an aberration ;)
And I don't think we need to force the return type to be `void`. Why would we ?

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
+	import std.format : formattedWrite;
+	import vibe.data.json;
+
+	struct X (alias T)
+	{
+		private int i;
+		private string s;
+		this (int i, string s) @safe {this.i = i; this.s = s;}
+
+		static if(T)
```
Geod24, line 610: Space

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
+{
+	import std.array : split;
+	import std.format : formattedWrite;
+	import vibe.data.json;
+
+	struct X (alias T)
+	{
+		private int i;
+		private string s;
+		this (int i, string s) @safe {this.i = i; this.s = s;}
```
Geod24, line 608: If it's a struct, you don't need the ctor, do you ?
If you do, missing attributes.

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
+
+		static if(T)
+		{
+			public void toString (SerializeSinkDg dg) @safe
+			{
+				formattedWrite(dg, "%d;%s", this.i, this.s);
+			}
+		}
+
+		public string toString () @safe
```
Geod24, line 618:
```suggestion
		public string toString () const @safe
```

data/vibe/data/serialization.d
```diff
@@ -571,6 +579,67 @@ private template serializeValueImpl(Serializer, alias Policy) {
 	}
 }
 
+///
+package template isSerializerSupportSinkType (SerT, ObjT)
+{
+	enum isSerializerSupportSinkType = is(typeof(SerT.serializeSinkType!(ObjT)(ObjT.init)) == void);
```
Geod24, line 585: Same comment about `void` return as below

Geod24 (approved):
The duplication between serializer and object handling looks a bit odd to me, but I'm not familiar enough with the code to suggest an alternative (or be 100% sure there's a better way either). So LGTM, minus a few comments. @s-ludwig ?

ferencdg:
> The duplication between serializer and object handling looks a bit odd to me, but I'm not familiar enough with the code to suggest an alternative (or be 100% sure there's a better way either). So LGTM, minus a few comments. @s-ludwig ?

we have serializers like JsonSerializer(differenet from JsonStringSerializer) and JsonSerializer writes into a Json temporary object instead of an OutputRange. Because of this my thinking was that this sink serialization has to be supported by both the object and the serializer

ferencdg:
thans for the comments @Geod24, I fixed them(same for #2492)

Geod24 (approved):

data/vibe/data/json.d
```diff
@@ -2009,6 +2009,25 @@ struct JsonStringSerializer(R, bool pretty = false)
 			m_range.put('\n');
 			foreach (i; 0 .. m_level) m_range.put('\t');
 		}
+
+		private void sink (scope const(char)[] data)
+		{
+			static if (isOutputRange!(R, char))
+			{
+				static if (__traits(compiles, m_range ~= data))
```
s-ludwig, line 2017: This needs to perform `jsonEscape`, or is the idea is that the object is coupled with the output serialization format and can write arbitrary data?

PR #2494, opened 2020-11-14
Maintainance

s-ludwig:


Geod24 (approved):
I guess you'll do the release notes later ?

s-ludwig:
Yeah, I usually do it as a batch process before each release. Doing it per-change has advantages, but also requires constant attention and causes friction for non-core contributors, which is why I prefer the batch approach.

I'd tag a proper release shortly, though, we just should get #2475 in. Next step would then be to look at the open API issues to target a 1.0.0 release.

Geod24:
Alright, I'll get back to #2475

Issue #2415, opened 2020-02-19
REST generator rejects `ref` parameter


Geod24:
I have the following code:
```D
import vibe.web.rest;

struct NC
{
    @disable this(this);

    int a;
    string b;
}

interface API
{
    void sendData (ref NC data);
}

void main()
{
    NC nc = NC(42, "42");
    auto client = new RestInterfaceClient!API("127.0.0.1");
    client.sendData(nc);
}
```

The REST generator errors out with:
```
restref ~master: building configuration "application"...
../../../../.dub/packages/vibe-d-0.8.6/vibe-d/web/vibe/web/internal/rest/common.d(59,2): Error: static assert:  "app.API.sendData: Parameter 'data' cannot be ref"
../../../../.dub/packages/vibe-d-0.8.6/vibe-d/web/vibe/web/rest.d(596,23):        instantiated from here: RestInterface!(API)
source/app.d(19,23):        instantiated from here: RestInterfaceClient!(API)
dmd failed with exit code 1.
```

However, since my struct is non-copyable, I cannot make this argument non-ref.
The use case is sending a struct that is `extern(C++)` and contains non-trivial copy constructors (some fields are `std::vector`).

s-ludwig:
I wonder if we could simply allow `auto ref` for that purpose.

s-ludwig:
Another possibility would be `in ref` or some additional UDA+`ref`. Or `ref` could simply be allowed for any parameter, AFAICS the only reason for the check is to avoid misuse by expecting the modified value to propagate to the caller.

Geod24:
Ah, that makes a lot of sense. I think `in ref` would be good (and enough for my use case).

Geod24:
Well, now that `-preview=in` exists, the same error triggers with `in` parameter, and I think it would make a lot of sense to recognize this.

PR #2495, opened 2020-11-16
Allowing refs in REST interface

ferencdg:
Following comments on https://github.com/vibe-d/vibe.d/issues/2415#issuecomment-588079108
closes: #2415

### DIFF
--- a/web/vibe/web/rest.d
+++ b/web/vibe/web/rest.d
@@ -2062,10 +2062,10 @@ do {
 			}
 		}
 
-		// Check for misplaced ref / out
+		// Check for misplaced out and non-const ref
 		alias PSC = ParameterStorageClass;
 		foreach (i, SC; ParameterStorageClassTuple!Func) {
-			static if (SC & PSC.out_ || SC & PSC.ref_) {
+			static if (SC & PSC.out_ || (SC & PSC.ref_ && !is(ConstOf!(PT[i]) == PT[i])) ) {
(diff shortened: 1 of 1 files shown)


web/vibe/web/rest.d
```diff
@@ -2062,10 +2062,10 @@ do {
 			}
 		}
 
-		// Check for misplaced ref / out
+		// Check for misplaced out
 		alias PSC = ParameterStorageClass;
 		foreach (i, SC; ParameterStorageClassTuple!Func) {
-			static if (SC & PSC.out_ || SC & PSC.ref_) {
+			static if (SC & PSC.out_) {
```
ferencdg, line 2068: following the comments on https://github.com/vibe-d/vibe.d/issues/2415#issuecomment-588079108

we could me more specific and just reject 'ref's, but allow 'in refs', but considering the long and complicated history of 'in' I chose this simple solution

checking for 'in' would require taking compiler versions and compiler flags (-preview=in) into consideration

Geod24: I suggest that instead of just allowing `ref`, only allow `ref` on a `const` parameter.
This will also cover `in [ref]` as a nice side effect, with or without the switch :)

ferencdg: okay, just made a new checkin

Geod24 (approved):
LGTM. @s-ludwig ?

PR #2496, opened 2020-11-20
Adding support for banning, un-banning TCP clients.

ferencdg:
This change will help to mitigate simple DOS attacks

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -200,6 +200,12 @@ void handleHTTPConnection(TCPConnection connection, HTTPServerContext context)
 
 	scope (exit) connection.close();
 
+	// check wether the client's address is banned
+	foreach (ref virtual_host; context.m_virtualHosts)
+		if ((virtual_host.settings.isBannedDg !is null) &&
+			virtual_host.settings.isBannedDg(connection.remoteAddress()))
+			return;
+
 	// Set NODELAY to true, to avoid delays caused by sending the response
 	// header and body in separate chunks. Note that to avoid other performance
 	// issues (caused by tiny packets), this requires using an output buffer in
@@ -632,6 +638,17 @@ final class HTTPServerSettings {
 		gets a request.
 	*/
 	string hostName;
+
+	/** Provides a way to ban and unban network addresses and reduce the
+		impact of DOS attack.
+
+		If the isBannedDg returns true for a specific NetworkAddress,
+		then all incoming requests from that address will be rejected.
+	*/
+	IsBannedDg isBannedDg;
+
+	/// Type of delegate accepted for `isBannedDg`
+	alias IsBannedDg = bool delegate (in NetworkAddress) @safe nothrow;
 
 	/** Configures optional features of the HTTP server
 


http/vibe/http/server.d
```diff
@@ -200,6 +201,16 @@ void handleHTTPConnection(TCPConnection connection, HTTPServerContext context)
 
 	scope (exit) connection.close();
 
+	// check wether the client's address is banned
+	foreach (ref virtual_host; context.m_virtualHosts)
+		foreach (ref ban_manager; virtual_host.settings.banManagers)
+		{
+			string remote_address = connection.remoteAddress().toAddressString();
+			ban_manager.onRequest(remote_address);
```
ferencdg, line 209: calling onRequest will allow implementing more sophisticated BanManagers that could throttle the number of requests per clients

Geod24:
I think this is way too specific to our use case. What I (originally) had in mind was just the ability to have a callback that returns `bool`, not to move the `BanManager` in Vibe.d

ferencdg:
> I think this is way too specific to our use case. What I (originally) had in mind was just the ability to have a callback that returns `bool`, not to move the `BanManager` in Vibe.d

I was thinking that vibe.d would also need this feature to handle the following scenraio: each IP can call REST endpoints only a certain number of times. This would be implemented by a different kind of BanManager that implements IBanManager and that implementation most probably should be in vibe.d

Geod24:
I think this could be handled by just passing the `HTTPServerRequest` to the client and let the client decide. It can use whatever is there then.

ferencdg:
> I think this could be handled by just passing the `HTTPServerRequest` to the client and let the client decide. It can use whatever is there then.

that is true, but then every client would need to write its own versioin versus having some useful/common ban logic in vibe.d

so in your suggested solution vibe.d would only receive a callback function called isBanned. This is very similar to having IBanManager with only 1 method: isBanned. In fact currently Rest only calls 2 methods on the IBanManager: isBanned, onRequest, this is quite similar to what you suggest I think.

Maybe I should would narrow the IBanManager interface to those 2(or maybe just to the 1) methods and move out our BanManager implementation from vibe.d(although other projects cannot use that implementation then)

s-ludwig:
I feel like this functionality should be fully fleshed out in a separate package before actually integrating it into the core library. It definitely has valid uses outside of the HTTP sub system, too, and I think that more metrics that are useful for the banning decision will emerge. Considering that `vibe.http` is in the (suspended) process of going to 1.0.0, it would be a bad idea to add APIs that are likely to be changed again.

What would probably make sense though is the `isBanned` callback in the form of a single `bool delegate(in ref NetworkAddress) @safe nothrow`, because then at some point the decision could be done before actually establishing the connection, at least on Windows using the `lpfnCondition` argument to `WSAAccept`.

Geod24:
> What would probably make sense though is the isBanned callback in the form of a single bool delegate(in ref NetworkAddress) @safe nothrow, because then at some point the decision could be done before actually establishing the connection, at least on Windows using the lpfnCondition argument to WSAAccept.

So IIUC, replace L653 (https://github.com/vibe-d/vibe.d/pull/2496/files#diff-982fb1ea1452e69afa1364a57a9ad8ff741507348c9cf29e4fd65da0ea5d7200R653) with a delegate ?
And yes, refusing the connection as early as possible is the use case here.

s-ludwig:
> So IIUC, replace L653 (https://github.com/vibe-d/vibe.d/pull/2496/files#diff-982fb1ea1452e69afa1364a57a9ad8ff741507348c9cf29e4fd65da0ea5d7200R653) with a delegate ?

Correct

http/vibe/http/server.d
```diff
@@ -633,6 +642,15 @@ final class HTTPServerSettings {
 	*/
 	string hostName;
 
+	/** Provides a way to ban and unban network addresses and reduce the
+		impact of DOS attack.
+
+		If any of the ban managers bans an address, then all incoming requests
+		from that address will be rejected.
+	*/
+	alias IsBannedDlg = bool delegate (ref in NetworkAddress) @safe nothrow;
```
Geod24, line 651:
```suggestion
	alias IsBannedDg = bool delegate (in NetworkAddress) @safe nothrow;
```

http/vibe/http/server.d
```diff
@@ -633,6 +642,15 @@ final class HTTPServerSettings {
...
+	IsBannedDlg[string] isBannedDlgs;
```
Geod24, line 652: Why is this an AA ? A simple delegate would do.

Geod24 (changes requested):
Two things:
- I think this should be handled directly at the TCP connection level, not just the HTTP;
  But then I went and looked at vibe-core and realized there was no way to do this without invasive changes, so let's keep this idea on the shelve;
- I think the whole "banman" package should go. We only need the change in `vibe/http/server.d` so far.

http/vibe/http/server.d
ferencdg: My idea was to allow multiple different types of BanManagers and if any of them would drop the connection, then it will be dropped. Should I just allow 1 BanManager?

Geod24: One delegate is enough. If a user needs multiple delegates, they can just wrap them in one delegate that do the iteration.

http/vibe/http/server.d
```diff
@@ -200,6 +200,12 @@ void handleHTTPConnection(TCPConnection connection, HTTPServerContext context)
 
 	scope (exit) connection.close();
 
+	// check wether the client's address is banned
+	foreach (ref virtual_host; context.m_virtualHosts)
+		if ((virtual_host.settings.isBannedDg !is null) &&
```
Geod24, line 205: You should move this `if` to the outer scope to avoid needless iteration

http/vibe/http/server.d
```diff
@@ -633,6 +639,15 @@ final class HTTPServerSettings {
...
 	string hostName;
 
+	/** Provides a way to ban and unban network addresses and reduce the
+		impact of DOS attack.
+
+		If any of the ban managers bans an address, then all incoming requests
+		from that address will be rejected.
+	*/
+	alias IsBannedDg = bool delegate (in NetworkAddress) @safe nothrow;
+	IsBannedDg isBannedDg;
```
Geod24, line 649:
```suggestion
	*/
	IsBannedDg isBannedDg;
	
	/// Type of delegate accepted ffor `isBannedDg`
	alias IsBannedDg = bool delegate (in NetworkAddress) @safe nothrow;
```

Otherwise the DDOC will be wrong.

Geod24 (approved):

PR #2497, opened 2020-11-24
Do not implicitly use deprecated ProxyStream ctor

Geod24:
```
A deprecation message was triggered in client code because 'opAssign'
was being used, which forwarded to emplace, which forwards to 'std.conv.to',
which attempt to use it.
Instead of constructing then assigning, we not construct directly.
```

This will silence the following messages:
```
vibe-d:http ~master: building configuration "library"...
/usr/local/Cellar/ldc/1.24.0/include/dlang/ldc/std/conv.d(4698,38): Deprecation: constructor vibe.stream.wrapper.ProxyStream.this is deprecated - Use createProxyStream instead.
/usr/local/Cellar/ldc/1.24.0/include/dlang/ldc/std/conv.d(4702,21): Deprecation: constructor vibe.stream.wrapper.ProxyStream.this is deprecated - Use createProxyStream instead.
```

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -486,10 +486,8 @@ HTTPServerResponse createTestHTTPServerResponse(OutputStream data_sink = null,
 		settings.sessionStore = session_store;
 	}
 
-	InterfaceProxy!Stream outstr;
-	if (data_sink && data_mode == TestHTTPResponseMode.plain)
-		outstr = createProxyStream(Stream.init, data_sink);
-	else outstr = createProxyStream(Stream.init, nullSink);
+	InterfaceProxy!Stream outstr = createProxyStream(Stream.init,
+		data_sink && data_mode == TestHTTPResponseMode.plain ? data_sink : nullSink);
 
 	auto ret = new HTTPServerResponse(outstr, InterfaceProxy!ConnectionStream.init,
 		settings, () @trusted { return vibeThreadAllocator(); } ());


PetarKirov (approved):

s-ludwig:
This should actually already be fixed by https://github.com/vibe-d/vibe-core/pull/234 - I just didn't tag a new version, yet.

Geod24:
Right. But aren't both fixes useful in their own right ?

s-ludwig:
It doesn't really matter now, but as soon as the two streams are `structs` instead of classes, the code won't compile anymore due to a missing common type (which should be the case in vibe-http, although it may not be, yet). I've tagged vibe-core 1.11.1 in the meantime so at least the deprecation warning is a non-issue now.

Geod24:
Isn't that true for the old code too ?

s-ludwig:
`createProxyStream` always returns a `InterfaceProxy!Stream`, regardless of the inputs, so with `if` instead of `?:`, to avoid the common type requirement for the input arguments, everything should be fine.

Geod24:
Ah, I was thinking about an earlier version. If the sink type change and don't have a common type, we can simply take the ternary to the outer scope, essentially doing `x = a ? ccreateProxyStream(...) : createProxyStream(b)`

Geod24:
Should I do this now ? It was my original change (it's the least "distance" from the original code), but then I realized I could simplify it further.

Geod24:
Ping @s-ludwig

Geod24:
Ping @s-ludwig

s-ludwig:
So as far as I see it, we should rather drop this, because although the code arguably gets a bit nicer, it doesn't have a practical benefit at this point anymore and it would have to be reverted/skipped when patching up vibe-http with all the latest changes from this repository, nothing big, but one more little complication overall. But I don't feel strongly about this in any way, so if you think that it's still worth it, feel free to go ahead.

In any case, I'll try to somehow fit in the vibe:http->vibe-http transition in during the coming weeks, so that the two can be decoupled and maintained individually (vibe:http only receiving fixes at that point).

Geod24:
Well if the code is soon to be changed to do construction instead of assignment, then indeed there's no practical point.

PR #2500, opened 2020-11-25
Replace travis-ci with Github Actions

Geod24:
This will be a bit problematic for Buildkite.

@s-ludwig : See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing and the result: https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

This only does Ubuntu, as currently we can't merge any PR because a "Required" check (Travis) is not working anymore.
In the future I will try to add Mac.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,56 @@
+# Run Vibe.d's test suite on all three platforms
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  main:
+    name: Run
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest ]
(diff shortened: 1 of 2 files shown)


s-ludwig:
Probably makes sense to also add Windows (as for [vibe-core](https://github.com/vibe-d/vibe-core/pull/230)) at some point.  Seems like the workflow hasn't generated any jobs, though.

Geod24:
Yeah, I wanted to add macOS / Windows, but since all PRs are being blocked from being merged, I figured I'd start slow :)
You don't see the checks in this PR because it wasn't merged. However, you can see the run in my fork: https://github.com/Geod24/vibe.d/actions/runs/382413068

There's still the question of the failing test. For some reason, the TLS client tunnel succeeds when a failure is expected. I assume that's because of configuration difference between the new mongodb/redis and the old one (e.g. the old one didn't support TLS?). It would be great if you could take a look.

s-ludwig:
Okay, I'll merge like this and then take a look at the test, so that I can actually see the failure.

PR #2501, opened 2020-11-25
Re-enable TLS test

s-ludwig:


RazvanN7:
@s-ludwig This seems to have broken the DLang test suite.  Any ideas on how to fix this?

s-ludwig:
Hum, interesting, so this seems like it may be OpenSSL version related. I'll change this so that it doesn't care about the client side result.

s-ludwig:
Opened a PR: #2550

PR #2503, opened 2020-11-26
Serialization rework

s-ludwig:
This follows up on #2492 and #2493 and addresses the remaining issues. @ferencdg, please comment on any issues you might see, I could have broken a use case that I didn't see here.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -1260,7 +1260,7 @@ struct Json {
 	Throws a JSONException if any parsing error occured.
 */
 Json parseJson(R)(ref R range, int* line = null, string filename = null)
-	if( is(R == string) )
+	if (isForwardRange!R)
 {
 	Json ret;
 	enforceJson(!range.empty, "JSON string is empty.", filename, 0);
@@ -1276,17 +1276,20 @@ Json parseJson(R)(ref R range, int* line = null, string filename = null)
 	bool minus = false;
 	switch( range.front ){
 		case 'f':
-			enforceJson(range[1 .. $].startsWith("alse"), "Expected 'false', got '"~range[0 .. min(5, $)]~"'.", filename, line);
+			enforceJson(range.save.dropOne.startsWith("alse"),
+				"Expected 'false', got '"~range.take(5).to!string~"'.", filename, line);
 			range.popFrontN(5);
 			ret = false;
 			break;
 		case 'n':
-			enforceJson(range[1 .. $].startsWith("ull"), "Expected 'null', got '"~range[0 .. min(4, $)]~"'.", filename, line);
+			enforceJson(range.save.dropOne.startsWith("ull"),
+				"Expected 'null', got '"~range.take(4).to!string~"'.", filename, line);
 			range.popFrontN(4);
 			ret = null;
 			break;
 		case 't':
-			enforceJson(range[1 .. $].startsWith("rue"), "Expected 'true', got '"~range[0 .. min(4, $)]~"'.", filename, line);
+			enforceJson(range.save.dropOne.startsWith("rue"),
+				"Expected 'true', got '"~range.take(4).to!string~"'.", filename, line);
 			range.popFrontN(4);
 			ret = true;
 			break;
@@ -1350,7 +1353,7 @@ Json parseJson(R)(ref R range, int* line = null, string filename = null)
 			ret = obj;
 			break;
 		default:
-			enforceJson(false, format("Expected valid JSON token, got '%s'.", range[0 .. min(12, $)]), filename, line);
+			enforceJson(false, format("Expected valid JSON token, got '%s'.", range.take(12)), filename, line);
 			assert(false);
 	}
 
@@ -1358,6 +1361,24 @@ Json parseJson(R)(ref R range, int* line = null, string filename = null)
 	version(JsonLineNumbers) ret.line = curline;
 	return ret;
 }
+
+
+unittest { // ensure parseJson works with a generic forward range
+	static struct R {
+		const(char)[] text;
+
+		@property char front() { return text[0]; }
+		@property R save() { return this; }
+		@property bool empty() const { return text.length == 0; }
+		void popFront() { text = text[1 .. $]; }
+	}
+
+	auto r = R(`{"i":42, "s": "foo"}`);
+	auto j = parseJson(r);
+	assert(j["i"] == 42);
+	assert(j["s"] == "foo");
+}
+
 
 /**
 	Parses the given JSON string and returns the corresponding Json object.
@@ -1556,7 +1577,7 @@ T deserializeJson(T)(Json src)
 }
 /// ditto
 T deserializeJson(T, R)(R input)
-	if (!is(R == Json) && isInputRange!R)
+	if (!is(R == Json) && isForwardRange!R)
 {
 	return deserialize!(JsonStringSerializer!R, T)(input);
 }
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -2007,6 +2007,27 @@ struct JsonStringSerializer(R, bool pretty = false)
 			} else static assert(false, "Unsupported type: " ~ UT.stringof);
 		}
 
+		void writeStringSinkValue(Traits, T)(scope auto ref T value)
+		{
+			void sink(scope const(char)[] str) {
+				m_range.jsonEscape(str);
+			}
+
+			final class R {
+				void put(char ch) { put(() @trusted { return (&ch)[0 .. 1]; } ()); }
+				void put(scope const(char)[] str) { m_range.jsonEscape(str); }
+			}
+
+			m_range.put('"');
+			static if (__traits(compiles, value.toString((scope s) => sink(s)))) {
+				value.toString((scope s) => sink(s));
+			} else {
+				scope r = new R;
+				value.toString(r);
+			}
+			m_range.put('"');
+		}
+
 		private void startComposite()
 		{
 			static if (pretty) m_level++;
@@ -2036,25 +2057,6 @@ struct JsonStringSerializer(R, bool pretty = false)
 		{
 			m_range.put('\n');
 			foreach (i; 0 .. m_level) m_range.put('\t');
-		}
-
-		private void sink (scope const(char)[] data)
-		{
-			static if (isOutputRange!(R, char))
-			{
-				static if (__traits(compiles, m_range ~= data))
-						m_range ~= data;
-					else
-						foreach (const c; data)
-							m_range.put(c);
-			}
-		}
-
-		package void serializeSinkType (T) (scope auto ref T record)
-		{
-			m_range.put('"');
-			record.toString(&this.sink);
-			m_range.put('"');
 		}
 	}
 
@@ -2186,6 +2188,18 @@ struct JsonStringSerializer(R, bool pretty = false)
 			return true;
 		}
 	}
+}
+
+unittest {
+	static assert(doesSerializerSupportStringSink!(JsonStringSerializer!(Appender!string)));
+
+	auto app = appender!string;
+	auto ser = JsonStringSerializer!(Appender!string)(app);
+	static struct T1 { void toString(scope void delegate(scope const(char)[])) {} }
+	static struct T2 { void toString(R)(scope ref R dst) { dst.put('f'); dst.put("foo"); } }
+
+	ser.writeStringSinkValue!void(T1.init);
+	ser.writeStringSinkValue!void(T2.init);
 }
 
 /// Cloning JSON arrays
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -595,7 +595,7 @@ package template doesSerializerSupportStringSink(SerT)
 }
 
 ///
-package template isStringSinkSerializable(T)
+template isStringSinkSerializable(T)
 {
 	import std.range : nullSink;
 
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -609,13 +609,12 @@ package template isStringSinkSerializable(T)
 		&& is(typeof(T.fromString(string.init)) : T);
 }
 
-version(unittest)
-{
+unittest {
 	import std.array : split;
 	import std.format : formattedWrite;
 	import vibe.data.json;
 
-	struct X(alias hasSink) {
+	static struct X(alias hasSink) {
 		private int i;
 		private string s;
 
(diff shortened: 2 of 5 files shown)


stream/vibe/stream/wrapper.d
```diff
@@ -254,7 +258,8 @@ struct StreamInputRange {
 		Buffer* m_buffer;
 	}
 
-	this (InputStream stream)
+	/// private
+	this(InputStream stream)
```
Geod24, line 262: Why the `private` comment instead of actually being `private` ?

stream/vibe/stream/wrapper.d
```diff
@@ -242,9 +242,13 @@ class ConnectionProxyStream : ConnectionStream {
 	after a call to front. This property allows the range to be used in
 	request-response scenarios.
 */
-struct StreamInputRange {
+deprecated("Use streamInputRange() instead.")
+StreamInputRange!OutputStream StreamInputRange()(InputStream stream) { return StreamInputRange!InputStream(stream); }
```
Geod24, line 246: Looking at the diff, it looks like this function didn't exists before, but is added as deprecated ? What am I missing ?

web/vibe/web/common.d
```diff
@@ -492,25 +492,70 @@ package struct MethodAttribute
...
 
+
+/** Uses a custom serializer for the method return value.
+
+	Instead of using the default serializer (JSON), this allows to define
+	custom serializers. Multiple serializers can be specified and will be
+	matched against the `Accept` header of the HTTP request.
+*/
+alias resultSerializer(alias serialize, alias deserialize, string content_type)
+	= ResultSerializer!(serialize, deserialize, content_type);
```
Geod24, line 503: Can you add a `Params` section ?
Also maybe name them `serializeFn` and `deserializeFn` to make things a bit clearer ?

web/vibe/web/common.d
```diff
@@ -492,25 +492,70 @@ package struct MethodAttribute
...
+	interface MyRestInterface {
+		static struct Point {
+			int x, y;
+		}
+
+		static void serialize(R)(ref R output_range, ref Point value)
+		{
+			output_range.put(nativeToBigEndian(value.x));
+			output_range.put(nativeToBigEndian(value.y));
+		}
```
Geod24, line 518: Can you extend the example to use an overload set (on `value`) ?
Also, maybe make `value` be `const` ?

web/vibe/web/rest.d
```diff
@@ -1603,12 +1603,14 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
...
-						auto serialized_output = appender!string;
-						static if (!__traits(compiles, () @safe {
-							serializer.serialize(serialized_output, ret);
-						}))
+						auto serialized_output = appender!(ubyte[]);
+						static if (
+							__traits(compiles, () @trusted { serializer.serialize(serialized_output, ret); })
+							&& !__traits(compiles, () @safe { serializer.serialize(serialized_output, ret); }))
+						{
 							pragma(msg, "Non-@safe serialization of REST return types deprecated - ensure that " ~
```
Geod24, line 1611: I really hate this BTW :D

data/vibe/data/json.d
```diff
@@ -2188,6 +2190,18 @@ struct JsonStringSerializer(R, bool pretty = false)
 	}
 }
 
+unittest {
+	static assert(doesSerializerSupportStringSink!(JsonStringSerializer!(Appender!string)));
+
+	auto app = appender!string;
+	auto ser = JsonStringSerializer!(Appender!string)(app);
+	static struct T1 { void toString(scope void delegate(scope const(char)[])) {} }
+	static struct T2 { void toString(R)(scope ref R dst) { dst.put('f'); dst.put("foo"); } }
```
Geod24, line 2199: Okay now I understand better what you meant. I don't really see a need for the second overload myself, as the `delegate` is a superset of this. It's trivial to wrap an output range in a delegate and does not allocate.

Geod24 (approved):
Nice! A couple comments. Would this also allow one to use `std_data_json` instead of `vibe.data.json` ? Maybe an example of this could be useful for user code.

stream/vibe/stream/wrapper.d
s-ludwig: Hm, I just copied it from the way `StreamOutputRange` does it, but at a second glance it looks like it could indeed just be properly `private`.

stream/vibe/stream/wrapper.d
s-ludwig: This is a backwards-compatibility replacement of the old constructor, which was simply used as `StreamInputRange(stream)`. Since the struct is now wrapped in a template, that syntax is otherwise not possible anymore.

web/vibe/web/common.d
s-ludwig: I thought about it, but was afraid ddoc would complain because they are not function parameters, but I'm probably wrong. I'll try it out.

web/vibe/web/rest.d
s-ludwig: Looks like a candidate for removal for 1.0.0

data/vibe/data/json.d
s-ludwig: True, but I think it makes sense to try to be compatible with [`formatValue`](https://dlang.org/library/std/format/format_value.html) in that regard, modulo the `fmt` parameter, since ranges are more idiomatic today and such an overload may already be defined for non-serialization purposes.

web/vibe/web/common.d
Geod24: Yeah DDOC supports that use case.

web/vibe/web/rest.d
Geod24: I mean, the fact that `@safe` is forced on the user :/

s-ludwig: That's very much as intended, because we are talking about a REST service that has no business in deliberately choosing the memory-unsafe mode by default. If someone really wants it that way, it's always possible to slap on a `@trusted`.

s-ludwig:
Hm, while writing the parameter documentation for `resultSerializer` I realized that the definition should probably changed, so that `deserialize` is called either as `deserialize!T`, or as `deserialize(range, [ref] result)` to allow generic serializers.

s-ludwig:
> Would this also allow one to use std_data_json instead of vibe.data.json?

In theory yes, but is there a serializer implementation for it?

stream/vibe/stream/wrapper.d
```diff
@@ -242,9 +242,13 @@ class ConnectionProxyStream : ConnectionStream {
...
+/// ditto
+struct StreamInputRange(InputStream, size_t buffer_size = 256)
+	if (isInputStream!InputStream)
+{
 @safe:
-
 	private {
 		struct Buffer {
 			ubyte[256] data = void;
```
ferencdg, line 254:
```suggestion
ubyte[buffer_size] data = void;
```

ferencdg, line 254:
```suggestion
ubyte[buffer_size] data = void;
```

web/vibe/web/common.d
s-ludwig: Made it const and turned it into a template parameter. Overload set would also be possible with the latest changes, that could maybe be added as a second example? But it does get quite lengthy then.

data/vibe/data/serialization.d
```diff
@@ -480,10 +479,15 @@ private template serializeValueImpl(Serializer, alias Policy) {
...
-		else static if (isStringSerializable!TU) {
+		} else static if (isStringSinkSerializable!TU) {
+			static if (doesSerializerSupportStringSink!Serializer) {
+				ser.writeStringSinkValue!Traits(value);
+			} else {
+				auto app = appender!string;
+				app.formattedWrite("%s", value);
+				ser.serializeValue!(string, ATTRIBUTES)(app.data);
+			}
+		} else static if (isStringSerializable!TU) {
```
ferencdg, line 490: I am thinking whether we should change the order of
(line 482) else static if (isStringSinkSerializable!TU) 
(line 490) else static if (isStringSerializable!TU)
for backword compatibility

I think classes might already have toString(sink) type methods, because std functions like formattedwrite already supports them. After this change is merged this sink type toString(sink) will be used instead of the normal toString which might suprises users.

ferencdg:
thanks for cleaning up my PR @s-ludwig

data/vibe/data/serialization.d
s-ludwig: Yeah, this might be an issue, although I'd expect in 99% of the cases that both overloads produce identical results and of course they *should* always do that. The fact that turning this around would force the user code to completely remove the string returning overload (if performance is desired) sounds like the worse trade-off to me, because having such an overload for convenience is quite common from my experience.

Geod24: @ferencdg : Having the sink overload take precedence is a well-established practice, actually.
`std.format` does that, and some utilities (e.g. `Throwable.toString`) take advantage of it. Ocean does that too.

web/vibe/web/rest.d
Geod24: Well, I don't think we should force a different default on the user.
My understanding is that was put in place because we had to make a choice: be usable from `@safe` code, or not. You can't really choose to support both at the moment, which is a huge PITA to write libraries. So I understand why it's seen as preferable to support `@safe`, but it's quite annoying to force this on the user.

Re: slapping `@trusted`, what's the point of attributes if you just bypass it ?

Anyway, that's slightly OT.

Geod24:
I don't think the `deserializer` should take its argument by `ref`.
This creates tons of issues, as it:
1) Forces to have a valid state on `init` (e.g. `invariant`) and that default construction not be disabled;
2) Might require copy because NRVO is less likely to be performed;
3) Prevents working with qualified types (const / immutable).

s-ludwig:
> I don't think the deserializer should take its argument by ref.
This creates tons of issues, as it:
>
> 1. Forces to have a valid state on init (e.g. invariant) and that default construction not be disabled;
> 2. Might require copy because NRVO is less likely to be performed;
> 3. Prevents working with qualified types (const / immutable).

All of these would only work under the assumption that the serializer performs a recursive construction of the result value using constructor calls. I don't think that this is realistic for a generic serializer for multiple reasons.

Also, the init state is practically mandated by the language to be valid anyway. Returning large structs by value, where performing an extra copy is a performance issue (in the same order of magnitude as the serialization and the network transfer), doesn't sound like a good idea, too, at least I wouldn't count on the library handling it with NRVO throughout if I were a user.

So while I do understand that these are issues in theory, I'm not really convinced that they are in practice. The counter argument is pretty simple, but maybe you have a better idea how to solve it:

Having it as a return value means that `deserialize` must be a template with the value type as its first template argument, which in turn means that overload sets won't work and type-specific serializers need to awkwardly define their fixed type as a template argument. Or we revert it to what it was before and generic serializers don't work.

Geod24:
> All of these would only work under the assumption that the serializer performs a recursive construction of the result value using constructor calls. I don't think that this is realistic for a generic serializer for multiple reasons.

Well, constructor is one thing, but the serializer an also define a hook (a la `toString`) for user-customization.

> Also, the init state is practically mandated by the language to be valid anyway.

It's mandated to exist. But some types *need* to be constructed to be valid, which is a common complain in D.

> So while I do understand that these are issues in theory, I'm not really convinced that they are in practice. The counter argument is pretty simple, but maybe you have a better idea how to solve it:

Well, those are drawn from experience writing a deserializer. Namely, [this](https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/common/Serializer.d).

The reason for doing so was that we needed to have, in our REST API, types that are not under our control. Those types are `extern(C++)` and have copy constructors / destructors which we do not want triggered. Currently we just didn't declare the copy ctors (so they don't get called, ever), but that's a hack.

The TL;DR from my experience was: If you want a generic serializer, it needs to handle const/immutable/shared.
The only proper way to handle `const`/`immutable`/`shared` is to be able to initialize your structs to the value you want them to be.

> Having it as a return value means that deserialize must be a template with the value type as its first template argument, which in turn means that overload sets won't work and type-specific serializers need to awkwardly define their fixed type as a template argument. Or we revert it to what it was before and generic serializers don't work.

I'm not sure I follow. If you can provide an instance of the type to be filled, you can provide the `typeof` of it. Do you have some (pseudo) code example of the issue ?

s-ludwig:
If the copy constructor is the issue, then why not just move the value instead of copying?

In your `deserializeFull` you make the assumption that there is a constructor that takes all fields in order as its arguments, even though that's the default this sounds like a dangerous assumption when any kind of type may be thrown at the serializer. I'm not sure whether that is a good compromise for an open set of serializable types. For a closed set, it would at least be a possible to define a `mixin` within each struct to make this safe.

Example:
```
int deserialize(R)(R range);
float deserialize(R)(R range);
T deserialize(T, R)(R range) if(is(T == struct));
```

How to invoke `deserialize` in each case alone or in a combination of those cases being defined?

Geod24:
> If the copy constructor is the issue, then why not just move the value instead of copying?

It breaks internal pointers. Which are invalid in D, but not in C++. And even in D, I know Weka is using them for example.

> In your `deserializeFull` you make the assumption that there is a constructor that takes all fields in order as its arguments, even though that's the default this sounds like a dangerous assumption when any kind of type may be thrown at the serializer.

Not only. Any type is able to explicitly support the serialization by specifying the static `fromBinary` methods, which will override the default behavior. It's true that there is no single assumption you can make that will hold for *all* types, so the contract here is that the default behavior is a good default, and we have a mean to override it if needed.

I would really have to have a way to provide a pattern matcher to the function, but I haven't found quite the right approach yet.

> How to invoke `deserialize` in each case alone or in a combination of those cases being defined?

The API I went with requires you to provide the type you want as template parameter. E.g. `auto x = deserializeFull!MyStruct(rng);`.

s-ludwig:
Serialization of possibly immutable C++ structs with internal pointers, set up in D, which may choose to move the struct implicitly, still sounds like a rather esoteric and dangerous combination to me (although I see how this could come up in custom IPC scenarios), and nothing to endorse. But since you have practical evidence I obviously can't really dismiss it.

So then that means it's template-only serialization, which, since it matches the general serialization API, is really not a big deal. Being able to work with concrete types and overload sets would have been nice, though.

Geod24 (approved):
LGTM. What's your timeline for a release ? We're going to try it out in our app very soon.

s-ludwig:
We can do a beta release immediately and I'll update the change log ASAP.

PR #2505, opened 2020-11-30
Optionally serializing ubyte[] as base64 string

ferencdg:
This can save brandwidth and decrease latency of REST queries

usage:
```D
@serPolicy!(Base64ArraySerPolicy)
interface ITestBase64
{
    @safe X getTest();
}
```

### DIFF
--- a/data/vibe/data/serialization.d
+++ b/data/vibe/data/serialization.d
@@ -132,6 +132,7 @@ import vibe.internal.meta.uda;
 import std.array : Appender, appender;
 import std.conv : to;
 import std.exception : enforce;
+import std.range.primitives : ElementType, isInputRange;
 import std.traits;
 import std.typetuple;
 
@@ -1499,6 +1500,22 @@ private template getExpandedFieldsData(T, FIELDS...)
 	enum subfieldsCount(alias F) = TypeTuple!(__traits(getMember, T, F)).length;
 	alias processSubfield(alias F) = aliasSeqOf!(zip(repeat(F), iota(subfieldsCount!F)));
 	alias getExpandedFieldsData = staticMap!(processSubfield, FIELDS);
+}
+
+/// Uses Base64 representation for `ubyte[]` instead of `to!string`
+public class Base64ArrayPolicy (R) if (isArray!R && is(ElementType!R : ubyte))
+{
+	public static string toRepresentation (in R data) @safe pure
+	{
+		import std.base64 : Base64;
+		return Base64.encode(data);
+	}
+
+	public static ubyte[] fromRepresentation (in string data) @safe pure
+	{
+		import std.base64 : Base64;
+		return Base64.decode(data);
+	}
 }
 
 /******************************************************************************/
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -554,7 +554,7 @@ struct ResultSerializer(alias ST, alias DT, string ContentType) {
 }
 
 
-package void defaultSerialize(T, R)(ref R output_range, in ref T value)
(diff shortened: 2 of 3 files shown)


ferencdg:
this is an other serialization change
@s-ludwig @Geod24 please let me know if anything should be changed

data/vibe/data/serialization.d
```diff
@@ -1487,6 +1488,22 @@ private template getExpandedFieldsData(T, FIELDS...)
...
+		import std.base64 : Base64;
+		return Base64.encode(data);
+	}
+
+	public static ubyte[] fromRepresentation (in string data) @safe pure
+	{
+		import std.base64 : Base64;
+		return Base64.decode(data);
+	}
+}
```
Geod24, line 1505: So, should we try writing to a sink / input range here ? :D

s-ludwig:
I wonder why the policy logic needs to be repeated in the web framework. Couldn't you just use the `@resultSerializer` attribute to specify a serializer that applies a certain policy?

data/vibe/data/serialization.d
ferencdg: I think if we want to have the policies support sink type serialization, then it should be a separate ticket. Polices were added to vibe.d long time ago.

ferencdg:
> I wonder why the policy logic needs to be repeated in the web framework. Couldn't you just use the `@resultSerializer` attribute to specify a serializer that applies a certain policy?
 
1. I wanted to have an easy way to specify the policy for ALL interface methods, so I declared it on the interface. ResultSerializers are attached to interface methods.
2. Result serializer only serializes the results, but when we define a policy on the interface it will affect the serialization of everything: result, query parameters, headers...

ferencdg:
> > I wonder why the policy logic needs to be repeated in the web framework. Couldn't you just use the `@resultSerializer` attribute to specify a serializer that applies a certain policy?
> 
> 1. I wanted to have an easy way to specify the policy for ALL interface methods, so I declared it on the interface. ResultSerializers are attached to interface methods, I would find it tedious to create REST interfaces and attach resultSerializer to all methods just to add a new policy.
> 2. Result serializer only serializes the results, but when we define a policy on the interface it will affect the serialization of everything: result, query parameters, headers...

@s-ludwig please let me know if you would like to see a different design

web/vibe/web/common.d
```diff
@@ -583,6 +584,88 @@ package template ResultSerializersT(alias func) {
...
+}
+
+///
+package struct SerPolicy (alias PolicyTemplatePar)
+{
+	alias PolicyTemplate = PolicyTemplatePar;
+}
+
+///
+public alias serPolicy (Args...) = SerPolicy!(Args);
```
Geod24, line 607:
```suggestion
public alias serializationPolicy (Args...) = SerializationPolicy!(Args);
```

It's better to be a little more verbose in the name for this public-facing method.

Geod24 (approved):
LGTM, just the one nit on the name, and I think this could use some documentation.

Geod24:
I did a few modifications:
- Fixed the documentation (indent was wrong, and the section name was wrong too, it should have been `Serialization_Policies:` instead of `Serialization Policies:` to count as a section name);
- Renamed `Base64ArraySerPolicy` to simply `Base64ArrayPolicy`, since it's already in the serialization package;
- Slightly rephrased the documentation and extended it to say it applies to *all* types;

@ferencdg : Sorry I missed it earlier, but why is this only on return values ? It should also apply to parameters. In fact, the issue was noticed on a parameter (when accepting an `SCPEnvelope`).

Geod24 (approved):

Issue #2506, opened 2020-12-07
HTTP request reusing TCP connection leads to wrong response being processed


ferencdg:
This issue is related to https://github.com/bpfkorea/agora/issues/1395#issuecomment-737626414

Here is how to reproduce:
```D
interface ServiceI
{
    @safe:
    
    StructA getTest ();
    StructA getTest2 ();
}
```
using the clients:

code to reproduce:

```D
setLogLevel(LogLevel.info);
auto settings = new HTTPServerSettings;
settings.port = 5000;
settings.bindAddresses = ["127.0.0.1"];
auto router = new URLRouter;
router.registerRestInterface(new Service);
auto server_addr = listenHTTP(settings, router);

runTask({
	scope (exit) exitEventLoop();
	auto settings = new RestInterfaceSettings();
	settings.httpClientSettings = new HTTPClientSettings();
	settings.httpClientSettings.readTimeout = 1.seconds;
	settings.baseURL = URL("http://127.0.0.1:5000/");
	auto service_client = new RestInterfaceClient!IService(settings);
	try {service_client.getTest(2000);} catch (Exception e) {}
	sleep(1200.msecs);
	auto result = service_client.getTest2(500);
	writeln("result:", result);
	assert(result.i == 500);
});

runApplication();
```

produces
```
[main(----) INF] Listening for requests on http://127.0.0.1:5000/
getTest receiving request with wait: 2000
getTest returning request with wait: 2000
result:TestStruct(2000)
core.exception.AssertError@source/tests.d(83): Assertion failure
----------------
??:? [0x5615a6884c05]
??:? [0x5615a68ad736]
??:? [0x5615a6890c3d]
??:? [0x5615a688479c]
/home/dan/dev/sandbox/sandbox-dlang/source/tests.d:83 [0x5615a662d4f1]
/home/dan/dev/sandbox/sandbox-dlang/../../../.dub/packages/vibe-core-1.11.1/vibe-core/source/vibe/core/task.d:712 [0x5615a666b3ac]
task.d:730 [0x5615a680a791]
task.d:439 [0x5615a6809edb]
??:? [0x5615a68869da]
[1]    171907 abort (core dumped)  ./build/app

```

The call to restClient.getTest2() might return the result that was intended for restClient.getTest() if the following happens:

1. restClient.getTest() uses a timeout of 3 seconds
2. the REST server is busy and only responds after the 3 seconds timeout
[...]

PR #2507, opened 2020-12-07
HTTP request reusing TCP connection leads to wrong response being processed

ferencdg:
fixes: #2506

### DIFF
--- a/http/vibe/http/client.d
+++ b/http/vibe/http/client.d
@@ -539,6 +539,11 @@ final class HTTPClient {
 		else
 			scope request_allocator = new RegionListAllocator!(shared(GCAllocator), true)(1024, GCAllocator.instance);
 
+		scope (failure) {
+			m_responding = false;
+			disconnect();
+		}
+
 		bool close_conn;
 		SysTime connected_time;
 		bool has_body = doRequestWithRetry(requester, false, close_conn, connected_time);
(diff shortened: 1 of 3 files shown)


Geod24:
Can you try to make the commit title a bit more descriptive ? "fixing major bug" isn't wrong, but it doesn't tell much to the reader, either :)
Additionally, is that something we could have a test for ?

ferencdg:
> Can you try to make the commit title a bit more descriptive ? "fixing major bug" isn't wrong, but it doesn't tell much to the reader, either :)
> Additionally, is that something we could have a test for ?

I can probably write a test code to reproduce it, but it will start up a rest server, so not sure if it is a good idea to make it a unit test.

ferencdg:
code to reproduce:

```D
	setLogLevel(LogLevel.info);
	settings.port = 5000;
	auto router = new URLRouter;
	router.registerRestInterface(new Service);
	auto server_addr = listenHTTP(settings, router);

	runTask({
		scope (exit) exitEventLoop();
		auto settings = new RestInterfaceSettings();
		settings.httpClientSettings = new HTTPClientSettings();
		settings.httpClientSettings.readTimeout = 1.seconds;
		settings.baseURL = URL("http://127.0.0.1:5000/");
		auto service_client = new RestInterfaceClient!IService(settings);
		try {service_client.getTest(2000);} catch (Exception e) {}
		sleep(1200.msecs);
		auto result = service_client.getTest2(500);
		writeln("result:", result);
		assert(result.i == 500);
	});

	runApplication();
```

PetarKirov:
Thanks for investigating and fixing this issue guys! That would surely save a lot of time for next one who bumps into this issue! 

> I can probably write a test code to reproduce it, but it will start up a rest server, so not sure if it is a good idea to make it a unit test.

@ferencdg in that case this folder seems like a good fit:
https://github.com/vibe-d/vibe.d/tree/master/tests

Geod24 (approved):

PR #2514, opened 2021-01-05
Handle Zlib inputs left hanging at the middle of a block

etcimon:
Handle inputs left hanging at the middle of a block

### DIFF
--- a/stream/vibe/stream/zlib.d
+++ b/stream/vibe/stream/zlib.d
@@ -371,6 +371,13 @@ class ZlibInputStream : InputStream {
 				m_in.read(m_inbuffer[0 .. clen]);
 				m_zstream.next_in = &m_inbuffer[0];
 				m_zstream.avail_in = cast(uint)clen;
+			} else {
+				import core.stdc.string : memmove;
+				memmove(m_inbuffer.ptr, m_zstream.next_in, m_zstream.avail_in);
+				auto clen = min(m_inbuffer.length - m_zstream.avail_in, m_in.leastSize);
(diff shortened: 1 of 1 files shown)


s-ludwig:
Wow, so this is really a case that needs to be handled? So many small fixes were necessary already here, because the zlib API is so brittle.

Based on the documentation, it seems like the input should always be fully drained, as long as enough output space is provided:

>  The detailed semantics are as follows. inflate performs one or both of the following actions:
>
>  -  Decompress more input starting at next_in and update next_in and avail_in accordingly. If not all input can be processed (because there is not enough room in the output buffer), then next_in and avail_in are updated accordingly, and processing will resume at this point for the next call of inflate().
>  -  Generate more output starting at next_out and update next_out and avail_out accordingly. inflate() provides as much output as possible, until there is no more input data or no more space in the output buffer (see below about the flush parameter).

...or at least not having enough output space is the only reason given. So if this indeed needs to be done anyway, I would suggest to only do it if there is still something in the input buffer *and* `inflate` did not write anything to the output buffer  - just to avoid adding unnecessary `memmove` overhead in the regular cases.

s-ludwig:
So I looked at this some more and tested with different kinds of inputs (different levels of compressibility), as well as input streams that only yielded one byte at a time. I couldn't make the existing code fail that way, unfortunately.

Do you remember where it failed for you? Did it fail an assertion (e.g. `assert(m_zstream.avail_out != m_outbuffer.peekDst.length || m_zstream.avail_in != avins)`), or did it hang in an infinite loop?

PR #2516, opened 2021-01-08
REST: Implement `viaBody`, `viaQuery`, `viaHeader` as parameters UDAs

Geod24:
Something I've been wanting for a while, should make user code much more friendly.
The names have changed to avoid weird overrides with the `headerParam` & co functions, but also to make function prototypes shorter.

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -370,27 +370,21 @@ interface Example6API
 {
 	@safe:
 
-	// The first parameter of @headerParam is the identifier (must match one of the parameter name).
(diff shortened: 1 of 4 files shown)


s-ludwig:
Looks good and definitely the way to go, I forgot about this when UDAs were made available for parameters.

The only thing I have would be to name this "viaX" instead of "byX", as it captures the bi-directional nature of the annotation and doesn't interfere with range constructors that are named like that.

Geod24:
Ah, I actually hesitated between `via` and `by`, but opted for the shorter name. Switched to `via`.

PR #2520, opened 2021-01-18
Do not use writefln in dtor

Geod24:
writefln could GC allocate, while printf is guaranteed not to.


s-ludwig:
The day that `writefln` in its basic form starts to allocate, a lot of other places will be broken, too.

Geod24:
Indeed, that's not the source of the `InvalidMemoryOperationError` we're seeing :(

PR #2525, opened 2021-01-21
Avoid `void[m_size]` in Json, as it leads to memory corruption in practice.

FeepingCreature:
This seems to only happen under load. We don't know exactly why it happens. The visible effect is a `Json` object containing a corrupted hashmap ("19 values, 4 buckets, all filled" or equivalent effects). My suspicion is there's somewhere in the compiler or Phobos that copies `void[m_size]` byte by byte, masking the pointer, leading to it being collected by the GC and reused.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -128,18 +128,24 @@ struct Json {
 		// memory leaks and, worse, std.algorithm.swap triggering an assertion
 		// because of internal pointers. This crude workaround seems to fix
 		// the issues.
-		enum m_size = max((BigInt.sizeof+(void*).sizeof), 2);
-		// NOTE : DMD 2.067.1 doesn't seem to init void[] correctly on its own.
-		// Explicity initializing it works around this issue. Using a void[]
-		// array here to guarantee that it's scanned by the GC.
-		void[m_size] m_data = (void[m_size]).init;
+		// NOTE: *Don't* use `void[m_size]`! It causes memory corruption and crashes.
+		// I suspect something in D copies `void[m_size]` byte by byte, causing the GC
+		// to temporarily lose track of the pointer.
+		// round up to the next (void*).sizeof multiple
+		enum m_size = [
+			BigInt.sizeof, long.sizeof, double.sizeof, bool.sizeof,
+			string.sizeof, (bool[string]).sizeof, (void[]).sizeof,
+		].fold!max.requiredNumberOfElementsOf((void*).sizeof);
+		// Using a void*[] array here to guarantee that it's both scanned by the
+		// GC and moved in pointer-sized steps.
+		void*[m_size] m_data;
 
 		static assert(m_data.offsetof == 0, "m_data must be the first struct member.");
 		static assert(BigInt.alignof <= 8, "Json struct alignment of 8 isn't sufficient to store BigInt.");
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Do you have any more information about the circumstances of the memory corruption (maybe specific to a certain compiler version, OS, or using the precise GC) in order to possibly trace down the cause? In contrast to GC memory leaks/false pointers, I never experienced this issue so far and *copying* a `Json` value byte-by-byte wouldn't quite explain it, since the source instance will always be complete.

Another question is how common the issue occurred for you, is it safe enough to say that the `void*` has actually fixed it?

FeepingCreature:
It's always been kinda random, and we've always only had one system that it's happened on, but we've gone from crashes maybe four times a day to no crashes for two and a half days, and no crashes for two days with a different solution. Associated versions are Linux x86_64, DMD 2.094.1, and Vibe.d 0.9.2 with notls. Conservative GC only. I'm sorry I can't narrow it down more, but the only repro we have is production. All I can tell you is that as far as I can tell, four hours between crashes to 50 without crashes, the problem is gone with this patch.

s-ludwig:
Okay, I see, doesn't sound realistic then to tackle it at a lower level...

The only issue then is that it runs into an error when used in CTFE. I currently don't see an obvious workaround for that, since pointers generally can't be bitcast into other types at CTFE (also not using a `union`). `if (__ctfe)` doesn't work at global scope either, so unless there is some other way to get around this, this may need to be made optional through a `version` block.

FeepingCreature:
Wait, pointers can't be arraycast but void[] can? That's odd... they should both fall down approximately for the same reason. Does DMD do a special handling for void[]?

s-ludwig:
I think bitcasts in general don't work during CTFE. In the `void[]` case, although it appears to do bitcasts, in reality it keeps the same internal type ("runtime compiletime type"). The current implementation only works as long as the type casts are symmetric, casting `int` to `void[]` and then casting back as `float` will result in an ICE/crash.

Now the annoying thing is that I don't remember whether it was just to combat false pointers, or if there was some memory corruption issue when I converted the code to use `void[]` instead of `union`. It might make sense to try `union` again and use that failing machine as the ultimate test (after testing locally and comparing memory usage).

Alternatively, as a quick fix and only within `if (__ctfe)`, going through `void[]` for a custom pseudo-bitcast should be an option: https://run.dlang.io/is/epOOiY

FeepingCreature:
`union` did work on the failing machine (that was that two-days-no-crashes run); I specifically switched to `void*[]` because it required fewer changes and so seemed more likely to be accepted. :) I can go back to union on the PR. At any rate, I don't understand how and why union *could* cause false pointers when void[] doesn't.

s-ludwig:
I think the issue was that the union's slack space did not get zero-initialized. However, when copying the union, all bytes were copied, duplicating any garbage or left-over pointers that were previously there. I'm not sure whether this was/is true for the stack and/or for the heap.

FeepingCreature:
Ah. Well then, should I make a PR that uses a union and just zero initializes the slack space?

s-ludwig:
Would be worth a shot, if everything works like that it would definitely be preferable.

FeepingCreature:
"Made" a PR. (2018... _sigh_)

s-ludwig:
Closing in favor of #2206, which was merged now.

PR #2529, opened 2021-01-29
Respect keep-alive on client requests

yazd:
`m_keepAliveLimit` must be compared to the current time.
`connected_time` is an `out` parameter that shouldn't be read before being initialized.

This bug was introduced in https://github.com/vibe-d/vibe.d/commit/560290a6bfc26a20e824a0c4808e32c4911b3f6c.


s-ludwig:
Thanks, that's a good catch! We should add a dedicated test for this at some point.

PR #2550, opened 2021-04-14
Change the client cert fail test to ignore the client-tunnel result

s-ludwig:
This obviously behaves differently based on the OpenSSL version used. Only the server tunnel result is relevant here.

See #2501

### DIFF
--- a/tests/tls/source/app.d
+++ b/tests/tls/source/app.d
@@ -27,14 +27,20 @@ void createPipePair(out Stream a, out Stream b)
 	b = createProxyStream(p2, p1);
 }
 
+enum Expected {
+	success,
+	fail,
+	dontCare
+}
+
 void testConn(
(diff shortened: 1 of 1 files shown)


Geod24:
```
+ dub build --combined -b release --compiler=dmd
Fetching vibe-core 1.16.0 (getting selected version)...
Failed to download package vibe-core from https://code.dlang.org/packages/vibe-core/1.16.0.zip
Error: Process completed with exit code 2.
```

Weird, dub v1.14 (I think?) seems broken now.

s-ludwig:
> Weird, dub v1.14 (I think?) seems broken now.

This has been the state ever since the last change to the test compiler versions, AFAIK. This is why I made the two tests non-required for the time being. TBH, I'm not really motivated to look into it and would rather let time/the next version bump fix this eventually.

PR #2551, opened 2021-04-15
Fix file example

### LINKED ISSUE #2547: file_operations example does not work for me
i get

    Main Thread: Writing small chunks in a yielding task (Task 1) for 150 ms
    Task 1: Starting write operations to hello.txt
    Task terminated with uncaught exception: Failed to write data to disk.
    Task 2: Task 1 has written 0B so far in 7 ms

instead of the expected output.
also the inner print function is not used.

gizmomogwai:
Debugging showed that the FileStream is destructed as soon as main is
exited (the task are running later). Interestingly, the bool can be
captured in the closure, but this seems not to work with the refcounted
FileStream. Passing the FileStream explicitely to the task as argument
solves that.

Unfortunately this is counter intuitive, as both bool and FileStream are
value types and we need to handle one differently from the other.

Fixes #2547

p.s. i also ran the app.d through dfmt (with defaults). If thats a problem
I am happy to do another patch without the formatting.

### DIFF
--- a/examples/file_operations/dub.json
+++ b/examples/file_operations/dub.json
@@ -4,5 +4,4 @@
 	"dependencies": {
 		"vibe-d:core": {"path": "../../"}
 	},
-	"versions": ["VibeDefaultMain"]
 }
--- a/examples/file_operations/source/app.d
+++ b/examples/file_operations/source/app.d
@@ -7,47 +7,40 @@ static if (__VERSION__ >= 2076)
 else
 	import std.datetime;
 
-auto peekMSecs(T)(T sw)
-{
-static if (__VERSION__ >= 2076)
-	return sw.peek.total!"msecs";
-else
-	return sw.peek.msecs;
+auto peekMSecs(T)(T sw) {
+	static if (__VERSION__ >= 2076)
+		return sw.peek.total!"msecs";
+	else
+		return sw.peek.msecs;
 }
 
-shared static this()
-{
+void main() {
 	FileStream fs = openFile("./hello.txt", FileMode.createTrunc);
 	StopWatch sw;
 	sw.start();
 	bool finished;
 	auto task = runTask({
-		void print(ulong pos, size_t sz) {
-			auto off = fs.tell();
-			fs.seek(pos);
-			ubyte[] dst = new ubyte[sz];
-			fs.read(dst);
-			writefln("%s", cast(string)dst);
-			fs.seek(off);
-		}
 		writeln("Task 1: Starting write operations to hello.txt");
-		while(sw.peekMSecs < 150) {
+		while (sw.peekMSecs < 150) {
 			fs.write("Some string is being written here ..");
 		}
-		writeln("Task 1: Final offset: ", fs.tell(), "B, file size: ", fs.size(), "B", ", total time: ", sw.peekMSecs, " ms");
+		writeln("Task 1: Final offset: ", fs.tell(), "B, file size: ",
+			fs.size(), "B", ", total time: ", sw.peekMSecs, " ms");
 		fs.close();
 		finished = true;
 	});
 
 	runTask({
-		while(!finished)
-		{
+		while (!finished) {
 			writeln("Task 2: Task 1 has written ", fs.size(), "B so far in ", sw.peekMSecs, " ms");
 			sleep(10.msecs);
 		}
 		removeFile("./hello.txt");
-		writeln("Task 2: Done. Press CTRL+C to exit.");
+		writeln("Task 2: Done.");
 		sw.stop();
+		exitEventLoop();
 	});
 	writeln("Main Thread: Writing small chunks in a yielding task (Task 1) for 150 ms");
+
+	runApplication();
 }


gizmomogwai:
@s-ludwig what do you think about the reference counting of the FileStream in the lambda, why is it not working as expected. Unfortunately I am not experienced enough in dlang to answer that.

s-ludwig:
The problem here is that the compiler builds a heap closure for the two task delegates, letting the GC deal with the destruction of the file stream. It's actually interesting that it doesn't complain, because it does complain with "`somevar` has scoped destruction. Cannot build closure", or similar, in other cases (I think when the variable is a function parameter instead of a local).

gizmomogwai:
hi @s-ludwig, thanks for the explanation ... i tried to print a stacktrace, when the filestream is destructed in case of the original program

```
 object.Exception@../../../../.dub/packages/vibe-core-1.16.0/vibe-core/source/vibe/core/file.d(659): destructing filestream
----------------
../../../../.dub/packages/vibe-core-1.16.0/vibe-core/source/vibe/core/file.d:659 @safe void vibe.core.file.FileStream.__dtor() [0x558f6cde14ff]
source/app.d:20 void app._sharedStaticCtor_L18_C1() [0x558f6cddaea8]
??:? void app.__modsharedctor() [0x558f6cddd4a4]
??:? void rt.minfo.runModuleFuncs!(rt.minfo.ModuleGroup.runCtors().__lambda2).runModuleFuncs(const(immutable(object.ModuleInfo)*)[]) [0x558f6cee2576]
??:? void rt.minfo.ModuleGroup.runCtors() [0x558f6cee228f]
??:? int rt.minfo.rt_moduleCtor().__foreachbody1(ref rt.sections_elf_shared.DSO) [0x558f6cea2709]
??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x558f6cea2ae4]
??:? rt_moduleCtor [0x558f6cea26d8]
??:? rt_init [0x558f6ce9c9aa]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x558f6ce9cfc3]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x558f6ce9cf5f]
??:? _d_run_main2 [0x558f6ce9cebb]
??:? _d_run_main [0x558f6ce9cc15]
/home/christiankoestlin/dlang/dmd-2.096.0/linux/bin64/../../src/druntime/import/core/internal/entrypoint.d:29 main [0x558f6cddd4d9]
??:? __libc_start_main [0x7f1231ea50b2]
??:? _start [0x558f6cddad1d]
Program exited with code -4
```
does this point somehow to the gc, or to the end of the `shared static this() {` scope? 

i am still confused where the `bool finished` is living in this case?

gizmomogwai:
you are for sure right about the error message, when i extract a method that takes the FileStream as parameter:
`source/app.d(19,6): Error: variable `app.test.fs` has scoped destruction, cannot build closure`.

s-ludwig:
Hm, okay, I misunderstood the part with `main` then. So it seems like the inverse is happening here. It *should* create a heap-closure but doesn't. This would then also explain the absence of the error message. I'll see if I can make a minimal test case and submit this to bugzilla.

For the sake of the example, instead of working around the issue, I would change the whole thing to use `main()` instead of `shared static this`/`VibeDefaultMain`, which is the idiomatic approach these days and avoids the limited scope. I can prepare a quick PR for that.

gizmomogwai:
hi @s-ludwig , after leeping over it I came to the same opinion, that a regular main would be better in this case. I do not really need it, but I think its good if the examples just work. What are your dfmt settings for the project?

s-ludwig:
I don't have settings for dfmt, as it wasn't able to support the [vibe.d style guide](https://vibed.org/style-guide) at the time I looked at it. I'll have another look at it.

s-ludwig:
Okay, so the main issue is that dfmt doesn't support K&R style braces, although it nowadays supports different styles. I'll see if I can make a quick PR to add support.

gizmomogwai:
I really like your project setup ... an answer for everything .. even nicely documented ....
I think even if dfmt would support K&R braces, the change in the complete codebase would be too big!

gizmomogwai:
mhh ... I did not see that coming with the #Fixes github comment...

gizmomogwai:
btw. formatted with `dfmt --brace_style otbs`.

s-ludwig:
Yeah, so there will be at least one issue left with K&R support merged (https://github.com/dlang-community/dfmt/pull/534), which is that I quite often have on-liner methods for simple getters that fit in a single line. `dfmt` currently has no option to leave that intact or even format it that way by itself. But at least for examples or new modules it would be nice to have it as an option or starting point, once/if that PR gets merged, I'd add the configuration to .editorconfig.

gizmomogwai:
there is this // dfmt off / on thing, but that would make the whole point of the one liner void :)

gizmomogwai:
its amazing, how few lines you needed to change in the formatter itself, and how big the total commit is with all the tests ...

gizmomogwai:
What do we do about the file example?

s-ludwig:
Can you also remove the "Press CTRL+C to exit." part of the last message? Otherwise LGTM!

BTW, I couldn't reproduce the issue manually in a minimal case. Looks like this would need some Dustmite based code reduction instead.

gizmomogwai:
Had that already prepared, but not pushed :/

s-ludwig:
Okay, thanks! Setting to auto-merge.

PR #2552, opened 2021-04-19
Allow `const` parameters in REST generator

omerfirmak:
This also allows `in` parameters to be used with
`--preview=in`.

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -460,6 +460,90 @@ class Example7 : Example7API {
 	}
 }
 
+/* --------- EXAMPLE 8 ---------- */
+
+@rootPathFromName
+interface Example8API
+{
+	// Methods need to be `@safe`:
+	@safe:
+
+	struct FooType {
+		int a;
+		string s;
+		double d;
+	}
+
+	FooType constFoo (const FooType param);
+	FooType constRefFoo (const ref FooType param);
+	FooType inFoo (in FooType param);
+	FooType immutableFoo (immutable FooType param);
+	int[] constArr (const int[] param);
+	int[] constRefArr (const ref int[] param);
+	int[] inArr (in int[] param);
+	int[] immutableArr (immutable int[] param);
+}
+
+class Example8 : Example8API
+{
+	override: // usage of this handy D feature is highly recommended
+
+	FooType constFoo (const FooType param)
+	{
+		return param;
+	}
+
+	FooType constRefFoo (const ref FooType param)
+	{
+		return param;
+	}
+
+	FooType immutableFoo (immutable FooType param)
+	{
+		return param;
+	}
+
+	int[] constArr (const int[] param)
+	{
+		return param.dup;
+	}
+
+	int[] constRefArr (const ref int[] param)
+	{
+		return param.dup;
+	}
+
+	int[] immutableArr (immutable int[] param)
+	{
+		return param.dup;
+	}
+
+	int[] inArr (in int[] param)
+	{
+		return param.dup;
+	}
+
+	FooType inFoo (in FooType param)
+	{
+		return param;
+	}
+}
+
+unittest
+{
+	auto router = new URLRouter;
+	registerRestInterface(router, new Example8());
+	auto routes = router.getAllRoutes();
+
+	assert (routes[0].method == HTTPMethod.POST && routes[0].pattern == "/example8_api/const_foo");
+	assert (routes[1].method == HTTPMethod.POST && routes[1].pattern == "/example8_api/const_ref_foo");
+	assert (routes[2].method == HTTPMethod.POST && routes[2].pattern == "/example8_api/in_foo");
+	assert (routes[3].method == HTTPMethod.POST && routes[3].pattern == "/example8_api/immutable_foo");
+	assert (routes[4].method == HTTPMethod.POST && routes[4].pattern == "/example8_api/const_arr");
+	assert (routes[5].method == HTTPMethod.POST && routes[5].pattern == "/example8_api/const_ref_arr");
+	assert (routes[6].method == HTTPMethod.POST && routes[6].pattern == "/example8_api/in_arr");
+	assert (routes[7].method == HTTPMethod.POST && routes[7].pattern == "/example8_api/immutable_arr");
+}
 
 void runTests(string url)
 {
@@ -601,6 +685,24 @@ void runTests(string url)
 			assert(answer == expected);
 		}
 	}
+
+	// Example 8
+	{
+		import std.algorithm;
+
+		auto api = new RestInterfaceClient!Example8API(url);
+		Example8API.FooType foo = Example8API.FooType(44, "firmak", 0.37);
+		assert(foo == api.constFoo(foo));
+		assert(foo == api.constRefFoo(foo));
+		assert(foo == api.inFoo(foo));
+		assert(foo == api.immutableFoo(foo));
+
+		int[] arr = [42, 37, 44];
+		assert(arr.equal(api.constArr(arr)));
+		assert(arr.equal(api.constRefArr(arr)));
+		assert(arr.equal(api.immutableArr(cast(immutable(int[])) arr)));
+		assert(arr.equal(api.inArr(arr)));
+	}
 }
 
 shared static this()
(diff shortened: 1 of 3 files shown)


Geod24:
Casting `const` away is not the right approach. Rather, mutability should not be required.
To do this, the value have to be constructed in one go, not piece-meal. That's for example why our deserializer does this:
https://github.com/bosagora/serialization/blob/8f1e4c124e7c909e7b65ffea39e996fc116ba181/source/agora/serialization/Serializer.d#L691-L699

The `staticMap` means we end up calling the constructor, and give it function calls as parameter.
Another "nice" trick is how we deal with static arrays: https://github.com/bosagora/serialization/blob/8f1e4c124e7c909e7b65ffea39e996fc116ba181/source/agora/serialization/Serializer.d#L609-L617

Nowhere in the serializer do we have away `const`ness (we do have `cast`, but they never remove `const`). While `cast`ing away constness is sometimes okay, modifying the underlying data (which this does) never is.

omerfirmak:
Updated

web/vibe/web/rest.d
```diff
@@ -1603,8 +1604,8 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
 		{
 			handleCors();
 			foreach (i, P; PTypes) {
-				static if (sroute.parameters[i].isOut) {
-					static assert (sroute.parameters[i].kind == ParameterKind.header);
+				static if (sroute.parameters[i].isOut &&
+					sroute.parameters[i].kind == ParameterKind.header) {
```
Geod24, line 1608: Why ?

Geod24:
Test plz

web/vibe/web/rest.d
omerfirmak: https://github.com/vibe-d/vibe.d/blob/f1554380fefa186298d930eaeccf6f439924ff47/web/vibe/web/internal/rest/common.d#L298-L306

`in`parameters that are passed by reference are marked as `isOut`and end up in the body not header.

web/vibe/web/rest.d
```diff
@@ -1447,7 +1448,7 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
 	import vibe.textfilter.urlencode : urlDecode;
 
 	enum Method = __traits(identifier, Func);
-	alias PTypes = ParameterTypeTuple!Func;
+	alias PTypes = staticMap!(Unqual, ParameterTypeTuple!Func);
```
Geod24, line 1451: This should come with a comment. From my understanding:

```suggestion
    // We need mutable types for deserialization
	alias PTypes = staticMap!(Unqual, ParameterTypeTuple!Func);
```

However there should probably be a check for immutable or the error messages aren't going to be great.

omerfirmak: immutable params compile fine tho?

web/vibe/web/rest.d
Geod24: Maybe fix that bug then ? :D

web/vibe/web/rest.d
Geod24: So, mutable does not implicitly converts to its `immutable` counterpart, unless the type has no indirection.
Which means that, by deserializing things as `T`, we can't later bind it to a parameter of type `immutable(T)`.
And currently, IIUC, the error is that we try to modify an `immutable(T)` during deserialization.
If it's not one, it has to be the other. If it's neither, two possibilities:
- We are deserializing in one go and we shouldn't have the error to begin with (or the fix is even more trivial?);
- There is a cast somewhere that breaks the type system;

omerfirmak:
> Test plz

Added

web/vibe/web/rest.d
omerfirmak: I am not sure if that can be fixed easily. `const ref` and `in` are indistinguishable by looking at their storage class.

omerfirmak: But since REST only allows `const ref` but not `ref` they cant really be outputs actually.

https://github.com/vibe-d/vibe.d/pull/2495

Geod24: > are indistinguishable by looking at their storage class.

They are, using `ParameterStorageClass`. But it requires a new-ish release (2.094 IIRC)

tests/rest/source/app.d
```diff
@@ -460,6 +460,64 @@ class Example7 : Example7API {
...
+	FooType constFoo (const FooType param);
+
+	FooType constRefFoo (const ref FooType param);
+
+	FooType inFoo (in FooType param);
+
+	FooType immutableFoo (immutable FooType param);
+	struct FooType {
+		int a;
+		string s;
```
Geod24, line 480: Now try with `int[]`

omerfirmak: Done but this will probably have to wait for the next compiler release to work.

Geod24:
```
web-framework-i18n-example ~master: building configuration "application"...
Compiling Diet HTML template home.dt...
../../web/vibe/web/i18n.d(222,4): Error: `"Missing translation key for en_US; home: This is the internationalization example app."`
../../web/vibe/web/i18n.d(175,24):        called from here: `tr(key, null, 0, context)`
../../web/vibe/web/web.d(340,99):        called from here: `tr(key, context)`
/home/runner/.dub/packages/diet-ng-1.8.0/diet-ng/source/diet/traits.d(75,23):        called from here: `translate(text, context)`
/home/runner/.dub/packages/diet-ng-1.8.0/diet-ng/source/diet/parser.d(1239,17):        called from here: `translate(kln, stripExtension(baseName(loc.file)))`
/home/runner/.dub/packages/diet-ng-1.8.0/diet-ng/source/diet/parser.d(1110,20):        called from here: `parseTextLine(remainder, ret, tmploc)`
/home/runner/.dub/packages/diet-ng-1.8.0/diet-ng/source/diet/parser.d(1044,38):        called from here: `parseTagLine(input, loc, has_nested)`
/home/runner/.dub/packages/diet-ng-1.8.0/diet-ng/source/diet/parser.d(77,76):        called from here: `parseDietRaw(f)`
```

@s-ludwig : Looks like our CI is broken ?

Geod24:
Indeed, it's also triggering in Buildkite: https://github.com/dlang/dmd/pull/12526#issuecomment-888582301

PetarKirov:
@Geod24 I think the cause may have been this PR: https://github.com/rejectedsoftware/diet-ng/pull/92

Geod24:
@omerfirmak : Please rebase

omerfirmak:
Rebased

Geod24:
```
[INFO] Running test rest
+ cd tests/rest
+ dub --compiler=dmd --build-mode=separate
Running pre-generate commands for vibe-d:tls...
Performing "debug" build using dmd for x86_64.
[...]
tests ~master: building configuration "application"...
../../web/vibe/web/rest.d(1609,6): Error: static assert:  `sroute.parameters[0].kind == ParameterKind.header` is false
../../web/vibe/web/rest.d(442,45):        instantiated from here: `jsonMethodHandler!(constRefFoo, 1LU, Example8)`
source/app.d(721,23):        instantiated from here: `registerRestInterface!(Example8)`
dmd failed with exit code 1.
```

omerfirmak:
What, https://github.com/dlang/dmd/pull/12458/commits/23333399783b5d8deb9edcb518c562c91404b833 should've fixed it.

Geod24:
https://github.com/dlang/dmd/commit/03c30609389c5dce7e22ce39e8a176081e3e9c10

Only for v2.096.1 and above (LDC 1.26). So you need to version out the test for anything less than that (we test with 10 versions of DMD). But it also fails on dmd-latest so no luck here.

web/vibe/web/internal/rest/common.d
```diff
@@ -300,7 +301,7 @@ import std.traits : hasUDA;
 					pi.isOut = true;
 				} else static if (SC & ParameterStorageClass.ref_) {
 					pi.isIn = true;
-					pi.isOut = true;
+					pi.isOut = isMutable!PT;
```
omerfirmak, line 304: Let's see if this will help. Works fine on LDC 1.27.0.

Geod24: Hold on, we use `inIn` for "input parameter", not the `in` storage class, right ? Damn that's confusing :D

PR #2571, opened 2021-06-02
Avoid using closures in REST router

omerfirmak:
This came up while trying to add explicit dtors to one of our API arguments.
Removing trusted dgs didn't cause any problems, for our application at least.

### DIFF
--- a/examples/rest-collections/source/api.d
+++ b/examples/rest-collections/source/api.d
@@ -5,6 +5,7 @@ import vibe.web.rest;
 
 interface ForumAPI {
 	// base path /threads/
+@safe:
 	Collection!ThreadAPI threads();
 }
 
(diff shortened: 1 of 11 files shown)


Geod24:
Both allocate ? They really should not.

omerfirmak:
> Both allocate ? They really should not.

Seems like only the first one does. Will update.

omerfirmak:
CI is not happy at all.

Geod24:
Looks like some example are using the (long deprecated) `@system` API methods and the trusted lambda worked around that.

omerfirmak:
> Looks like some example are using the (long deprecated) `@system` API methods and the trusted lambda worked around that.

I think it is a good time to end the deprecation period, what do you think?

Geod24:
Yes

omerfirmak:
Updated

omerfirmak:
Updated again, turns out I missed a bunch of non-safe API.

PR #2576, opened 2021-06-11
CI: Update supported compiler

Geod24:
Change 'every 5 releases' to 'last 4 releases then every 5 releases',
as it is most likely that users are stuck one or two version back.
Currently assumes that ldc-latest is v1.27.0, which is yet to be released,
but should be soon.


Geod24:
@s-ludwig : You'll have to change branch protections tho.

s-ludwig:
Removed the outdated required checks, but I'll wait with the merge until all checks have passed.

PR #2581, opened 2021-07-13
Four new functions for the Json type

burner:
In my day work this type of access pattern is all over the code base.
Therefore, I thought this might be useful for other people as well.

I think the new example at the top shows how this is intended to be
used.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -65,6 +65,21 @@ module vibe.data.json;
 	string j4 = S("foo", 32, true).serializeToJsonString();
 }
 
+/// Range based Nullable based access
+@safe unittest {
+	// using the constructor
+	Json input = Json.emptyObject;
+	input["intArray"] = Json([Json("foo"), Json(42), Json(24)]);
+
+	long theSum = input.asArray("intArray")
+		.map!(it => it.as!long())
+		.filter!(it => !it.isNull)
+		.map!(it => it.get())
+		.sum;
+
+	assert(theSum == 66);
+}
+
 
 public import vibe.data.serialization;
 
@@ -80,7 +95,7 @@ import std.json : JSONValue, JSONType;
 import std.range;
 import std.string;
 import std.traits;
-import std.typecons : Tuple;
+import std.typecons : nullable, Tuple, Nullable;
 import std.uuid;
 
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


Geod24:
I think that `std_data_json` has `opt` for this ? But yeah, the `as` thing is something we do as well.

burner:
@Geod24 does that mean I should rename the function, xor do you like them the way they are?

Geod24:
I like the names the way they are. The `opt` thing was a different approach for you to consider.
See [this code](https://github.com/dlang-community/std_data_json/blob/113bec1a37451d1dc3c33ecafc51eb6f440dc981/source/stdx/data/json/value.d#L232-L239).
Essentially it uses `opDispatch` to allow any kind of access. The [unittest](https://github.com/dlang-community/std_data_json/blob/113bec1a37451d1dc3c33ecafc51eb6f440dc981/source/stdx/data/json/value.d#L414) has some examples.

s-ludwig:
I've got to say that the value compared to the existing `Json.opt()` (not the one from std-data-json) is rather limited:

```
value.asArray("a")
value["a"].opt!(Json[])

value.as!long("a")
value["a"].opt!long
```

In the array case the advantage is higher due to the amount of punctuation. To combat that, we could also introduce something like this, to be more in line with `opt`: `value["a"].optArray`

burner:
I would argue the other way round. The non array version gives you more information due to the Nullable being returned.
Returning T.init does not make it clear if there was a value or not.

s-ludwig:
There are of course many situations in which it doesn't matter whether the value is there or not, in which case the `Nullable` approach means a higher syntactical overhead due to the additional null check. The `Nullable` approach also loses the information about whether a value doesn't exist, or has the wrong type.

```
// current approach
auto a = value["a"];
if (a.type == Json.Type.string)
    foo(a.get!string);

// Nullable approach
auto a = value.as!string("a")
if (!a.isNull)
    foo(a.get);

// throws on type error
if (auto a = "a" in value)
    foo(a.get!string);
```

I don't really see the advantage here and would usually prefer the third, and shortest, approach in order to not hide input format errors (well, actually I'd use `deserializeJson` if possible, but that's a different matter). This is the example adjusted to those approaches:

```
// current approach
input["intArray"].opt!(Json[])
	.filter!(it => it.type == Json.Type.long)
	.map!(it => it.get!long())
	.sum;

// Nullable approach
input.asArray("intArray")
	.map!(it => it.as!long())
	.filter!(it => !it.isNull)
	.map!(it => it.get())
	.sum;

// throws on type error
input["intArray"].opt!(Json[])
	.map!(it => it.get!long())
	.sum;
```

burner:
I think we are talking past each other.

```D
// current approach
auto a = value["a"];
if (a.type == Json.Type.string)
    foo(a.get!string);
```

will throw if value is not an object the Nullable approach will not.

s-ludwig:
Okay, I didn't realize that that was what you were after. May I ask where you'd apply this kind of (overly?) permissive parsing? I'd at least imagine some kind of diagnostic output in real-world examples.

Adjusted example:
```
auto a = "a" in value.opt!(Json[string]);
if (a && a.type == Json.Type.string)
    foo(a.get!string);
```

BTW, it could be argued to simply change the semantics of `opIndex` to simply return `undefined` instead of throwing an exception, in case the value is not an object, since this is modeled after JavaScript anyway (which I wouldn't repeat nowadays, see std_data_json, but again, that's a different matter).

burner:
I use that in our internal graphqld based app. 

checkType is the one that throws, and that is all over json.d. I looked into changing that, but that seemed to be too big of a change.

s-ludwig:
Silent semantic changes are of course always an issue, but in this case it's hard for me to imagine realistic cases where this would introduce bugs. I think we should at least seriously consider this, before resorting to introducing overlapping APIs.

burner:
fair enough, your project your call. I'll close this and take a look at changing checkType

PR #2586, opened 2021-08-05
Meson: fixes using vibe.d as subproject

denizzzka:
Fixes error of using vibe.d as subproject:

subprojects/vibe.d-0.9.3/meson.build:37:4: ERROR: Function 'add_global_arguments' cannot be used in subprojects because there is no way to make that reliable.
Please only call this if is_subproject() returns false. Alternatively, define a variable that
contains your language-specific arguments and add it to the appropriate *_args kwarg in each target.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,15 +35,13 @@ endif
 
 if meson.is_subproject() == false
     if meson.get_compiler('d').get_id() == 'llvm'
-        add_global_arguments(['-d-version=VibeLibeventDriver',
-                              '-d-version=Have_openssl',
+        add_global_arguments(['-d-version=Have_openssl',
                               '-d-version=Have_diet_ng',
                               '-d-version=Have_stdx_allocator',
                               flag_new_openssl_ldc], language : 'd')
     endif
     if meson.get_compiler('d').get_id() == 'dmd'
-        add_global_arguments(['-version=VibeLibeventDriver',
-                              '-version=Have_openssl',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -33,20 +33,23 @@ if ssl_dep.version().version_compare('>=1.1')
...
-endif
-if meson.get_compiler('d').get_id() == 'dmd'
-    add_global_arguments(['-version=VibeLibeventDriver',
-                          '-version=Have_openssl',
-                          '-version=Have_diet_ng',
-                          '-version=Have_stdx_allocator',
-                          flag_new_openssl_dmd], language : 'd')
+if meson.is_subproject() == false
+    if meson.get_compiler('d').get_id() == 'llvm'
+        add_global_arguments(['-d-version=VibeLibeventDriver',
```
s-ludwig, line 38: `VibeLibeventDriver` has no effect these days, it's always eventcore.

s-ludwig:
I can't really review this properly due to lack of Meson experience, but I guess that this also needs these two to be merged in some form:
- https://github.com/vibe-d/eventcore/pull/171
- https://github.com/vibe-d/vibe-core/pull/254

denizzzka:
@s-ludwig both of these is needed, and this PR too

s-ludwig:
Okay. I mean, nothing can break, so I can just as well merge them in their current state.

denizzzka:
> nothing can break

...because it is already broken at least 3 years :-)

Issue #2588, opened 2021-08-05
[Meson] vibe.d/core/meson.build points to empty dir


denizzzka:
`vibe.d/core/meson.build` file points to files what been moved into `vibe-core` repository

denizzzka:
Related PR: https://github.com/vibe-d/vibe-core/pull/254

PR #2591, opened 2021-08-06
[Meson] Removes meson.build pointing to deleted "core" sources

denizzzka:
#2588 fix


denizzzka:
Looks like `vibe.d/core/dub.sdl` also can be removed?

PR #2594, opened 2021-08-13
Remove OpenSSL 0.9 support, correct README about default OpenSSL

Geod24:
See each commit, those should be fairly straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,15 +178,15 @@ Optionally, run `./setup-freebsd.sh` to create a user/group pair for privilege l
 Switching between OpenSSL versions
 ----------------------------------
 
-By default, vibe.d is built against OpenSSL 1.0.x. On systems that use the newer
-1.1.x branch, this can be overridden on the DUB command line using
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -178,15 +178,15 @@ Optionally, run `./setup-freebsd.sh` to create a user/group pair for privilege l
 Switching between OpenSSL versions
 ----------------------------------
 
-By default, vibe.d is built against OpenSSL 1.0.x. On systems that use the newer
-1.1.x branch, this can be overridden on the DUB command line using
-`--override-config vibe-d:tls/openssl-1.1`. Alternatively, the same can be done
+By default, vibe.d is built against OpenSSL 1.1.x. On systems that use the newer
```
s-ludwig, line 181:
```suggestion
By default, vibe.d is built against OpenSSL 1.1.x. On systems that use the older
```

Geod24: Fixed

PR #2595, opened 2021-08-16
Fix language matching for basic language specifiers.

s-ludwig:
Accepting "en" should match "en_US", not just as a fallback, like it's the case with accepting "en_GB" and providing "en_US" instead.

This required fixing two existing bad test cases.

### DIFF
--- a/web/vibe/web/i18n.d
+++ b/web/vibe/web/i18n.d
@@ -271,6 +271,9 @@ public string determineLanguageByHeader(T)(string accept_language, T allowed_lan
 				// request en_* == serve en
 				if (lcode == alang && !lextra.length)
 					return lang;
+				// request en == serve en_*
+				if (lcode == alang && !aextra.length)
+					return lang;
 				// request en* == serve en_* && be first occurence
 				if (lcode == alang && lextra.length && !fallback.length)
 					fallback = lang;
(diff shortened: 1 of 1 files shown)


s-ludwig:
@WebFreak001, note the adjusted test cases. This should be the correct behavior, according to what I've read in the specs.

web/vibe/web/i18n.d
```diff
@@ -303,19 +306,20 @@ public string determineLanguageByHeader(Tuple...)(HTTPServerRequest req, Tuple a
 }
 
 @safe unittest {
-	assert(determineLanguageByHeader("de,de-DE;q=0.8,en;q=0.6,en-US;q=0.4", ["en-US", "de_DE", "de_CH"]) == "de_DE");
-	assert(determineLanguageByHeader("de,de-CH;q=0.8,en;q=0.6,en-US;q=0.4", ["en_US", "de_DE", "de-CH"]) == "de-CH");
+	assert(determineLanguageByHeader("de-DE,de;q=0.8,en;q=0.6,en-US;q=0.4", ["en-US", "de_DE", "de_CH"]) == "de_DE");
+	assert(determineLanguageByHeader("de-CH,de;q=0.8,en;q=0.6,en-US;q=0.4", ["en_US", "de_DE", "de-CH"]) == "de-CH");
```
WebFreak001, line 310: old test cases should be kept, those were what is sent by Firefox.

WebFreak001 (approved):
generally good change, but old tests should stay

web/vibe/web/i18n.d
s-ludwig: The thing just is that I don't see how de (implicitly q=1.0), de-CH (q=0.8) would match the de-CH instead of the de, according to the descriptions of how the matching works. So I think that the test is wrong in assuming that the country specific language is matched.

BTW, from what I've seen, Firefox sends `de,en-US;q=0.7,en;q=0.3` (or ending with q=0.6, as in the test that I've added) by default for a German installation, could those other variants be from a manual change of language settings?

WebFreak001: unsure where I got the headers from at that point, it might have been old versions using that or just modified/extended to contain more languages.

For The quality thing: The provided languages are de-DE and de-CH, in this test we don't offer (second argument) a generic de translation. The user says "give me any de" or "give me de-CH", as we have an exact match for de-CH at quality 0.8 and just a language (not region) match for de at quality 1 with de-DE I think's the correct choice choosing that, given some sufficient quality. The quality number is kind of arbitrary by the UA and I don't know of any web servers which take it into account, so I think this works sufficiently.

For example we could say that matching de-DE on de would give it quality 0.8, but that would be an arbitrary hardcoded value. For someone in region Switzerland I would expect the browser to send "give me any German, preferably Swiss German" and `de,de-CH;q=0.8` feels valid to me for that and thus sending de-CH even if we have other DEs (but not just `de`) seems to be the right thing.

s-ludwig: But there is no description that I found that says how the weights (which we don't even parse) should affect the selection other than defining a priority order. All filtering/selection approaches are only based on the (literal or quality based) order of the entries. The idea for the client is to order the accepted languages in a way that the most specific preference comes first.

WebFreak001: Yeah I think the problem with the `de,de-CH;q=0.8` string is that if we only have `de-DE,de-CH` translations both of those would match `de` (we shouldn't hardcode de-DE to be a default `de`, other languages have different spelling of regions and languages) so which one should we pick? I think as the browser sends de-CH explicitly, we should use that one.

Of course if we now had `de,de-AT` and our languages are de-DE and de-CH, which one do we pick? I think the app dev should probably provide a generic `de` translation and only add region specific strings to the region specific translation, which would clear up this ambiguity.

s-ludwig: The point is that this would in theory be desirable, but looking at the related RFCs, there is no such attempt made. The list is simply an ordered list of possible matches.

> both of those would match de (we shouldn't hardcode de-DE to be a default de, other languages have different spelling of regions and languages)

The current solution is to match the first "de" language in the list of available languages. That way the website author is free to set the defaults. Same goes for choosing the first available language in case there is no match at all, instead of hard-coding "en_US".

PR #2602, opened 2021-09-01
Add support for InputStreamProxy REST return type

s-ludwig:
Allows to process large response bodies without loading everything into RAM.

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -7,6 +7,7 @@
 import vibe.appmain;
 import vibe.core.core;
 import vibe.core.log;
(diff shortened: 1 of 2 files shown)


Geod24:
Any chance that we could recognize https://dlang.org/phobos/std_range_interfaces.html too ?
I've been wanting to return ranges from my interfaces since forever.

s-ludwig:
@Geod24: That should be relatively easy with a wrapped `StreamInputRange` and also seems to me like a reasonable extension. I'd just leave that for a separate PR.

Geod24 (approved):
Fairly straightforward, LGTM

PR #2607, opened 2021-09-15
Fix bogus diagnostic message for unapplied SSL option

s-ludwig:
See #2589.


s-ludwig:
@Geod24 Notice the `(newopts & options) != options` - regardless of `SSL_OP_NO_COMPRESSION` I had to do that due to default options as soon as the correct `SSL_CTX_set_options` was used.

Geod24:
Even better, thanks !

PR #2609, opened 2021-09-21
Fix secret length mismatch at digest auth

kubo39:
This fixes https://github.com/vibe-d/vibe.d/issues/2597.
It has been broken after https://github.com/vibe-d/vibe.d/commit/0046c48bd8d85fe68fd07b1fa94ddabaf57e2f13 was committed.
This also fixes the expiration logic bug.

### DIFF
--- a/http/vibe/http/auth/digest_auth.d
+++ b/http/vibe/http/auth/digest_auth.d
@@ -56,7 +56,7 @@ class DigestAuthInfo
 		if (decoded.length != now.sizeof + secret.length) return NonceState.Invalid;
 		auto timebytes = decoded[0 .. now.sizeof];
 		auto time = () @trusted { return (cast(typeof(now)[])timebytes)[0]; } ();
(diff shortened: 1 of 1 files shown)


Geod24:
Can you add a test to prevent a future regression ?

kubo39:
@Geod24 test added.

http/vibe/http/auth/digest_auth.d
```diff
@@ -56,7 +56,7 @@ class DigestAuthInfo
 		if (decoded.length != now.sizeof + secret.length) return NonceState.Invalid;
 		auto timebytes = decoded[0 .. now.sizeof];
 		auto time = () @trusted { return (cast(typeof(now)[])timebytes)[0]; } ();
-		if (timeout + time > now) return NonceState.Expired;
+		if (timeout + time < now) return NonceState.Expired;
```
Geod24, line 59: Looks like a second fix, which could itself use a test, but let's not delay this any longer.

PR #2613, opened 2021-09-29
300 hecto-nanoseconds is too strict

kubo39:
This commit changes digest auth timeout interval: 5 minutes.

### DIFF
--- a/http/vibe/http/auth/digest_auth.d
+++ b/http/vibe/http/auth/digest_auth.d
@@ -30,12 +30,12 @@ class DigestAuthInfo
 
 	string realm;
 	ubyte[16] secret;
-	ulong timeout;
+	Duration timeout;
 
 	this()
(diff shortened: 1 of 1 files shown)


Geod24:
I think it'd be better to either:
- Make `timeout` a `Duration`;
- Make the code using `timeout` call `.seconds`, as it seems to be the intended resolution;

Instead of storing hnsecs in `timeout`, which would be surprising to most people IMO.

Geod24:
Since this is (technically) a breaking change, paging @s-ludwig .
I think it's one of those changes that make sense (considering the class is broken to the point of being unusable in a real-world scenario IIRC) so IMO it's acceptable.

s-ludwig:
Agreed, makes sense especially because obviously nobody has used or could use digest auth productively.

PR #2616, opened 2021-10-04
AWS DocumentDB support through OP_MSG and handling different OK type

jmt-lab:
I wouldn't suggest merging this directly as is, as it could probably break compatibility with older mongodb <3.6 i believe is the cutoff here. But I thought i'd share the changes here as this was what I had to do to get vibe-d mongo driver to work with AWS DocumentDB. This is because even though docdb is mongo compliant it only has implemented OP_MSG for the operations i modified below. Also docdb doesn't support $external.$cmd for handshaking and so with this fork you MUST specify a database in the client settings so that it can auth against it.

I hope this can help you find a cleaner way to integrate these changes where they won't break non-docdb use cases. This was a quick and dirty way of getting compatibility working.

### DIFF
--- a/mongodb/vibe/db/mongo/connection.d
+++ b/mongodb/vibe/db/mongo/connection.d
@@ -17,7 +17,7 @@ import vibe.db.mongo.flags;
 import vibe.inet.webform;
 import vibe.stream.tls;
 
-import std.algorithm : map, splitter;
+import std.algorithm : map, splitter, canFind;
 import std.array;
 import std.conv;
 import std.digest.md;
@@ -208,6 +208,7 @@ final class MongoConnection {
 			"os": Bson(["type": Bson(os.os.to!string), "architecture": Bson(hostArchitecture)]),
 			"platform": Bson(platform)
 		]);
+		string cn = (m_settings.database == string.init ? "admin" : m_settings.database) ~ ".$cmd";
 
 		if (m_settings.appName.length) {
 			enforce!MongoAuthException(m_settings.appName.length <= 128,
@@ -215,13 +216,14 @@ final class MongoConnection {
 			handshake["client"]["application"] = Bson(["name": Bson(m_settings.appName)]);
 		}
 
-		query!Bson("$external.$cmd", QueryFlags.none, 0, -1, handshake, Bson(null),
+		query!Bson(cn, QueryFlags.none, 0, -1, handshake, Bson(null),
 			(cursor, flags, first_doc, num_docs) {
 				enforce!MongoDriverException(!(flags & ReplyFlags.QueryFailure) && num_docs == 1,
 					"Authentication handshake failed.");
 			},
 			(idx, ref doc) {
-				enforce!MongoAuthException(doc["ok"].get!double == 1.0, "Authentication failed.");
+				auto flag = doc["ok"].type == Bson.Type.int_ ? cast(double)doc["ok"].get!int : doc["ok"].get!double;
+				enforce!MongoAuthException(flag == 1.0, "Authentication failed.");
 				m_description = deserializeBson!ServerDescription(doc);
 			});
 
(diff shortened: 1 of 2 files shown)


mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
 	{
 		scope(failure) disconnect();
 		foreach (d; documents) if (d["_id"].isNull()) d["_id"] = Bson(BsonObjectID.generate());
-		send(OpCode.Insert, -1, cast(int)flags, collection_name, documents);
+		string collection = collection_name.canFind(".") ? collection_name.split(".")[1] : collection_name;
```
s-ludwig, line 311: What would be the semantic of the first part of `collection'_name` here, the database name? Since it is unused, would that result in the expected behavior if it mismatches?

mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
...
+		Bson operation = Bson([
+			"insert": Bson(collection),
+			"$db": Bson(m_settings.database == string.init ? "admin" : m_settings.database),
+			"documents": Bson(documents)
+		]);
+		auto id = send(OpCode.Msg, -1, cast(int)0, cast(ubyte)0, operation);
```
s-ludwig, line 317: This uses `0` instead of the `flags` argument that was used for `OpCode.insert` - I'm not sure how the semantics translate to `OpCode.Msg`, though, probably they have to be added as fields to `operation`?

mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
...
-		else
-			id = send(OpCode.Query, -1, cast(int)flags, collection_name, nskip, nret, query, returnFieldSelector);
-		recvReply!T(id, on_msg, on_doc);
+		Bson op = query;
+		op["$db"] = Bson(m_settings.database == string.init ? "admin" : m_settings.database);
+
+		if (!returnFieldSelector.isNull)
+			op["projection"] = returnFieldSelector;
+
+		if (flags && QueryFlags.tailableCursor)
```
s-ludwig, line 335:
```suggestion
		if (flags & QueryFlags.tailableCursor)
```

mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
...
+			op["tailable"] = Bson(true);
+		if (flags && QueryFlags.awaitData)
```
s-ludwig, line 337:
```suggestion
		if (flags & QueryFlags.awaitData)
```

mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
...
+			op["awaitData"] = Bson(true);
+
+		Bson operation = op;
+		id = send(OpCode.Msg, -1, cast(int)0, cast(ubyte)0, operation);
```
s-ludwig, line 341: There are a few more query flags that are unhandled here: `slaveOk`, `oplogReplay`, `noCursorTimeout`, `exhaust`, `partial`

Also, the `collection_name` parameter is now unused.

mongodb/vibe/db/mongo/connection.d
```diff
@@ -306,26 +308,50 @@ final class MongoConnection {
...
+		recvReply!T(id, on_msg, on_doc, isAggregate);
 	}
 
 	void getMore(T)(string collection_name, int nret, long cursor_id, scope ReplyDelegate on_msg, scope DocDelegate!T on_doc)
 	{
 		scope(failure) disconnect();
-		auto id = send(OpCode.GetMore, -1, cast(int)0, collection_name, nret, cursor_id);
+		string collection = collection_name.canFind(".") ? collection_name.split(".")[1] : collection_name;
```
s-ludwig, line 348: Same question as for `insert` here

s-ludwig (changes requested):
The part of handling `int` in addition to `double` looks okay, I'd just factor the expression into a `private` function to avoid code duplication\*. Using an empty string to mean `"admin"` doesn't feel very clean, but I also can't think of a better approach for now, so that's okay, too.

If possible, I'd ask you to separate these two into a separate PR that can be merged first, so that we can concentrate on the `OpCode.Msg` part here, as I feel that there might be a few more hidden issues not caught by the existing tests.

\* e.g. `void enforceReplyOK(Bson reply, string failure_message)`

WebFreak001:
I've implemented this as well in #2691 now, although I more strictly separated the OP_MSG implementation from OP_QUERY. My implementation checks if OP_MSG can be used as per driver spec and then only continues to use that from that point on. There are a few changes that are crucial for that imo, which aren't in this PR, like changing how `runCommand` works to use either OP_QUERY or OP_MSG. Additionally I have added MongoDB CI tests there from 3.6 and onward. I have decided to drop older compatibility at this point because it's already technically unsupported since Ubuntu 18.04 and has been long EOLd.

I have upgraded most basic queries (find, update, etc) to OP_MSG there, but have kept incompatible ones (e.g. return selector no longer exists afaics) using the legacy OP_QUERY in. I have also deprecated our custom non-conformant CRUD APIs there and replaced them with fully configurable driver-spec conformant versions.

PR #2627, opened 2021-10-27
Normalization capability for URL

mkykadir:
Depends on #2626 due to an unittest, dependency can be removed easily.

URL normalization without semantic changes capability is added

- `normalize` method is added
- `normalized` method is added

### DIFF
--- a/inet/vibe/inet/url.d
+++ b/inet/vibe/inet/url.d
@@ -13,6 +13,7 @@ import vibe.textfilter.urlencode;
 import vibe.utils.string;
 
 import std.array;
+import std.algorithm;
 import std.conv;
 import std.exception;
 import std.string;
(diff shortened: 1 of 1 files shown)


inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
 		return cast(NativePath)this.path;
 	}
 
+	/// Decode percent encoded triplets for unreserved or convert to uppercase
+	private string normalize_percent_encoding(const string input)
```
Geod24, line 424: In general, a function should not allocate if it doesn't have it. So if it's taking a `string`, it should return that same `string` if it has nothing to normalize. If the function decides to always allocate, it doesn't need to take a `string`, it should take a `scope const(char)`, as `string` implies `immutable`, but the function does not require immutability.

inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
...
+		if (this.m_path.empty || isDirectory)
+			this.m_path.endsWithSlash = true;		
+	}
+
+	/** Returns the normalized form of the URL.
+
+		See `normalize` for a full description.
+	*/
+	@property URL normalized()
+	const {
```
Geod24, line 505: Since this is making a copy and returning that copy, i don't think `@property` is right here.
Note that `@property` is almost useless (the only thing it does is make `typeof(normalized)` be `URL` instead of `URL delegate()`).
I would say, just keep `normalize`, if someone wants to keep the original, it's trivial to write this function from outside the `struct`.

inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
...
+	/**
+		Normalizes URL by:
+			- Converting schema and host to lowercase
+			- Removing port if it is the default port for schema
+			- Removing dot segments in path
+			- Converting percent-encoded triplets to uppercase
+			- Adding slash when path is empty
+			- Adding slash to path when path represents a directory
+			- Decoding percent encoded triplets for unreserved characters
+				A-Z a-z 0-9 - . _ ~
```
Geod24, line 468: This won't look too good in DDOC. Remember that in DDOC, the first paragraph (anything before the first empty line or section definition) is the title. Then the description. Both are optional.

In other words:
```D
/**
  * Title which is a concise description of this function
  *
  * More extended description of what's going on, perhaps corner cases.
  *
  * FooBar:
  *   Will appear under its own section, if "FooBar" means something special for the section.
  *   Note that `Params`, `Returns`, and co are just sections, with a special meaning.
  *
  * If_your_section_name_needs_whitespace:
  *   Then you can just use underscore.
  */
```

So in this example, the documentation will all show up in the title, which is far from concise.
One suggestion would be to use something like:
```D
/**
  * Normalize the content of this `URL` in place
  *
  * Normalization can be used to create a more consistent and human-friendly
  * string representation of the `URL`.
  * The list of transformations applied in the process of normalization is as follows:
  * [...] 
  */
```

inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
...
+			- Removing dot segments in path
+			- Converting percent-encoded triplets to uppercase
+			- Adding slash when path is empty
+			- Adding slash to path when path represents a directory
+			- Decoding percent encoded triplets for unreserved characters
+				A-Z a-z 0-9 - . _ ~ 
+
+		Params:
+			isDirectory = Path of the URL represents a directory, a trailing
+				slash is added when true
```
Geod24, line 472: "a trailing slash is added when true"

Two things:
- "if one is not already present, a trailing slash will be appended when `true`"
- What happens if this is false but there is a trailing slash ?

inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
...
+	*/
+	void normalize(bool isDirectory = false)
+	{
+		import std.uni : toLower;
```
Geod24, line 476: Already imported at the top level

inet/vibe/inet/url.d
```diff
@@ -791,6 +880,31 @@ unittest { // native path <-> URL conversion
...
+	assert(url.normalized.toString() == "http://example.com/~foo");
+	
+	url = URL.parse("http://example.com/foo/./bar/baz/../qux");
+	assert(url.normalized.toString() == "http://example.com/foo/bar/qux");
+	
+	url = URL.parse("http://example.com");
+	assert(url.normalized.toString() == "http://example.com/");
+	
+	url = URL.parse("http://example.com:80/");
+	assert(url.normalized.toString() == "http://example.com/");
```
Geod24, line 900: Add a test with https too

inet/vibe/inet/url.d
```diff
@@ -791,6 +880,31 @@ unittest { // native path <-> URL conversion
...
+	
+	url = URL.parse("http://example.com/foo");
+	url.normalize(true);
+	assert(url.toString() == "http://example.com/foo/");
+
```
Geod24, line 905: Empty line

inet/vibe/inet/url.d
```diff
@@ -419,6 +420,94 @@ struct URL {
...
+			this.m_port = 0;
+
+		this.m_path.normalize();
+
+		// Normalize percent encoding, decode unreserved or uppercase hex
+		this.m_queryString = normalize_percent_encoding(this.m_queryString);
+		this.m_anchor = normalize_percent_encoding(this.m_anchor);
+		this.m_path = InetPath(this.m_path.bySegment2.map!(
+				n => InetPath.Segment2.fromTrustedEncodedString(normalize_percent_encoding(n.encodedName))
+			).array);
```
Geod24, line 493: Any chance we can avoid modifying `m_path` twice ?

Geod24 (changes requested):

Geod24:
The dependent PR has been merged, this can be rebased.

inet/vibe/inet/url.d
mkykadir: Trailing slash is never removed nor added when this is `false` (except when path is empty it is added)

inet/vibe/inet/url.d
mkykadir: It is `ascii.toLower` (at module level)

inet/vibe/inet/url.d
mkykadir: Normalization process always allocates without checking need for it (i.e. schema and host `toLower` case); I think we can do it for this one also.

inet/vibe/inet/url.d
```diff
@@ -818,6 +911,33 @@ unittest { // native path <-> URL conversion
...
+	url = URL.parse("http://example.com/foo/./bar/baz/../qux");
+	assert(url.normalized.toString() == "http://example.com/foo/bar/qux");
+	
+	url = URL.parse("http://example.com");
+	assert(url.normalized.toString() == "http://example.com/");
+	
+	url = URL.parse("http://example.com:80/");
+	assert(url.normalized.toString() == "http://example.com/");
+
+	url = URL.parse("hTTPs://examPLe.COM/my/path");
```
Geod24, line 933: Sorry, should have been more clear: Add an example with the HTTPS port (443)

inet/vibe/inet/url.d
```diff
@@ -418,6 +419,98 @@ struct URL {
...
+						break;
+				}
+
+				i += 2;
+			}
+			else
+				normalized ~= c;
+		}
+
+		return normalized.idup;
```
Geod24, line 455:
```suggestion
		return normalized.assumeUnique;
```

Or use an `Appender` :)

mkykadir: `assumeUnique` is not `@safe`, thus using appender

Geod24 (approved):

PR #2633, opened 2021-12-29
Fixes #2575: properly handle LimitedInputStream read IOMode

### LINKED ISSUE #2575: LimitedInputStream read mode handling
`LimitedInputStream` in `vibe.stream.counting` is used to limit the number of bytes available to readers in cases similar to when an HTTP response specifies the body content-length.

However, current implementation forbids the usage of the following pattern that uses `IOMode.once`:
```D
// input is an instance of LimitedInputStream in case Content-length header is available
auto input = res.bodyReader();

// 1024 is larger than received content length
const bufferLength = exactBuffer ? res.headers["Content-length"].to!size_t() : 1024;
ubyte[] buffer = new ubyte[](bufferLength);

while (!input.empty)
{
	const chunk = input.read(buffer[], IOMode.once); // <---- IOMode.once is important
	logInfo("received: %s", buffer[0 .. chunk].assumeUTF);
}
```

My understanding of `IOMode.once` is that the stream will provide whatever data is *currently available* (limited by the provided buffer size of course) or will block until at least some data is available.

Current implementation of `LimitedInputStream` checks whether the provided buffer length exceeds the maximum allowed length to be read and errors if so without considering `IOMode`.

So is my understanding and my expectation of the pattern above to work correctly correct?
If so, I would change `LimitedInputStream.read` to:
- always check length only in `IOMode.all`
- provide whatever number of bytes is allowed in `IOMode.immediate` and `IOMode.once`.
- only error out when read is called with `IOMode.immediate` and `IOMode.once` when all allowed bytes were consumed *previously*.

Below is a runnable test case:
```D
import vibe.core.core;
import vibe.core.net;
import vibe.http.server, vibe.http.client;
import vibe.stream.operations;
import vibe.core.log;

import std.format: format;
import std.string: assumeUTF;
import std.conv: to;

void main()
{
	/*
	when exactBuffer is true,
	the buffer into which the data is being read has
	equal length to the content-length of the response.

	when exactBuffer is false, the buffer is 1024 bytes long.

	The problem manifests when a larger buffer than the data
	available on the stream is used:
	vibe.stream.counting.LimitedInputStream errors with
	`vibe.stream.counting.LimitException@../../stream/vibe/stream/counting.d(134): Size limit reached`
	*/

	bool exactBuffer = false;

	auto server = setupServer();
	runTask(&client, server.bindAddresses[0], exactBuffer);
	runApplication();
	server.stopListening();
}

HTTPListener setupServer()
{
	auto settings = new HTTPServerSettings();
	settings.port = 0;

	return listenHTTP(settings, &handler);
}

bool client(NetworkAddress addr, bool exactBuffer) nothrow
{
	try
	{
		const url = "http://%s/".format(addr.toString());

		requestHTTP(
			url,
			(scope req) {},
			(scope res)
			{
				auto input = res.bodyReader();

				// 1024 is larger than received content length
				const bufferLength = exactBuffer ? res.headers["Content-length"].to!ulong() : 1024;
				ubyte[] buffer = new ubyte[](bufferLength);

				while (!input.empty)
				{
					const chunk = input.read(buffer[], IOMode.once);
					logInfo("received: %s", buffer[0 .. chunk].assumeUTF);
				}
			}
		);

		return true;
	}
	catch (Exception ex)
	{
		logError("Exception thrown: %s", ex);

		return false;
	}
	finally
	{
		exitEventLoop();
	}
}

void handler(scope HTTPServerRequest _, scope HTTPServerResponse res)
{
	res.writeBody("my length is 15");
}
```

yazd:
Fixes https://github.com/vibe-d/vibe.d/issues/2575.

### DIFF
--- a/stream/vibe/stream/counting.d
+++ b/stream/vibe/stream/counting.d
@@ -122,8 +122,12 @@ class LimitedInputStream : InputStream {
 
 	size_t read(scope ubyte[] dst, IOMode mode)
 	{
-		if (dst.length > m_sizeLimit) onSizeLimitReached();
-		auto ret = m_input.read(dst, mode);
+		import std.algorithm: min;
+
+		if ((mode == IOMode.all || m_sizeLimit == 0) && dst.length > m_sizeLimit) onSizeLimitReached();
+
+		const validReadSize = min(dst.length, m_sizeLimit);
+		auto ret = m_input.read(dst[0 .. validReadSize], mode);
 		m_sizeLimit -= ret;
 		return ret;
 	}
--- a/stream/vibe/stream/counting.d
+++ b/stream/vibe/stream/counting.d
@@ -139,6 +139,56 @@ class LimitedInputStream : InputStream {
 	}
 }
 
+unittest { // issue 2575
+	import vibe.stream.memory : createMemoryStream;
+	import std.exception : assertThrown;
+
+	auto buf = new ubyte[](1024);
+	foreach (i, ref b; buf) b = cast(ubyte) i;
+	auto input = createMemoryStream(buf, false);
+
+	// test IOMode.once and IOMode.immediate
+	static foreach (bufferSize; [100, 128, 200])
+	static foreach (ioMode; [IOMode.once, IOMode.immediate])
+	{{
+		input.seek(0);
+		auto limitedStream = createLimitedInputStream(input, 128);
+
+		ubyte[] result;
+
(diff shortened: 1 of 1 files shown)


Geod24:
Sorry for the delayed response. Any chance we can get a test on this ?

yazd:
No problem.
Tests added.

s-ludwig (approved):

PR #2634, opened 2021-12-29
Add nothrow to vibe.http.client.defaultSettings

yazd:
`defaultSettings` does not throw.


Geod24:
I don't even know why it's a method. Shouldn't a manifest constant just work ?

yazd:
I'm not sure. I don't know why it's a class either.

s-ludwig:
The class thing has purely historical reasons (same for most other classes, as is evident from the transition to struct for stream classes). A manifest constant or `immutable` constant would probably work fine here, it's also `package`, so from my side either way is fine to merge.

PR #2635, opened 2021-12-30
Remove usage of shared static this in examples

Geod24:
```
It has been discussed and decided in vibe.d/vibe.d#2370 that 'shared static this'
was an early feature that nowadays should go.
The first step to that is to remove it from examples so new users
will not be tempted to use it.

In addition, the examples were tidied up a bit, by always adding a module declaration,
and storing the return values of listen* / runTask whenever applicable.
Some places used spaces instead of tabs, which was also fixed.
```

### DIFF
--- a/examples/app_skeleton/dub.json
+++ b/examples/app_skeleton/dub.json
@@ -1,13 +1,7 @@
 ﻿{
 	"name": "app-skeleton",
 	"description": "An example project skeleton",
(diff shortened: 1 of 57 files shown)


maxhaton (approved):

Geod24:
Given @s-ludwig has voiced his agreement to the whole idea, but is quite busy, just going to move ahead with this.

PR #2637, opened 2022-01-18
http.client: Introduce an overload of connectHTTP that accepts URL

Geod24:
Fairly simple, de-duplicate the work done in multiple methods and allow other callers to do so, e.g. if one wants to have persistent connections in `vibe.web.rest`.


PetarKirov (approved):
LGTM

PR #2643, opened 2022-02-24
Invoke reject connection predicate with `X-Forwarded-For`

### LINKED ISSUE #2639: HTTP connection rejection might be too early
Vibe.d can be used behind a reverse proxy/load balancer configurations in which `TCPConnection.remoteAddress` might point to reverse proxy. HTTP has `Forwarded` header for those configurations but `rejectConnectionPredicate` is too early for considering those header fields. **Another** ban checking mechanism can be placed after HTTP Headers are parsed for providing flexibility under those configurations without loosing early banning capability.

mkykadir:
TCP connection client address was not enough for configurations behind
reverse proxies. This will allow rejecting connections from actual
clients behind proxies.

Fixes #2639

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -2280,6 +2280,25 @@ private bool handleRequest(InterfaceProxy!Stream http_stream, TCPConnection tcp_
 		logTrace("persist: %s", req.persistent);
 		keep_alive = req.persistent;
 
+		if (context.settings.rejectConnectionPredicate !is null)
+		{
+			import std.socket : Address, parseAddress;
+			
+			auto forward = req.headers.get("X-Forwarded-For", null);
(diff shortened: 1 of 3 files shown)


Geod24:
Can you add test(s) ?

mkykadir:
> Can you add test(s) ?

Added one

Geod24:
@yazd : I don't see how spoofing `X-Forwarded-For` would help though. The connection delegate is still called with the IP address of the requester, but also with the value of the header. Hence, someone cannot just bypass the check by providing an IP in the header. I assume Nginx and other proxies will do the "right thing" and add to the header whenever it is already present. Or am I missing something ?

Geod24 (approved):

PR #2646, opened 2022-04-01
OpenSSL TLS version handling improvements

s-ludwig:
Updates the default cypher suite, removes dysfunctional SSLv3 support and adds a set of tests to check proper TLS version matching between client and server.


l-kramer (approved):

RazvanN7:
@s-ludwig @l-kramer it seems that this has broken the dmd pipeline suite. Please see #2647 .

PR #2651, opened 2022-05-11
Fix using URL("...") at compile time.

s-ludwig:
Avoids calling isCommonInternetSchema and instead derives this information from the URL itself. This is possible, because URLs that don't conform to the schema are not allowed to start with "//". See RFC 3986 section 3.3.

### DIFF
--- a/inet/vibe/inet/url.d
+++ b/inet/vibe/inet/url.d
@@ -157,9 +157,8 @@ struct URL {
 			str = str[idx+1 .. $];
 			bool requires_host = false;
 
-			if (isCommonInternetSchema(m_schema)) {
+			if (str.startsWith("//")) {
 				// proto://server/path style
-				enforce(str.startsWith("//"), "URL must start with proto://...");
 				requires_host = true;
 				str = str[2 .. $];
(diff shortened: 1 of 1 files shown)


Geod24:
So we don't have to register schemas anymore ?
One usage we did of this was:
- https://github.com/bosagora/agora/blob/a9ee679d8872f70aa5c25c82d83f8935dd5ac470/source/agora/common/Types.d#L39-L43
- https://github.com/bosagora/agora/blob/a9ee679d8872f70aa5c25c82d83f8935dd5ac470/source/agora/common/Types.d#L130-L143

LGTM

s-ludwig:
Yeah, this kinda removes the need for the registration, although there is still the default port. The only downside is that now the parser doesn't enforce common internet style for such URLs, but parses them as if they had an empty host name. Since there are often other restrictions to the URL schema, I'm not sure how much value that enforcement actually has, though.

But from an API POV this feels a lot cleaner, not relying on global runtime information for a generic URL representation...

Geod24 (approved):

PR #2653, opened 2022-05-13
Separate the relaxed URL parsing API from the general URL parsing and…

s-ludwig:
… support relative references.

This is supposed to be extended with more rules (following the w3c guidelines) later.


inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
 import std.string;
 import std.traits : isInstanceOf;
 import std.ascii : isAlpha, isASCII, toLower;
-import std.uri: encode;
+import std.uri: decode, encode;
 
 import core.checkedint : addu;
 
 
+/** Parses a user-entered URL with relaxed parsing rules.
```
Geod24, line 27:
```suggestion
/** Parses a user-provided URL with relaxed rules.
```

Entered is a bit too specific, and "parsing rules" is redundant.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
...
+
+	Params:
+		url = String representation of the URL
+		base_url = Base
+*/
+URL parseUserURL(string url, URL base_url = URL.init)
+{
+	// test if this is a URL or a relative (path) reference
+	auto sidx = url.indexOf(':');
+	if (sidx < 0) {
```
Geod24, line 43: I would say that the fact that it's an URL is the special case, so use an early return:
```D
if (sidx >= 0)
    return URL(url, false);
```

This way we can avoid putting 90% f the function in an `if`.
Also, why do we pass `false` as `decode` ? It differs from what the ctor does by default and thus violates the principle of least surprise.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
...
 import core.checkedint : addu;
 
 
+/** Parses a user-entered URL with relaxed parsing rules.
+
+	Unlike `URL.parse`, this allows the URL to use special charaters as part of
+	the host name and path, automatically emplying punycode or percent-encoding
+	to convert this to a valid URL.
+
+	Also, relative URL references are supported by supplying a base URL.
```
Geod24, line 33: There's a few typos (`charaters`, `emplying`), but I'm a bit confused about what this is supposed to do.
Will ask questions in the review comment and then suggest a rewrite.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
...
+
+		auto path = InetPath(url);
+
+		if (base_url != URL.init) {
+			if (path.absolute)
+				base_url.path = path;
+			else if (base_url.path.endsWithSlash)
+				base_url ~= path;
+			else base_url.path = base_url.path.parentPath ~ path;
+			return base_url;
```
Geod24, line 59: Personally I have different expectations for the "base URL". I assumed that if the user provides a "full" URL, then it is used. Otherwise, I provide a default. I am not saying my expectations are more sensible than what this function is doing, but it would be best if what the function does was documented a bit more.

Something like:
> If a `base` URL is provided, it will always be used, and only the path part of `url` will be taken into account. Moreover, if the base URL is not a directory (e.g. `http://example.com/foo`), the file path (here: `foo`) will be replaced. To ensure the base URL is appended to, make sure it ends with a directory (e.g. `http://example.com/foo/`).

Geod24: Having a few documented unittests serving as example would help, so instead of having examples between parenthesis, one could write:
```D
unittest
{
    // More basic tests...
    
    // When a base URL is provided, it is always used
    assert(parseUserURL("vibe-d.org/foo/bar", URL("http://example.com")) == URL("http://example.com/foo/bar"));
    // Etc...
}
```

Actually while writing the example I realized one thing: The base URL is not *always* used. If the user-inputted string is a full URL with proto, then it is ignored, another thing that might be surprising to users.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
...
+		} else {
+			if (!path.absolute) path = InetPath.fromTrustedString("/") ~ path;
+			return URL("file", null, 0, path);
+		}
+	}
+
+	return URL(url, false);
```
Geod24, line 66: I'd recommend calling `normalize` on the return value, in both branches.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,75 @@ import std.exception;
...
+}
+
+unittest {
```
Geod24, line 69: I'm not very found of moving those unittests. They test different functions / part of the code.
The previous unittests could be moved to documented unittests for the ctor.

inet/vibe/inet/url.d
```diff
@@ -220,21 +289,6 @@ struct URL {
...
-	* Parse a 'plain' string into an `URL`.
-	*
-	* Unlike `URL.parse`, this expects its argument to be a 
-	* plain url (not percent-encoded nor punyencoded), and thus can contain
-	* non-ASCII characters as well as reserved ones (e.g. a space).
-	*
-	* Params:
-	*   url_string = A plaintext URL, which will be percent-encoded in the result.
-	*/
-	static URL parsePlain(string url_string)
```
Geod24, line 234: Isn't that breaking code ? Or was it not part of a release ?

Geod24 (changes requested):
Quite a few points. The main one being that the API feels inconsistent and the expectations should be better documented IMO. Some documented unittests showcasing the main features in a simple manner might be the best approach.

inet/vibe/inet/url.d
s-ludwig: I just did an incremental change here to keep the diff small. Ultimately, I'd expect to have the whole encoding logic in this function instead of in a special constructor of `URL`.

inet/vibe/inet/url.d
s-ludwig: I'll fix those, this was mostly copy-pasted from `parsePlain`.

inet/vibe/inet/url.d
s-ludwig: This is a good point, right now the logic is to treat something that is not explicitly a URL (with ":") as a path, whereas the expectation for entering something in the browser's address bar is to treat a non-absolute path as a http(s) URL. I'll probably just remove the whole `base_url` logic then, because I didn't really intend to invest time into the relaxed parsing at this point.

But fortunately we can simply follow the W3C guidelines here.

inet/vibe/inet/url.d
s-ludwig: Do they test different parts, though?

inet/vibe/inet/url.d
s-ludwig: This was not yet released.

inet/vibe/inet/url.d
Geod24: Yes and no. Yes because the encoding logic is mixed with the rest of the ctor parsing because, AFAIR, it has to be.
No because a machine-provided URL should be its `URL` representation, so in that sense the support in the constructor is "part of" `parseUserURL`. Still, unittests for the encoding itself should be next to the code, while documented unittests should be next to the API.

inet/vibe/inet/url.d
Geod24: I think what the default protocol should be is up to the application.
A browser will try `https`, then `http`, but if the URL is read from the configuration file of an FTP client, then that would be the default protocol (or one of the alternative).
In our case, we had a P2P protocol, so we didn't want people to have to type `agora://host.foo`.
IOW, instead of a base URL, maybe consider a default (as in, used in absence of explicit value) protocol ?

inet/vibe/inet/url.d
s-ludwig: Well, as soon as we go further with this, we'll have to separate the parsing logic completely. Right now it kind of makes sense to share the logic, but eventually there will be too many special rules in the `parseUserURL` path. IMO we could just remove the `encoded == false` path from the constructor and instead duplicate it to `parseUserURL` directly.

inet/vibe/inet/url.d
s-ludwig: I have already removed the relative reference handling. I was thinking about the way relative references can be used within e.g. HTML documents, but thinking about it, I can't really see the use case from a user's point of view. But having a default protocol parameter makes a lot of sense.

inet/vibe/inet/url.d
```diff
@@ -19,11 +19,43 @@ import std.exception;
...
+/** Parses a user-provided URL with relaxed rules.
+
+	Unlike `URL.parse`, this allows the URL to use special characters as part of
+	the host name and path, automatically employing puny code or percent-encoding
+	to convert this to a valid URL.
+
+	Params:
+		url = String representation of the URL
+*/
+URL parseUserURL(string url)
```
Geod24, line 36: Okay, I now see that essentially, you just wanted to tidy up the API a bit ?
I don't think `parseUserURL` is very useful in its current incarnation though: future changes might need to add parameters, so that'll be a breaking change, which means we'll probably deprecate this overload to add a `URL parseUserURL(string url, string default_protocol)` sooner than later. In that regard, keeping `parsePlain` doesn't seem to bad to me.

But if you prefer to go with that approach, it won't make a huge difference when the time to improve `parseUserURL` comes, so feel free to merge as-is. I'm quite eager to be able to compile Vibe.d on my machine again :D

Geod24 (approved):
LGTM

inet/vibe/inet/url.d
s-ludwig: Agreed, we'll probably add it like that later anyway. I'll add the parameter now then and ignore it until the logic is implemented (still some time until the final 0.9.5 release is tagged).

s-ludwig: > In that regard, keeping parsePlain doesn't seem to bad to me.

IMO the word "plain" just isn't right here and I also wanted to separate the code from the struct definition, because I'd expect that the implementation will become considerably more complex in the future.

s-ludwig:
Rebased and added `default_schema`.

Geod24:
The parameter is there, but it's not used ?

s-ludwig:
Yep, I formulated the documentation for it in a way that leaves enough room for that kind of implementation ;-) Seriously, though, I'd just implement the functionality before the final release, but tag a beta now with just the API.

s-ludwig:
v0.9.5-beta.1 is now available

PR #2655, opened 2022-05-16
openssl: Work around deimos/openssl change in 2.0.2+1.1.1.h

Geod24:
As explained in the comment, a breaking change happened while fixing a bug.


Geod24:
Works locally with deimos/openssl HEAD.

s-ludwig (approved):

PR #2658, opened 2022-05-21
Update Deimos OpenSSL bindings to v3.0.0

Geod24:
```
For a long time, the tls subpackage was stuck on OpenSSL bindings v1.x.x.
Those haven't been released for a long time, and the binding's version
did not match OpenSSL. Over time, workarounds were built to support
multiple versions of OpenSSL, but the issue applied to any OpenSSL user,
and the pin to the v1 major version means that a user could not use
the v2 branch and Vibe.d in the same application.
The main improvement that was done downstream was to auto-detect
the OpenSSL version. The script has now been ported to the binding package,
which can now be used by all downstream applications.
The position on multiple version has now been clarified,
and it has been decided that the bindings will now support a range of versions,
with static if / version condition when necessary.

In order for applications to move forward, we now update Vibe.d to depend on this new major version.
This could prove disruptive to end user, but the change is forward compatible,
and should be (almost) a no-op for end users. There is no good transition path here,
as dub does not allow different subconfigurations to depend on different major versions.
```

@s-ludwig : The underlying motivation here is that I want to work with a library that uses v2 of the bindings, and for it to work, we currently use a fork. I'd like to upgrade the whole system to v3 and save everyone some trouble.
This could prove a bit disruptive for those not using the auto-detection, but I think the gains far outweight the cost.

I need to do more tests on this, but I think ultimately, we could just get rid of the different versions, and let the end user (application) define which configuration of openssl to use.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,12 +15,12 @@ jobs:
           - dmd-2.097.1
           - dmd-2.096.1
           - dmd-2.093.1
-          - dmd-2.088.1
+          - dmd-2.090.1
           - ldc-latest
           - ldc-1.27.1
           - ldc-1.26.0
           - ldc-1.23.0
-          - ldc-1.18.0
+          - ldc-1.21.0
         include:
           # Default
           - { parts: 'builds,unittests,examples,tests,mongo', extra_dflags: '' }
--- a/.gitignore
+++ b/.gitignore
@@ -71,5 +71,3 @@ examples/web/web-framework-example
 examples/websocket/websocket-example
 *.exe
 
-# Ignore auto-generated OpenSSL version
-tls/openssl_version.d
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,37 +3,37 @@ environment:
  dub_version: 1.7.2
  matrix:
   - DC: dmd
-    DVersion: 2.098.0
+    DVersion: 2.100.0
     arch: x64
   - DC: dmd
-    DVersion: 2.098.0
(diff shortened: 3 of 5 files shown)


Geod24:
@s-ludwig : How likely is it that this makes it to the next release ?

Geod24:
Blocked by: https://github.com/D-Programming-Deimos/openssl/pull/67

s-ludwig (approved):
Looks good on the outside. I'll merge and see whether this breaks anything unexpectedly.

Unless there is anything important in the pipeline that I'm missing, I'd also tag another beta and schedule the final release in a week or so.

ljmf00:
@Geod24 @s-ludwig This seems to side effect on @dlang buildkite. See https://buildkite.com/dlang/dmd/builds/26850#0180faa1-d994-4b3e-a04a-b9b7c3fa20da/98-311

Geod24:
Looking into ut

Geod24:
The TL;DR is that the Deimos OpenSSL bindings are missing some v1.0.0 definitions.
Fixing it would take an unpredictable amount of time as I don't have merge access to deimos (and it involves making a new release). Instead, we can work around it here fairly easily. But to ensure that it doesn't take multiple PRs to fix, I'll add a CI for OpenSSL v1.0.2u in the process (it's fairly easy). Once Openssl 1.0.x / Buildkite is fixed, I'll upstream everything (including the CI).

Geod24:
Next: https://github.com/vibe-d/vibe.d/pull/2661

PR #2659, opened 2022-05-22
Small improvements

s-ludwig:


Geod24 (approved):
I tend to use `static assert` over template `if` condition because the former allows to provide a user-friendly error message, while the later is only useful for overloads. But most use cases are not about overloads.

l-kramer (approved):

PR #2665, opened 2022-05-31
openssl: Remove OPENSSL_VERSION, use Deimos' features

Geod24:
Using OPENSSL_VERSION.startsWith is prone to error,
while using OPENSSL_VERSION_BEFORE and OPENSSL_VERSION_AT_LEAST
is neither ambiguous nor as brittle.

### DIFF
--- a/tls/vibe/stream/openssl.d
+++ b/tls/vibe/stream/openssl.d
@@ -31,18 +31,11 @@ import core.thread;
 
 import deimos.openssl.bio;
 import deimos.openssl.err;
+import deimos.openssl.opensslv;
 import deimos.openssl.rand;
 import deimos.openssl.ssl;
 import deimos.openssl.stack;
 import deimos.openssl.x509v3;
-
-version (Botan)
-	enum OPENSSL_VERSION = "0.0.0";
-else
-{
-    import deimos.openssl.opensslv;
-    enum OPENSSL_VERSION = OpenSSLVersion.text;
-}
 
 version (VibePragmaLib) {
 	pragma(lib, "ssl");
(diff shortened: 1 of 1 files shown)


tls/vibe/stream/openssl.d
```diff
@@ -1504,7 +1495,7 @@ private void setSSLError(string msg, string submsg, int line = __LINE__, string
 	ERR_add_error_data(3, msg.toStringz, ": ".ptr, submsg.toStringz);
 }
 
-static if (OPENSSL_VERSION.startsWith("1.1")) {
+static if (OPENSSL_VERSION_AT_LEAST(1, 1, 0)) {
```
Geod24, line 1498: Note that this is probable very broken on OpenSSL 3.0 as a result, because it uses a (most likely outdated) internal struct definition.

tls/vibe/stream/openssl.d
```diff
@@ -1041,7 +1032,7 @@ final class OpenSSLContext : TLSContext {
 			// as it is always enabled by default.
 			// https://github.com/openssl/openssl/issues/1437
 			// https://github.com/openssl/openssl/commit/2ecb9f2d18614fb7b7b42830a358b7163ed43221
-			static if (OPENSSL_VERSION_NUMBER >= 0x10200000 && OPENSSL_VERSION_NUMBER < OPENSSL_MAKE_VERSION(1, 1, 0, 0)) {
+			static if (OPENSSL_VERSION_AT_LEAST(1, 0, 2) && OPENSSL_VERSION_BEFORE(1, 1, 0)) {
```
Geod24, line 1035: Looks like this introduce a functional change (?)

Geod24: Because the check was wrong:
```D
import std.format;
pragma(msg, format!"%X"(OPENSSL_MAKE_VERSION(1, 0, 1, 0)));
pragma(msg, format!"%X"(OPENSSL_MAKE_VERSION(1, 1, 0, 0)));
pragma(msg, format!"%X"(OPENSSL_MAKE_VERSION(3, 0, 0, 0)));
```

```shell
1000100F
1010000F
3000000F
```

Geod24: I'll revert this specific line for now

PR #2666, opened 2022-05-31
openssl: All supported versions have ECDH

Geod24:
The check included all versions >= 1.0.1, which covers the last 10 years.


Geod24:
Should be fairly straightforward so self-merging

PR #2679, opened 2022-07-31
Add authSource support for MongoDB connection string

ReallySnazzy:


ReallySnazzy:
#2691 mentions it covers this change

Issue #2656, opened 2022-05-18
deprecation message spam


schveiguy:
In the latest version, vibe.http.status.HTTPStatus has an entire group of deprecated members. While this makes sense for people who are using them individually, there is a problem that `formatValue` will still build code that uses those deprecated members.

So if I try to log an error, I get pages and pages of deprecation messages, none of which are important.

Is there a way we can use maybe a version to completely remove those deprecated members from the enum?

PR #2680, opened 2022-08-01
Fix issue #2656

schveiguy:
By defining VibeNoDeprecatedEnums, the HTTPStatus enum will not have any deprecated members, and deprecations due to trying to format such an enum variable will not spew pages of deprecation spam.

I'm not sure how the doc generator will behave, but it can't be worse than the deprecation spam I got on every build.

Note that this doesn't actually fix the spam for existing build configurations, you still have to define the version.

### DIFF
--- a/http/vibe/http/status.d
+++ b/http/vibe/http/status.d
@@ -7,10 +7,7 @@
 */
 module vibe.http.status;
 
-/**
-	Definitions of all standard HTTP status codes.
-*/
-enum HTTPStatus {
+enum HTTPStatusBase {
 	continue_                    = 100,
 	switchingProtocols           = 101,
 	ok                           = 200,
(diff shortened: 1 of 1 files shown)


schveiguy:
FYI, I think all the deprecated enums should be turned off by this version, but I didn't take the time to find them all. If this technique looks good, I can do it.

s-ludwig:
Sorry, I wanted to get to this earlier, but got sidetracked. So for this particular case I think it's just time to remove the deprecated values (WIP in #2684). However, there are also a bunch of enum values "scheduled for deprecation" in the JSON and BSON modules, which may result in the same issue, so this might be a way to fix it there.

Ultimately, I think there should be a way in generic code to access deprecated members without triggering a warning. It wasn't obvious how to even test whether a symbol is deprecated without triggering the warning, when I was writing [this code](https://github.com/s-ludwig/dynamic/blob/master/source/dynamic.d#L87), but unfortunately that's not enough for `format`.

schveiguy:
I think just `__traits(isDeprecated)` is the only way to use a symbol without triggering the deprecation warning.

There really isn't a way to have generic code access deprecated members without triggering a warning except via that mechanism. In the case of `format`, it would have to determine at runtime whether you are using a deprecated member, so that doesn't even cover that case anyway.

It's not an easy problem to solve for sure. In the past, I've just removed those deprecated symbols to avoid the spam. But clearly that can't be done for all projects that might actually still use the symbols. This was the best compromise I could come up with.

s-ludwig:
> I think just __traits(isDeprecated) is the only way to use a symbol without triggering the deprecation warning.

Okay, that one appears to have been fixed to gag the warning in the meantime. It didn't do that at the time.

> There really isn't a way to have generic code access deprecated members without triggering a warning except via that mechanism. In the case of format, it would have to determine at runtime whether you are using a deprecated member, so that doesn't even cover that case anyway.

[This](https://github.com/vibe-d/vibe.d/pull/2684/commits/84bea25566488ec8b96779aabc7fefd03ada1bac#diff-387a5f807fd678bd31095ba87356cbfbf0fe480f04087986549b5a386abfdeca) is my solution for now.

schveiguy:
I'll close this for now. If the changes you are making fix my issue then I'll close that too.

I *think* it will work, but I'm not sure if I'm directly calling `format` on an enum value itself, rather than something that contains it.

PR #2684, opened 2022-08-19
Remove deprecated symbols/modules and deprecate symbols scheduled for deprecation

s-ludwig:


utils/vibe/internal/conv.d
```diff
@@ -0,0 +1,22 @@
...
+				case __traits(getMember, E, m): return m;
+			}
+		}
+	}
+}
+
+private enum isDeprecated(alias parent, string symbol) =
+	!__traits(compiles, {
+		static assert(!__traits(isDeprecated, __traits(getMember, parent, symbol)));
+	});
```
schveiguy, line 22: Why isn't this just: ```d
private enum isDeprecated(alias parent, string symbol) = __traits(isDeprecated, __traits(getMember, parent, symbol));
```
In my PR, just using that mechanism didn't trigger the deprecation. Honestly, I don't know why you need this wrapper here.

schveiguy: Also, this is fine for enums without repeat values, but will not work for ones with repeat values. It might be OK if this is just for internal use.

s-ludwig: Okay, testing on run.dang.io, this has apparently been fixed in 2.089.0. We support 2.090.1 onwards, so I'll remove this workaround.

s-ludwig: > It might be OK if this is just for internal use.

That's the case here.

l-kramer (approved):

PR #2699, opened 2022-11-03
REST input stream fix

s-ludwig:
Fixes issues with wrong content type specifications and handling `InputStream` return values in the REST interface client and server.

### DIFF
--- a/web/vibe/web/rest.d
+++ b/web/vibe/web/rest.d
@@ -1641,6 +1641,7 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
 				auto ret = () @trusted {
 					return evaluateOutputModifiers!CFunc(
 						__traits(getMember, inst, Method)(params), req, res); } ();
+				res.headers["Content-Type"] = "application/octet-stream";
 				ret.pipe(res.bodyWriter);
 			} else {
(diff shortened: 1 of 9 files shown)


web/vibe/web/rest.d
```diff
@@ -1641,6 +1641,7 @@ private HTTPServerRequestDelegate jsonMethodHandler(alias Func, size_t ridx, T)(
 				auto ret = () @trusted {
 					return evaluateOutputModifiers!CFunc(
 						__traits(getMember, inst, Method)(params), req, res); } ();
+				res.headers["Content-Type"] = "application/octet-stream";
```
Geod24, line 1644: Wouldn't this overwrite the `Content-Type` a user sets using `@headerParam` ?

s-ludwig: I guess, although I'm unsure why you would want to set the content type using a header parameter. `@contentType` doesn't appear to be supported by the REST code so far, which is what I thought should be minded here.

s-ludwig: It should be said, though, that the other paths using `writeBody` have already been overwriting the content type until now. IMO this is okay until `@contentType` support gets added.

l-kramer (approved):

PR #2707, opened 2023-02-20
Attempt to fix OpenSSL linker errors for the default configuration

s-ludwig:


tls/dub.sdl
```diff
@@ -16,6 +16,8 @@ configuration "openssl-mscoff" {
 	platforms "windows-x86_mscoff" "windows-x86_64" "windows-x86-ldc"
 	dependency "openssl" version="~>3.1"
 	subConfiguration "openssl" "library-manual-version"
+	dependency "openssl-static" version="~>1.0"
```
Geod24, line 19: New dependency ? Why ?

s-ludwig: That's just to allow forcing the "none" configuration. I'll try to rework that into an optional dependency to avoid the download.

PR #2708, opened 2023-02-20
Fix MongoDB CI tests

s-ludwig:


mongodb/vibe/db/mongo/client.d
```diff
@@ -56,26 +56,23 @@ final class MongoClient {
...
-				ret.connect();
+				try ret.connect();
+				catch (Exception e) {
+					// avoid leaking the connection to the GC, which might
+					// destroy it during shutdown when all of vibe.d has
+					// already been destroyed, which in turn causes a null
+					// pointer
+					() @trusted { destroy(ret); } ();
+					throw e;
+				}
```
Geod24, line 75: `scope (failure)` ?

s-ludwig: `scope (failure)` has the big disadvantage that it also catches `Error` and other `Throwable`s, which may lead to infinite recursions and all other kinds of bad effects. I'm only using it in very few situations for that reason nowadays.

s-ludwig:
Okay, all green again now.

PR #2709, opened 2023-02-21
Make the openssl-static dependency optional and avoid copying unnecessary DLLs

s-ludwig:


Geod24:
I still don't understand why we need another library. If we can't handle this scenario with dub, something is seriously wrong.

s-ludwig:
Hm, what exactly do you mean with "can't handle this scenario with dub"? The openssl-static package simply provides a uniform set of static OpenSSL binaries across a wide range of platforms, because especially on macOS/iOS that's otherwise a PITA to handle.

s-ludwig:
@Geod24: I'll have to merge this soon, as it also fixes a dependency issue in a downstream project (there might be an issue in the optional dependency logic).

PR #2711, opened 2023-02-23
Add vibe-core 2.x.x compatibility

s-ludwig:


Geod24:
I think we have a good example of why forcing `nothrow` runTask is quite annoying :)

s-ludwig:
The point is that the two alternative scenarios are much more annoying:
- Terminate the app for uncaught exceptions - so instead of letting the compiler do its job we go the JavaScript route and get unforseen runtime errors
- Ignore uncaught exceptions (as it was until now) - silently running into bugs at runtime that may have much more severe consequences later during runtime

s-ludwig:
Since we have no language tools to enforce handling such exceptions, except maybe for variants like "scoped tasks" that automatically get `join`ed at the end of the calling scope, I don't see an alternative way to go.

Geod24:
@s-ludwig : What else do we need for a new release ? The amount of deprecations experienced by people while building is really high.

s-ludwig:
I think the only change left is the removal of `GenericPath.Segment` in vibe-core, anything else should then hopefully be backwards compatible changes. I'll open a PR for that.

PR #2715, opened 2023-03-04
Reverse the vibe-core 1.x.x/2.x.x test configurations.

s-ludwig:
Now that vibe-core 2.0.0 is tagged, it will be chosen by default, so we need to force testing 1.x.x.

This revealed a few issues in the build matrix and some additional compile errors in the example projects that need to be fixed.


s-ludwig:
Note that I had to disable required status checks due to the changed names for the new build matrix. I'll add new required checks after the matrix is active on master.

Geod24 (approved):

Geod24:
That fixed Buildkite, nice

PR #2724, opened 2023-04-06
WIP: Qualify some members as scope

nordlow:


nordlow:
Ready for review.

nordlow:
Why isn't CI activated? Does https://github.com/vibe-d/vibe.d/pull/2726 needs to merged first?

s-ludwig:
CI checks needed to be approved manually.

data/vibe/data/bson.d
```diff
@@ -165,7 +165,7 @@ struct Bson {
 		A slice of the first bytes of `data` is stored, containg the data related to the value. An
 		exception is thrown if `data` is too short.
 	*/
-	this(Type type, bdata_t data)
+	this(Type type, bdata_t data) scope
```
s-ludwig, line 168: I honestly find these annotations quite ridiculous. Why would a struct in D ever have non-scope methods? Since structs can always be moved, storing a pointer to them is never safe outside of a defined scope. Actually, I began fixing similar deprecation warnings before stopping and throwing the diff away, because this just can't be how this is supposed to work.

@Geod24, do you possibly have more insight here?

PetarKirov: > Since structs can always be moved, storing a pointer to them is never safe

I'm not 100% sure, but since C++ interop was considered important a few years back (way before all new importC stuff), supporting self-references was considered important enough (e.g. for certain C++ STL data type implementations, but also for certain niche D applications, IIRC Weka), that first [opPostBlit - DIP1014](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1014.md) was accepted (but never implemented IIRC), later to be replaced by [copy constructors - DIP1018](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1018.md). 

I don't if this interaction with DIP1000 was specifically considered, but in general my understanding is that struct moving with self references is supposed to be supported.

s-ludwig: Do copy constructors have any influence on move operations? My impression was that they are purely a generalization of the postblit mechanics. Also, performing a move is specifically done to avoid calling a copy constructor, so that would be a pretty strange choice of semantics.

But apart from the question of *always* having struct methods be `scope`, shouldn't `scope` at least be inferred here?

And then of course also why the compiler thinks that the `this` pointer is getting escaped by calling a method like this in the first place (which I assumed is the warning that is getting fixed by these changes):
```D
void opAssign(int value)
{
	m_data = toBsonData(value).idup;
	m_type = Type.int_;
}
```

RazvanN7: >  Why would a struct in D ever have non-scope methods?

The plan was to also introduce move constructors to enable the use of interior pointers in structs. From that point of view having non-scope struct methods makes sense. Also, I remember that at Weka they used to have a scenario where they would register structs in a global array and perform some operations on them, but because dmd moves structs freely they were having problems and that was what led to the opPostMove proposal (which was abandoned because it has the same flaws as the postblit).

> Do copy constructors have any influence on move operations?

The copy constructor, currently, is orthogonal to moving. If you are trying to make a copy of an lvalue, the copy constructor will be called; if you are trying to make a copy of an rvalue, a move is performed (even if the copy constructor is disabled).

> But apart from the question of always having struct methods be scope, shouldn't scope at least be inferred here?

In short, if we impose that struct methods are always scope then we basically give up on interior pointers. If we plan on having a move constructor that will finally enable the use of such pointers, then having them by default as non-scope would make sense because if we automatically consider them scope there is no way for the user to mark a method as non-scope. However, indeed, I think that inferring scope for struct methods is the way to go.

skoppe: > I honestly find these annotations quite ridiculous. Why would a struct in D ever have non-scope methods?

Whenever a struct wants to escape something.

data/vibe/data/bson.d
```diff
@@ -202,70 +202,71 @@ struct Bson {
...
 	}
 	/// ditto
-	void opAssign(scope const(char)[] value)
+	void opAssign(scope const(char)[] value) scope
 	{
 		import std.utf;
 		import std.string : representation;
 		debug std.utf.validate(value);
-		auto app = appender!bdata_t();
+		auto app = appender!(ubyte[])();
```
Geod24, line 260: Is this needed ?

data/vibe/data/bson.d
```diff
@@ -661,15 +661,13 @@ struct Bson {
...
-	const {
+	string toString() const scope {
 		auto ret = appender!string;
 		toString(ret);
 		return ret.data;
 	}
 
-	void toString(R)(ref R range)
-	const {
+	void toString(R)(ref R range) const scope {
```
Geod24, line 670: That shouldn't be necessary, is it ?

data/vibe/data/bson.d
```diff
@@ -723,14 +721,14 @@ struct Bson {
...
-	inout(Bson) opIndex(string idx) inout {
+	inout(Bson) opIndex(string idx) inout scope {
 		foreach (string key, v; this.byKeyValue)
 			if( key == idx )
 				return v;
 		return Bson(null);
 	}
 	/// ditto
-	void opIndexAssign(T)(in T value, string idx){
+	void opIndexAssign(T)(in T value, string idx) scope {
```
Geod24, line 731: Shouldn't be necessary

data/vibe/data/bson.d
```diff
@@ -928,7 +926,7 @@ struct Bson {
 	}
 
 	///
-	bool opEquals(ref const Bson other) const {
+	bool opEquals(ref const Bson other) const scope {
```
Geod24, line 929: Qualify `other` as `scope` too ?

data/vibe/data/bson.d
```diff
@@ -948,14 +946,13 @@ struct Bson {
 		return true;
 	}
 	/// ditto
-	bool opEquals(const Bson other) const {
+	bool opEquals(const Bson other) const scope {
```
Geod24, line 949: `scope` on `other` ?

data/vibe/data/bson.d
```diff
@@ -1923,7 +1920,7 @@ private string skipCString(ref bdata_t data)
 	return cast(string)ret;
 }
 
-private void putCString(R)(ref R dst, string str)
+private void putCString(R)(scope ref R dst, string str)
```
Geod24, line 1923: Shouldn't be necessary

data/vibe/data/bson.d
Geod24: @s-ludwig : I'd rather not publicly comment on my opinion of DIP1000. IMHO there's no good solution here, but this will help users without creating any drawbacks.

data/vibe/data/bson.d
nordlow: No. Fixed.

data/vibe/data/bson.d
nordlow: Fixed.

data/vibe/data/bson.d
nordlow: Fixed.

data/vibe/data/bson.d
nordlow: Fixed.

data/vibe/data/bson.d
nordlow: Fixed.

data/vibe/data/bson.d
nordlow: Fixed.

nordlow (commented):
Code adjusted for all comments.

data/vibe/data/bson.d
s-ludwig: Okay, so that confirms what I understood of the situation. Under the *current* regime, a struct shouldn't be able to be `@safe` and non-`scope`, so we really shouldn't have this deprecation warning in the first place, regardless of the technical reason why it is triggered here. What I haven't yet understood, though, is why this randomly occurs in some places, but not every time a method is called on a `scope` struct instance\*.

The problem with just going ahead adding `scope` everywhere is that

1. this code will in practice linger there basically forever, adding permanent clutter when the fix really should be in the compiler
2. this is just the tip of the iceberg - there are tons of methods in other places where the same warning occurs, which is why I stopped going this route earlier

Both together make me really reluctant to just follow the warnings here, although I can certainly understand the urge to just get rid of them as quickly as possible.

\* If nobody has an explanation for this, I'd start a dustmite run and see where that leads.

nordlow: Ok, so how to move forward with this?

WalterBright: > a struct shouldn't be able to be @safe and non-scope
```
struct S
{
    int s;
    void abc(int i) @safe
    {
        s = i;
    }
}
```
compiles.

The thing about struct member functions is they are equivalent to a static member function having the first parameter be `ref this`, as in:
```
struct S
{
    void abc(ref S this) { ... }
}
```
If you could give a specific example of a member function that exhibits baffling behavior, I can help.

WalterBright: Perhaps this will help:
```
struct S {
    int* p;
    void abc(int i) @safe { }
}

void test() @safe {
    int i;
    S s;
    s.p = &i;
    s.abc(1);
}
```
Compiling it yields:
```
test.d(13): Error: scope variable `s` assigned to non-scope parameter `this` calling `abc`
```
but this compiles without error:
```
struct S {
    int* p;
    void abc(int i) @safe { }
}

void test() @safe {
    int i;
    S s;
//    s.p = &i;
    s.abc(1);
}
```
So what's going on? `s` is a local variable. The member `p` is a pointer. The pointer is assigned the address of a local variable `i`. This then makes `s` a scope variable.

Now `s` is passed to `abc()` by ref. The ref prevents the address of `s` from escaping `abc()`. But `s` is wrapping a pointer to a local, so the escape of the contents of `s` must be protected as well. Making `abc()` a scope function will accomplish this. Sure enough, adding `scope` to `abc()` causes the error to go away.

This is semantically no different from:
```
@safe void abc(ref int* p, int i) { }
@safe void test() {
    int i;
    int* p;
    p = &i;
    abc(p, 1);
}
```

s-ludwig:
How do you have to compile this to trigger the warnings? I've just fixed some deprecations triggered by `formattedWrite` in conjunction with a `scope` range (#2732) and would like to see if the ones here can also be fixed in a more isolated place.

PR #2726, opened 2023-04-12
Relax requirement of string value to const(char)[] value

nordlow:
immutable is not required because the `value` characters are copied into the internal buffer.

### DIFF
--- a/data/vibe/data/bson.d
+++ b/data/vibe/data/bson.d
@@ -204,7 +204,7 @@ struct Bson {
 	*/
 	this(double value) { opAssign(value); }
 	/// ditto
-	this(string value, Type type = Type.string)
+	this(scope const(char)[] value, Type type = Type.string)
 	{
 		assert(type == Type.string || type == Type.code || type == Type.symbol);
 		opAssign(value);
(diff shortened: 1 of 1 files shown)


nordlow:
Ready for review.

data/vibe/data/bson.d
```diff
@@ -252,15 +252,17 @@ struct Bson {
...
 	}
 	/// ditto
-	void opAssign(string value)
+	void opAssign(scope const(char)[] value)
 	{
 		import std.utf;
+		import std.string : representation;
 		debug std.utf.validate(value);
-		auto app = appender!bdata_t();
+		auto app = appender!(ubyte[])();
```
WebFreak001, line 260: if you keep this `appender!bdata_t` you don't need the assumeUnique / trusted block below, since appender already does the trusting for you.

Is there any reason in changing the type like this?

nordlow: Using

```d
		auto app = appender!bdata_t();
```

makes the lines

```d
		app.put(toBsonData(cast(int)value.length+1));
		app.put(value.representation);
		app.put(cast(ubyte)0);
```

error as

```
 bson.d   261  10 error           none of the overloads of template `std.array.Appender!(immutable(ubyte)[][]).Appender.put` are callable using argument types `!()(ubyte[])` (d-dmd-all)
 bson.d   262  10 error           none of the overloads of template `std.array.Appender!(immutable(ubyte)[][]).Appender.put` are callable using argument types `!()(const(ubyte)[])` (d-dmd-all)
 bson.d   263  10 error           none of the overloads of template `std.array.Appender!(immutable(ubyte)[][]).Appender.put` are callable using argument types `!()(ubyte)` (d-dmd-all)
```

WebFreak001: I thought bdata_t was declared as `immutable(ubyte)[]`, but from the error messages it looks like it's declared as `immutable(ubyte)[][]` (which I think is really weird)

Then instead use `appender!(immutable(ubyte)[])` and these code examples will work properly.

nordlow:
@Geod24, can you merge this?

data/vibe/data/bson.d
nordlow: Excellent! Thanks!

nordlow:
Read for merge, @Geod24.

PR #2732, opened 2023-05-22
Fix scope related deprecation warnings in vibe.data.json

s-ludwig:
Avoids using formattedWrite directly in order to avoid warnings when passing a scope output range.

Also marks the `line` parameter of `parseJson` `scope` (which probably isn't inferred as such due to the recursive nature of `parseJson`).

See also #2724.

### DIFF
--- a/data/vibe/data/json.d
+++ b/data/vibe/data/json.d
@@ -69,7 +69,7 @@ module vibe.data.json;
 public import vibe.data.serialization;
 public import std.json : JSONException;
 
-import vibe.internal.conv : enumToString;
+import vibe.internal.conv : enumToString, formattedWriteFixed;
 
 import std.algorithm;
 import std.array;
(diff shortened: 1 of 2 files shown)


data/vibe/data/json.d
```diff
@@ -1141,27 +1141,27 @@ struct Json {
...
+			void delegate(scope const(char)[]) @safe sink;
+			void put(scope const(char)[] str) @safe { sink(str); }
 			void put(char ch) @trusted { sink((&ch)[0 .. 1]); }
 		}
 		auto r = DummyRangeS(sink);
 		writeJsonString(r, this);
 	}
 	/// ditto
-	void toString(scope void delegate(const(char)[]) @system sink, FormatSpec!char fmt)
+	void toString(scope void delegate(scope const(char)[]) @system sink, FormatSpec!char fmt)
```
Geod24, line 1157: That's a breaking change isn't it ? Giving it a non-scope delegate will break.

s-ludwig: True, I'll add a separate overload. Not sure what the warning here was anyway, at least the method should be `@system`.

PR #2741, opened 2023-08-10
Test new openssl version

Geod24:


s-ludwig:
Is this something that is still relevant for release? I've tagged an RC, as everything looked okay now after adjusting for "mysterious" CI failures in #2743.

Geod24:
No longer relevant. We had problems internally as this triggered a compiler bug in recent releases (leading to a linking error), but the latest 3 releases are fine.

s-ludwig:
Okay, yeah, we tripped over the same issue probably, and also fixed it with a compiler upgrade.

PR #2748, opened 2023-09-05
Remove support for non-safe REST methods

Geod24:
```
The deprecation has been introduced in v0.8.0, released 2017-07-10, over 6 years ago. The workaround also introduces some issues, as a closure is created (memory allocation) and it does not work for types with scoped destruction.
```

Deprecation introduced in https://github.com/vibe-d/vibe.d/commit/2d9967e23e24d2957651bd1228b5360e54ab7a16

### DIFF
--- a/examples/rest-collections/source/api.d
+++ b/examples/rest-collections/source/api.d
@@ -4,11 +4,13 @@ module api;
 import vibe.web.rest;
 
 interface ForumAPI {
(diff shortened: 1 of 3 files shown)


Geod24:
I think for this to be reasonable it needs to be an error message in a static assert. A compile error deep within Vibe doesn't cut it.

PR #2755, opened 2023-11-23
Add server setting for max HTTP request header line length.

charlesreiss:
Instead of using a hard-coded length of 4096 for the maximum size of lines in the HTTP headers (including the status line), make this configurable in HTTPServerSettings.

### DIFF
--- a/http/vibe/http/server.d
+++ b/http/vibe/http/server.d
@@ -457,7 +457,7 @@ HTTPServerRequest createTestHTTPServerRequest(URL url, HTTPMethod method, InetHe
 @safe {
 	auto tls = url.schema == "https";
 	auto ret = new HTTPServerRequest(Clock.currTime(UTC()), url.port ? url.port : tls ? 443 : 80);
-        ret.m_settings = new HTTPServerSettings;
+	ret.m_settings = new HTTPServerSettings;
 	ret.requestPath = url.path;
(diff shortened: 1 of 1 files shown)


s-ludwig (approved):

s-ludwig:
`dub test :web` appears to result in a segmentation fault with the changes. From the diff alone I'm not sure what causes it.

charlesreiss:
It looked like the problem was that createTestHTTPServerRequest didn't setup m_settings in the HTTPServerRequest, so the test used a null pointer when trying to access the new max line length settings (e.g. for form-parsing).

I believe I've fixed it (by using a default-constructed HTTPServerSettings).

http/vibe/http/server.d
```diff
@@ -657,6 +658,9 @@ final class HTTPServerSettings {
 	/// the url and all headers.
 	ulong maxRequestHeaderSize = 8192;
 
+        /// Maximum number of bytes in a single line in the request header.
+        ulong maxRequestHeaderLineSize = 4096;
```
s-ludwig, line 662: This should be `size_t` in order to be able to be passed to `parseFormData` on 32-bit builds. `maxRequestHeaderSize` should probably also be a `size_t`, since it needs to fully fit into memory, too, but that's a different issue.

Minor note: the indentation here uses spaces instead of tabs.

PR #2759, opened 2023-12-12
Removes */meson.build files

### LINKED ISSUE #2753: Proposal: remove whole Meson support at all
I suggest to stop pretending that everything is fine with Meson support and propose to remove whole Meson support at all from Vibe.d and its dependencies(!) which are run by Vibe.d community

I tried to build current `~master` with latest Meson (version 1.2.3):
```
$ ~/Dev/meson/meson.py setup check_meson_build
utils/meson.build:33:17: ERROR: File vibe/utils/memory.d does not exist.
```

And I look into this about once every few months and nothing was changed. Previously it was told that someone using this Meson scripts (https://github.com/vibe-d/vibe.d/issues/2531#issuecomment-866710499), but I think that this is no longer true due to the fact that the scripts are still broken.

Also, I don't propose to fix these scripts because way they were written is weak from the beginning because the Meson was just in early state. Instead of correcting Meson scripts it need to be completely rewritten in accordance with modern Meson requirements. And this should be done by the who actually use these scripts. And this must be done simultaneously in all dependencies, which is an almost impossible task (as for me)

denizzzka:
Fix #2753


s-ludwig (approved):

s-ludwig:
For some reason the merge didn't happen as intended. I've merged this manually now due to conflicts: 7de62abaf67872972d291ceb7dfbfc0629d197af

PR #2762, opened 2024-01-28
Proxy handles repeated headers

Alexibu:
The proxy code was replacing previous instances of a header during copying.
Using addField instead of opIndexAssign preserves multiple instances of the same header.

### DIFF
--- a/web/vibe/web/common.d
+++ b/web/vibe/web/common.d
@@ -1256,11 +1256,11 @@ package void setVoid(T, U)(ref T dst, U value)
 }
 
 unittest {
-	static assert(!__traits(compiles, { bool[] barr; ParamError err;readFormParamRec(null, barr, "f", true, NestedNameStyle.d, err); }));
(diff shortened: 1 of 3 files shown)


Geod24:
Do you have a test for this ?

Alexibu:
No I don't have a test. I was going to do one now, but can't see a lot of tests. Is the stuff in tests/ all there is ?

s-ludwig:
@Alexibu The whole HTTP code base is now found at https://github.com/vibe-d/vibe-http/ - including the tests. Sorry that this happened while your PR was still open. The code is identical for now though, so the current diff should apply as-is.

The main question that I had when looking at the changes is the handling of the "Connection" field. Did you make the change to `addField` there on purpose, or just for consistency? The reason I'm asking is that this field should, under normal circumstances, never be distributed over multiple entries and (without having double checked in the actual code) I'm a little worried that we might accidentally end up with two entries some way.

PR #2763, opened 2024-02-04
Fix handling of enum parameters for REST interface clients

s-ludwig:
Instead of by name, enum values were serialized by value, causing a conversion error on the server side. This now mirrors the cases in `fromRestString` to `toRestString`, ensuring they match up.

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -483,6 +483,8 @@ interface Example8API
 		double d;
 	}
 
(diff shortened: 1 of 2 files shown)


Geod24:
Can we add a test with an enum with duplicated values ?
`enum E { foo = 0, bar = 1, foobar =0 }` ?

s-ludwig:
Done

l-kramer (approved):

PR #2771, opened 2024-02-06
Remove support for non-safe REST methods

s-ludwig:
Takes up on #2748 and uses a clear error message instead of letting the compiler output indirect errors.

Closes #2748.

### DIFF
--- a/examples/rest-collections/source/app.d
+++ b/examples/rest-collections/source/app.d
@@ -48,7 +48,10 @@ class LocalThreadAPI : ThreadAPI {
 
 	string[] get()
 	{
-		return m_data.threads.keys;
+		static if (__VERSION__ < 2099) {
+			// NOTE: .keys is not @safe on older compiler versions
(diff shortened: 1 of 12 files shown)


examples/rest-collections/source/app.d
```diff
@@ -48,7 +48,8 @@ class LocalThreadAPI : ThreadAPI {
 
 	string[] get()
 	{
-		return m_data.threads.keys;
+		// NOTE: .keys is not @safe on older compiler versions
```
Geod24, line 51: Do you know from what version it is ?

s-ludwig: I tested it now on run.dlang.io and it works starting with 2.098.1.

s-ludwig: Rebased and added a `static if(__VERSION__)`.

l-kramer (approved):

PR #2778, opened 2024-02-10
Fix a few scope related deprecation warnings

s-ludwig:


http/vibe/http/common.d
```diff
@@ -300,7 +300,7 @@ class HTTPResponse {
 		InetHeaderMap headers;
 
 		/// All cookies that shall be set on the client for this request
-		@property ref DictionaryList!Cookie cookies() { return m_cookies; }
+		@property scope ref DictionaryList!Cookie cookies() return { return m_cookies; }
```
Geod24, line 303:
```suggestion
		@property ref DictionaryList!Cookie cookies() scope return { return m_cookies; }
```
?

stream/vibe/stream/multicast.d
```diff
@@ -75,10 +75,10 @@ struct MulticastStream(OutputStreams...) {
 		if (m_tasks.length > 0) {
 			Exception ex;
 			foreach (i, T; OutputStreams[1 .. $])
-				m_tasks[i] = runTask({
-					try m_outputs[i+1].flush();
+				m_tasks[i] = runTask((MulticastStream _this) {
```
Geod24, line 78:
```suggestion
				m_tasks[i] = runTask((scope MulticastStream _this) {
```
?

Geod24 (approved):

http/vibe/http/common.d
s-ludwig: Do we even need the `scope` at all here? I still couldn't force myself to look through the spec to fully grasp the `scope` additions (assuming that everything is correctly documented by now) and from the observed effects I'm unsure whether `return` a) implies `scope`, b) is required to be used in conjunction with `scope`, or c) has an independent effect on the parameter/`this` pointer.

stream/vibe/stream/multicast.d
s-ludwig: Applied to both similar places.

http/vibe/http/common.d
Geod24: We need `scope` only if we intend to call this function with a `scope DictionaryList!Cookie` is my understanding. But I have a similar issue with the spec.

PR #2803, opened 2024-08-20
Mail: Implement SMTP XOAUTH2 authentication

Geod24:
A fairly simple implementation allowing to send emails to M365 / gmail.


Geod24:
CI is broken @s-ludwig ?

s-ludwig:
CI is behaving in a strange way. I restarted some jobs manually and they passed, but re-running all failed jobs made some of them fail again. Seems as if some runners might have a lower RAM limit?

In any case, lets just merge this, as there is nothing that should be compiler version specific.

PR #2811, opened 2025-01-02
ci: Upgrade supported compiler & dub version

Geod24:
Test with only the latest 10 releases.


Geod24:
Seems to work but required pipelines will need to be changed by an admin (@s-ludwig )

PR #2813, opened 2025-01-11
Fix Packet size mismatch! error on collection drop

gedaiu:
fix https://github.com/vibe-d/vibe.d/issues/2812


gedaiu:
@WebFreak001 @s-ludwig any thoughts on this pr?

WebFreak001 (approved):

WebFreak001:
sounds good but please don't put it in the fundamental `_connection` test, but rather in some (perhaps new) collection management test

gedaiu:
@s-ludwig @WebFreak001 ready for another look

tests/mongodb/collection/source/app.d
```diff
@@ -0,0 +1,39 @@
...
+		? args[1].to!ushort
+		: MongoClientSettings.defaultPort;
+	runTask(() nothrow {
+		try runTest(port);
+		catch (Exception e) assert(false, e.toString());
+		finally exitEventLoop(true);
+	});
+	runEventLoop();
+	return ret;
+}
```
s-ludwig, line 39: I think we should simplify this a bit, some of the other tests have accumulated some historic cruft

```suggestion
void main(string[] args)
{
	ushort port = args.length > 1
		? args[1].to!ushort
		: MongoClientSettings.defaultPort;
    runTest(port);
}
```

tests/mongodb/collection/source/app.d
```diff
@@ -0,0 +1,32 @@
...
+import std.conv : to;
+import std.encoding : sanitize;
+import std.exception : assertThrown;
+
+
+void runTest(ushort port)
+{
+	MongoClient client = connectMongoDB("127.0.0.1", port);
+
+	/// Drop a collection
```
WebFreak001, line 21: is this test only testing dropping non-existent collections?

Should probably drop it twice just in case if it randomly existed beforehand on CI. It should behave the same both times after all.

Otherwise it might make sense to also create a collection here.

gedaiu:
From what I tested, the issue happens even when the collection exists. So
it does not matter if I do it once or twice.

s-ludwig:
I've just tested locally and for me the test passes when I insert just a single `{_id: ...}` entry into the collection beforehand. So I'd say let's just add a second drop to make it safe.

WebFreak001:
> it does not matter if I do it once or twice.

prove it in the test then? :D

Right now the CI fails with `command failed: ns not found in vibe.db.mongo.collection.MongoCollection.drop (../../../mongodb/vibe/db/mongo/collection.d:1238)` on all MongoDB versions prior to 7.0

So it looks like in older MongoDB versions this didn't work.

Now we could either ignore this error on the vibe.d side to make it behave the same with all MongoDB versions (and perhaps return a boolean if we can extract it in >=7 as well and make it part of the API) or just say double free has unspecified behavior.

WebFreak001:
tests still failing due to double drop

gedaiu:
hey @WebFreak001 can you give this PR another chance? I tried yesterday to use this client with a mongo 8 server and the CI failed because of this issue.

PR #2818, opened 2025-03-10
feat(smtp): Allow to specify a timeout in sendMail

Geod24:
Otherwise a program may get stuck indefinitely.


Geod24:
We've had two outages recently, so moving ahead with a release in the next 30 minutes.

PR #2819, opened 2025-03-11
feat(smtp/sendMail): Also use timeout as readTimeout

Geod24:
As the connection is more likely to hand on read than connect.


s-ludwig (approved):

s-ludwig:
By the way, in order to avoid the extra work for a dedicated release, assuming that the sending is done from within a task, you should also be able to pretty easily work around the hang using something like this:
```D
auto task = Task.getThis();
auto watchdog = setTimer(30.seconds, { task.interrupt(); });
scope (exit) watchdog.stop();
sendMail(...);
```

PR #2825, opened 2025-06-06
chore(README): Remove outdated instructions for OpenSSL

Geod24:
Since 2022 (OpenSSL 3 support in Deimos), the way to select the OpenSSL version has changed, and those instructions are no longer accurate. The code for it is now in vibe-stream, which is a separate package.
Instead of updating those instructions, I think it is better to remove them, as at the time they were added, it was a common problem to have to select the OpenSSL version. However, with the auto-detect script now in Deimos/OpenSSL, we are no longer seeing issues about this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,36 +168,3 @@ Install the DMD compiler and vibe.d's dependencies using portupgrade or a simila
     sudo portupgrade -PN devel/pkgconf
 
 Optionally, run `./setup-freebsd.sh` to create a user/group pair for privilege lowering.
-
-
-Switching between OpenSSL versions
-----------------------------------
-
-By default, vibe.d is built against OpenSSL 1.1.x. On systems that use the older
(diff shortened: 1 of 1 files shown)


s-ludwig:
> no longer accurate

Have you double checked whether this is actually true? From a quick glance, the instructions still seem to be correct within the vibe.d structure. If this doesn't work anymore, then it probably needs to be fixed in `vibe-stream:tls`' dub.sdl. I do occasionally still have issues with the automatic detection, or Linux dependency issues where the "notls" configuration is an easy way out, so I think it still makes sense to have these instructions somewhere.

Geod24:
> Have you double checked whether this is actually true?

I had and I got it wrong, so good shout out ! I looked at the `vibe-stream` package but not the `tls` subpackage...
What would be a good place to put it ? `vibe.stream.tls` module documentation ?

PR #2830, opened 2025-08-04
Reduce compile time

burner:
As the Endian enum only contains two members doing a static if compiles a lot faster than call std.conv.to on the enum at CT.

Compare the uploads trace_base and trace. It reduces the compile time by about 10ms from 435 to 425 ms.

[trace.json](https://github.com/user-attachments/files/21578334/trace.json)
[trace_base.json](https://github.com/user-attachments/files/21578335/trace_base.json)


burner:
@s-ludwig thank you

PR #2837, opened 2025-11-21
feat(rest): Provide more information on error

Geod24:
I have a use case with an API (Nomad) that doesn't set any Content-Type on error, returns a 500 (for something that should be 400), and leads to a very user-unfriendly experience when using the tool. This should provide better out-of-the-box experience for vibe.web.rest user, by recognizing 'text/plain' messages, and otherwise encoding the binary data so the caller may be able to do something with it.

### DIFF
--- a/tests/vibe.web.rest.error/source/app.d
+++ b/tests/vibe.web.rest.error/source/app.d
@@ -1,0 +1,87 @@
+import std;
+import vibe.core.core;
+import vibe.core.stream;
+import vibe.data.bson;
+import vibe.data.json;
+import vibe.data.serialization;
+import vibe.http.client;
+import vibe.http.router;
+import vibe.http.server;
+import vibe.stream.memory;
+import vibe.stream.operations;
+import vibe.web.rest;
+
+void main()
+@safe {
+	auto settings = new HTTPServerSettings;
+	settings.port = 0;
+	settings.bindAddresses = ["127.0.0.1"];
+	auto router = new URLRouter;
+    router.registerRestInterface(new Server);
+    router.post("/api/error1", &handleError1);
+    router.post("/api/error2", &handleError2);
+	auto listener = listenHTTP(settings, router);
+	scope (exit) listener.stopListening();
+	immutable addr = listener.bindAddresses[0];
+
+	auto api = new RestInterfaceClient!API("http://"~addr.toString);
+
+    // Test that regular exception lead to JSON error
+    try {
+        api.getError();
+        assert(0);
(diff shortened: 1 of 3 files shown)


web/vibe/web/rest.d
```diff
@@ -2107,10 +2107,26 @@ private HTTPClientResponse request(URL base_url,
...
-		Json msg = Json(["statusMessage": Json(client_res.statusPhrase)]);
-		if (client_res.contentType.length)
-			if (client_res.contentType.splitter(";").front.strip.sicmp("application/json") == 0)
+		import std.base64;
+
+		Json msg;
+		if (client_res.contentType.length) {
+			const ctype = client_res.contentType.splitter(";").front.strip;
+			if (ctype.sicmp("application/json") == 0)
 				msg = client_res.readJson();
```
MrSmith33: Will this case contain `statusMessage`? Should we have `"statusMessage": Json(client_res.statusPhrase)` here too?

tests/vibe.web.rest.error/source/app.d
```diff
@@ -0,0 +1,87 @@
...
+
+	auto api = new RestInterfaceClient!API("http://"~addr.toString);
+
+    // Test that regular exception lead to JSON error
+    try {
+        api.getError();
+        assert(0);
+    } catch (RestException exc) {
+        assert(exc.message() == "Something very bad happened");
+        assert(exc.jsonResult["statusDebugMessage"].get!string.canFind("object.Exception@source/app.d"));
```
MrSmith33: Fails on Windows due to different directory separator: `/` vs `\`. On Windows it is `core.exception.AssertError@source\app.d(35): Assertion failure`

web/vibe/web/rest.d
MrSmith33: Yeah, `RestException` already has `statusCode`.

Just pointing out that all cases except `application/json` have `statusMessage` field set.

Geod24:
Ready for review @s-ludwig

s-ludwig:
I've been thinking about the "statusMessage" topic a bit, but ultimately I think the change is best like it is now. It might make sense to add an additional `statusPhrase` field to `HTTPException` or `RestException` to catch the phrase in a semantically unambiguous place, but considering that it carries no defined semantic meaning this doesn't look like a particularly high-priority addition.

The only thing I'd improve would be to merge the two almost identical "unknown content type" else branches (have `ctype` just be empty if no content type is given and then do the two comparisons).

PR #2841, opened 2026-01-15
Fix and Example how to return binary data with Content-Type

burner:
While looking into how to return a image from a REST api, I tried using the viaHeader out parameter to assign the Content-Type.

This didn't work, as the DefaultSerializer sets the Content-Type always to "application/json". Defining a custom Serializer doesn't work as it only allows a single Content-Type and images can have multiple. Returning an InputStream does bypass the DefaultSerializer, but the Content-Type set via the out parameter still failed. The fix was the to change the order of the
* evaluateOutputModifiers
* returnHeaders calls, making them similar to the other code paths. The example shows how to set the Content-Type and return a stream.

### DIFF
--- a/tests/rest/source/app.d
+++ b/tests/rest/source/app.d
@@ -8,6 +8,7 @@ import vibe.appmain;
 import vibe.core.core;
 import vibe.core.log;
 import vibe.core.stream : InputStreamProxy;
+import vibe.stream.memory;
 import vibe.data.json;
 import vibe.http.router;
 import vibe.http.server;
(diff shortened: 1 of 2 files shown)


tests/rest/source/app.d
```diff
@@ -562,6 +563,34 @@ class Example8 : Example8API
 	}
 }
 
+/* --------- EXAMPLE 9 ---------- */
+
+/*
+ * This example shows how to return data from a rest endpoint and setting the
+ * Content-Type. If and only if the return type is a vibe.d InputStream, which
+ * MemoryStream is, assigning the Content-Type via out variable will work.
```
Geod24: Why limit it to an `InputStream` ?

Geod24: There is the same issue with serializer:
> res.writeBody(serialized_output.data, serializer.contentType);

PR #2869, opened 2026-03-25
Implement OP_COMPRESSED wire protocol support with zlib compression

### LINKED ISSUE #2857: [vibe-d/mongodb] [important] [mongo8] OP_COMPRESSED defined but not implemented
We have the `OpCode.Compressed` enum value, but never actually compress anything. No snappy, zlib, or zstd, and we don't negotiate compression in the handshake.

Matters most for large documents or high throughput. zstd gives the best compression, and snappy is the cheapest on CPU.

gedaiu:
Adds support for MongoDB OP_COMPRESSED (opcode 2012) to the wire protocol. Compression is negotiated during the hello handshake and applied transparently to outgoing OP_MSG messages. Incoming compressed responses are detected and decompressed before parsing. Only zlib is implemented for now; the architecture supports adding snappy and zstd later. Configure via connection string with ?compressors=zlib.

fix #2857

### DIFF
--- a/mongodb/vibe/db/mongo/connection.d
+++ b/mongodb/vibe/db/mongo/connection.d
@@ -1521,6 +1521,7 @@ private void parseOpMsgBody(bool dupBson)(
 	}
 }
 
+/// negotiateCompressor picks first client-preferred compressor supported by server
 unittest
 {
 	assert(negotiateCompressor([Compressor.zlib], ["zlib"]) == Compressor.zlib);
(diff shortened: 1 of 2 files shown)


s-ludwig (approved):
Ultimately, we should have `parseOpMsgBody` be used for both the uncompressed and the compressed path, possibly by using an input range for both.

gedaiu:
@s-ludwig what's missing from this PR to be merged?

PR #2870, opened 2026-03-25
Shared topology across MongoDB connections

### LINKED ISSUE #2846: [vibe-d/mongodb] Topology should be shared across connections
Every time a connection calls `connect()`, it creates a fresh `TopologyDescription`, probes every seed host, picks one, then discards the topology. With a 3-node replica set, that's 4 TCP handshakes per connection.

The topology should live on `MongoClient` and be shared by all connections in the pool. Connections should obtain their target host from the client rather than discover it themselves.

This is a prerequisite for write routing, per-query read preference, and background monitoring.

gedaiu:
Topology discovery now occurs once per MongoClient, not per connection. With a 3-node replica set, new pool connections go from 4 TCP handshakes to 1.

Modelled after the Node.js driver's SDAM architecture. MongoClient owns the TopologyDescription and passes the selected host directly to each connection. On connection failure, the topology is re-discovered and retried once.

fix #2846


mongodb/vibe/db/mongo/connection.d
```diff
@@ -309,11 +313,13 @@ final class MongoConnection {
 		}
 	}
 
+	deprecated("Topology discovery is now handled by MongoClient. Use connectToHost instead.")
```
s-ludwig: Since `MongoConnection` is an internal class, I think we can simply remove this without deprecation.

gedaiu:
@s-ludwig what's missing from this PR to be merged?

s-ludwig:
Sorry, I haven't been able to keep up with GitHub in April and missed the updates. Needs a manual rebase now that #2869 is merged.

gedaiu:
No worries. I fixed the conflicts, and the branch should be ready now. I want to move forward with my crate library, and not having modern MongoDB support is a blocker for me :(

PR #2873, opened 2026-04-10
Fix estimatedDocumentCount crash on non-existent collections

gedaiu:
The $collStats aggregation pipeline used by estimatedDocumentCount throws a MongoDriverException when the collection does not exist in the database. This causes repeated errors in applications that register collections before they are created. The fix catches the exception and returns 0, matching the behavior of MongoDB's shell estimatedDocumentCount() on missing collections.


gedaiu:
@s-ludwig would you have time to give a quick look at this PR?

PR #2879, opened 2026-05-28
Fix #2878 - cannot request body on a HEAD request, even on error

### LINKED ISSUE #2878: vibe.web.rest doesn't support `@method(HTTPMethod.HEAD)`
```D
/+ dub.sdl:
    name "head-test"
    dependency "vibe-d" version="~>0.9"
+/

import vibe.vibe;
import vibe.web.rest;
import std.stdio;

interface IGoogleAPI
{
    @method(HTTPMethod.HEAD)
    @path("/")
    void root (out @viaStatus int status) @safe;
}

void main() @safe
{
    runTask(() @safe nothrow {
        scope (exit) exitEventLoop();

        try {
            auto settings = new RestInterfaceSettings;
            settings.baseURL = URL("https://www.google.com");
            auto client = new RestInterfaceClient!IGoogleAPI(settings);
            int status;
            client.root(status);
            logInfo("HEAD https://www.google.com/ => %d", status);
        } catch (HTTPStatusException e) {
            logError("HEAD https://www.google.com/ => HTTP %d %s", e.status, e.msg);
        } catch (Exception e) {
            logError("Error: %s", e.msg);
        }
    });

    runEventLoop();
}
```

adamdruppe:
see https://github.com/vibe-d/vibe.d/issues/2878

### DIFF
--- a/tests/vibe.web.rest.2878/dub.sdl
+++ b/tests/vibe.web.rest.2878/dub.sdl
@@ -1,0 +1,2 @@
+name "vibe-web-rest-2878-test"
+dependency "vibe-d:web" path="../../"
--- a/tests/vibe.web.rest.2878/source/app.d
+++ b/tests/vibe.web.rest.2878/source/app.d
@@ -1,0 +1,46 @@
+import vibe.core.core;
+import vibe.http.client;
+import vibe.http.router;
+import vibe.http.server;
+import vibe.web.rest;
+
+void main()
+@safe {
+	auto settings = new HTTPServerSettings;
+	settings.port = 0;
+	settings.bindAddresses = ["127.0.0.1"];
+	auto router = new URLRouter;
+    router.get("/api/error", &handleError);
+    router.get("/api/ok", &handleOk);
+	auto listener = listenHTTP(settings, router);
+	scope (exit) listener.stopListening();
+	immutable addr = listener.bindAddresses[0];
+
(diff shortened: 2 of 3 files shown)


adamdruppe:
whoops copy/paste missed a letter

s-ludwig:
Doesn't this currently mix two separate issues?
- Fixes the issue with attempting to read the body of a HEAD response
- Allows to bypass the `RestException` logic

I think the best way would be to prepend the following `if` case to the `if`-`else` chain of the `!isSuccessCode` block:
```D
if (verb == HTTPMethod.HEAD)
	msg = Json(["statusMessage": Json(client_res.statusPhrase)]);
else if ...
```

And then the `throw RestException` part could be changed so that it is only executed if there is no `@viaStatus` parameter present (in `executeClientMethod` checking for `sroute.parameters.any!(p => p.kind == ParameterKind.status)`). It would be pretty nice to generally allow bypassing exceptions like that, for cases where the status code is transporting more fine grained semantics.

adamdruppe:
yeah that sounds reasonable. i'll look into it a bit later.

adamdruppe:
@s-ludwig this commit kinda what you had in mind? https://github.com/vibe-d/vibe.d/pull/2879/changes/a67ddc899e23c52f5d5a5e9cf5ba8182a5df230f

s-ludwig (approved):
Yeah, that looks like what I had in mind, thanks!

PR #2880, opened 2026-06-11
feat(mongo): modernize driver with SDAM, sessions, and wire features

### LINKED ISSUE #2845: [vibe-d/mongodb] Small cleanup: dead imports, noisy log, single-pass discovery
Three quick things:

1. `allKnownHosts()` in `topology.d` imports `std.algorithm` and `std.array` but doesn't use them
2. `connectToHost()` logs at `logInfo` level on every connection — should be `logDiagnostic`
3. Host discovery in `connect()` only does one pass. If a newly discovered host reports more hosts we don't know about, we miss them. Should loop until there's nothing new.

### LINKED ISSUE #2846: [vibe-d/mongodb] Topology should be shared across connections

### LINKED ISSUE #2847: [vibe-d/mongodb] Writes don't get routed to the primary
Writes currently go through whatever connection the pool hands out. If that connection points to a secondary, the write just fails with a server error instead of being sent to the right place.

The client should use the shared topology to always send writes (insert, update, delete) to the primary, regardless of the configured read preference.

Depends on shared topology. https://github.com/vibe-d/vibe.d/issues/2846

### LINKED ISSUE #2848: [vibe-d/mongodb] Can't override read preference per query
Read preference is set once at the client level. There's no way to say "run this specific query on a secondary" while keeping the default on primary.

This matters for analytics queries that should go to secondaries while normal reads stay on the primary.

Add an optional `ReadPreference` parameter to `find()`, `aggregate()`, and `runCommand()`.

Depends on shared topology. https://github.com/vibe-d/vibe.d/issues/2846

### LINKED ISSUE #2849: [vibe-d/mongodb] No background health monitoring
The driver only learns about the replica set state during `connect()`. Between operations, it has no idea if a member went down or a new primary was elected.

We need a background task that periodically sends `hello` to all known members and updates the shared topology. The MongoDB SDAM spec recommends a 10-second default interval.

Depends on shared topology. Topology update methods will need to be thread-safe. https://github.com/vibe-d/vibe.d/issues/2846

### LINKED ISSUE #2850: [vibe-d/mongodb] No retry when primary steps down
If the primary steps down mid-operation, the driver gets a `NotWritablePrimary` error and gives up. It should catch that, refresh the topology, find the new primary, and retry the operation once.

Only safe to retry idempotent operations or those with session support.

Depends on write routing and heartbeat monitoring. https://github.com/vibe-d/vibe.d/issues/2846 https://github.com/vibe-d/vibe.d/issues/2849

### LINKED ISSUE #2851: [vibe-d/mongodb] No read preference tag support
Can't target specific secondaries by tags like `{dc: "east"}` or `{role: "analytics"}`. The `tags` field is already parsed in `ServerDescription` but nothing uses it.

Need to parse `readPreferenceTags` from the connection URI and filter candidates in `selectServer()` by matching tag sets before picking one.

Depends on the per-operation read preference. https://github.com/vibe-d/vibe.d/issues/2848

### LINKED ISSUE #2854: [vibe-d/mongodb] [critical] [mongo8] No client session support
We parse `logicalSessionTimeoutMinutes` from the handshake but never actually create or use sessions. Without them we can't do transactions, causal consistency, or retryable writes.

Need a `startSession()` on `MongoClient`, attach `lsid` to commands, and clean up with `endSession()`. Also need implicit sessions for operations that require them.

Blocks transactions and retryable writes.

### LINKED ISSUE #2855: [vibe-d/mongodb] [critical] [mongo8] No multi-document transaction support
No way to start, commit, or abort transactions. They've been around since MongoDB 4.0 for replica sets and 4.2 for sharded clusters.

Transactions need to pin to a specific server, so this depends on sessions (https://github.com/vibe-d/vibe.d/issues/2854) and shared topology (https://github.com/vibe-d/vibe.d/issues/2846).

### LINKED ISSUE #2856: [vibe-d/mongodb] [critical] [mongo8] No retryable writes
Since MongoDB 3.6, drivers are supposed to automatically retry writes once on network errors or elections. The server deduplicates using the session's transaction number, so it's safe.

Different from failover retry (https://github.com/vibe-d/vibe.d/issues/2850) — this is a session-level mechanism that the server coordinates, not just topology re-selection.

Requires sessions. https://github.com/vibe-d/vibe.d/issues/2854

gedaiu:
Brings the MongoDB driver up to modern server-spec compliance so it can run reliably against replica sets and current MongoDB releases: SDAM-based topology monitoring with primary step-down recovery, server sessions with retryable writes and multi-document transactions, and per-query read preferences. On top of that it adds change streams, mongodb+srv seedlist resolution, load-balanced mode, Stable API, OP_COMPRESSED wire compression, client-side field level encryption, GridFS, and MongoDB 8.0 server-level bulkWrite, each backed by integration tests under tests/mongodb. Also fixes estimatedDocumentCount on missing collections and switches to X.509 auth when an SSL key file is configured.

Fixes #2845, fixes #2846, fixes #2847, fixes #2848, fixes #2849, fixes #2850, fixes #2851, fixes #2854, fixes #2855, fixes #2856, fixes #2858, fixes #2859, fixes #2860, fixes #2861, fixes #2862, fixes #2863, fixes #2864.


gedaiu:
Closing in favor of a stacked, reviewable decomposition

Hey! This branch grew into a ~14k-line near-total rewrite of the MongoDB driver, which is far too large to review in one piece. I've split it into 9 focused, dependency-ordered PRs, each based on the current upstream master, each compiling on its own with unit tests green, and each closing specific tracked issues. Together they reproduce this branch exactly, so I'm closing it here and moving the work into the stack below (please review in order):

#2882 — MongoDB driver SDAM core: topology discovery, health monitoring, primary write-routing, per-query read preferences, read-preference tags. Closes #2845, #2847, #2848, #2849, #2851
#2883 — Client sessions, multi-document transactions, retryable writes. Closes #2854, #2855, #2856, #2850 · merge after #2882
#2884 — Change streams. Closes #2860 · merge after #2883
#2885 — Server-level bulkWrite. Closes #2861 · merge after #2884
#2886 — GridFS. Closes #2862 · merge after #2885
#2887 — CSFLE (client-side field level encryption). Closes #2863 · merge after #2886
#2888 — Stable API (Versioned API). Closes #2858 · merge after #2887
#2890 — Load-balancer mode. Closes #2864 · merge after #2888
#2889 — mongodb+srv:// connection strings. Closes #2859 · merge after #2890
Each PR's diff is scoped to a single feature, so they're small enough to review properly. Thanks for taking a look!

